Issue #409, opened 2014-03-13
Fix export for new unified key list


dschuermann:
We now have a unified key list thanks to @Valodim with pull request https://github.com/openpgp-keychain/openpgp-keychain/pull/403
Some things need to be reworked now.

Export of public keys works
Export of private keys is broken: File has no content.

In general: The export dialog should always export the public key of a key pair. Additionally, we should add a CheckBox to the Dialog "Export private key". This should now enable the export of the private part.

Issue #255, opened 2014-02-04
On tablets: Always opened navigation drawer


dschuermann:
This would significantly improve the layout for tablets:

http://stackoverflow.com/questions/17133541/navigation-drawer-set-as-always-opened-on-tablets
http://derekrwoods.com/2013/09/creating-a-static-navigation-drawer-in-android/

PR #421, opened 2014-03-14
Fix export for new unified key list #409 and lock drawer in tablets #255

uberspot:
Fixes issue #409 and implemented #255 
Now the drawer stays always open in tablets (check /res/layout-large ). It works as expected. You can also easily limit it to only landscape mode in tablets if you rename /res/layout-large to /res/layout-large-land if i'm not mistaken. 
I also cleaned the xml layouts a tiny bit to make them more organized. 
Tested on normal smartphone with 4.1.2, on nexus 7 tablet with 4.3 and on emulator with 2.3.3
Works as expected in all of them but it's better if you test it as well.

### DIFF
--- a/OpenPGP-Keychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
+++ b/OpenPGP-Keychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
@@ -19,6 +19,11 @@ package org.sufficientlysecure.keychain;
 import android.os.Environment;
 
 import org.spongycastle.jce.provider.BouncyCastleProvider;
+import org.sufficientlysecure.keychain.service.remote.RegisteredAppsListActivity;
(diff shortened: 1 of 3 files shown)


alexandracisneros:
Does this works out of the box? Do I need to add anything special to the classic drawer code to lock it as open in tablets? I ask because I tried the code I have always used for a navigation drawer and it doesn't lock it as open in tablets.

dschuermann:
You are commenting in a very old PR. Wie are now using a library for nav drawer.

alexandracisneros:
and which library is that?

dschuermann:
@alexandracisneros it.neokree:MaterialNavigationDrawer

alexandracisneros:
Wow this is great!!! All I ever wanted in a single library. Thank you very much.

Issue #954, opened 2014-10-09
IllegalArgumentException: Unknown URL in DeleteFileDialogFragment


dschuermann:
3.0.1
Android 4.4 and Android 4.1

```
java.lang.IllegalArgumentException: Unknown URL file:///storage/emulated/0/Download/encrypted.asc
at android.content.ContentResolver.delete(ContentResolver.java:1278)
at org.sufficientlysecure.keychain.ui.dialog.DeleteFileDialogFragment$1.onClick(DeleteFileDialogFragment.java:81)
at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:166)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:136)
at android.app.ActivityThread.main(ActivityThread.java:5097)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:515)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:785)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:601)
at dalvik.system.NativeStart.main(Native Method)
```

PR #1023, opened 2015-01-09
Issue #954 - fixed crash when file scheme is passed, broadened deletion scope

adithyaphilip:
Refers to https://github.com/open-keychain/open-keychain/issues/954

Fixed crash due to "file" scheme in Uri instead of expected "content" scheme by adding a java.io.File delete method using the path of the Uri, inside the "file" part of the deletion.

Broadened the scope of files we can delete by adding the aforementioned java.io.File delete method to the "content" part of the deletion also, which is called in case the ContentResolver fails. This was found to be necessary to delete the file while testing with ES File Explorer. Probably because the ContentProvider on the application we obtain the Uri from did not/refused to handle the delete operation?


dschuermann:
Great work, thanks!

PR #1037, opened 2015-01-29
Add ability to compile with java8

ligi:
otherwise we get:

```
:OpenKeychain:compileDebugJava
/home/ligi/git/3rd/open-keychain/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java:686: error: incomparable types: Object and int
                    if (values.get(UserPacketsColumns.RANK) == 0 && values.get(UserPacketsColumns.USER_ID) == null) {
                                                            ^
Note: Some input files use or override a deprecated API.
```

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -32,7 +32,7 @@ dependencies {
 
 android {
     compileSdkVersion 21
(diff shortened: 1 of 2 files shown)


dschuermann:
Thanks, merged!

PR #1047, opened 2015-02-21
Some Polish translated strings and some tiny cleanups

steelman:


dschuermann:
Thanks. Please translate using Transifex at https://www.transifex.com/projects/p/open-keychain/ . Your changes will be overwritten when we pull the transifex translations into our repo. I accepted your pull request due to db8ee77

steelman:
Isn't there any other way to help with the translation? I can use TX if it is absolutely necessary but, honestly, I hate creating new accounts.

dschuermann:
Yes this is the only way for this project. For most average people github is too complicated, thus we only do Transifex and then Transifex -> Github pulls.

Issue #1046, opened 2015-02-21
openkey-chain fails to build without android-19 platform


steelman:
I have prepared my build environment according to the instructions in the README.md. After a few failed builds I had to install

* android-19 platform (required by html-textview)
* Build Tools revision 19.1.0 (required by html-textview)
* Build Tools revision 21.1.1 (required by android-lockpattern)

It appears not only html-textview needs it:

```
find -name build.gradle | xargs grep targetSdkVer 
./OpenKeychain/build.gradle:        targetSdkVersion 21
./extern/TokenAutoComplete/library/build.gradle:        targetSdkVersion 19
./extern/TokenAutoComplete/example/build.gradle:        targetSdkVersion 19
./extern/KeybaseLib/Lib/build.gradle:        targetSdkVersion 19
./extern/html-textview/build.gradle:        targetSdkVersion 19
./extern/SuperToasts/supertoasts/build.gradle:          targetSdkVersion 19
./extern/SuperToasts/demo/build.gradle:         targetSdkVersion 19
```
Please either

* enumerate in the README.md all the versions of platform (API)  and Build Tools that open-keychain and the submodules require for a successful build or
* make everything work with the same (newest) available version.

After installing the older platform and Build Tools packages I managed to build the package.

dschuermann:
We should pick "make everything work with the same (newest) available version.".
Open for pull requests.

steelman:
I am a complete Android noob.

dschuermann:
@steelman No problem :) No need to do it, just leaving it open for others

montimaj:
Check this out! Hope it has resolved this issue
https://github.com/open-keychain/open-keychain/pull/1069

dschuermann:
@montimaj @steelman Can you try if it now builds with build tools 21.1.2 and SDK 21 only? This should be fixed now

montimaj:
Yup it works fine!

On Mon, Mar 2, 2015 at 10:03 PM, Dominik Schürmann <notifications@github.com
> wrote:

> Closed #1046 <https://github.com/open-keychain/open-keychain/issues/1046>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/open-keychain/open-keychain/issues/1046#event-244482704>
> .
>



-- 
*Sayantan Majumdar*

steelman:
```
FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #1070, opened 2015-03-01
Dynamic selection of build-tool & sdk

montimaj:
Hoping that issue #1046 has been resolved ! I have slightly modified the android {} blocks in most of the gradle files in the extern directory to allow the dynamic selection of compileSdkVersion and buildToolsVersion.I replaced the default compileSdkVersion and buildToolsVersion in the all the gradle files that used these two android attributes with

compileSdkVersion rootProject.ext.compileSdkVersion
buildToolsVersion rootProject.ext.buildToolsVersion
The project was successfully built and executed!


dschuermann:
Thanks for your effort.
Based on your pull request I implemented something similar now.
We are not using the dynamic detection of SDK and build tools because we want deterministic builds.

Issue #1050, opened 2015-02-22
Passphrase strength indicator


dschuermann:
see https://github.com/matt-allen/android-password-strength-indicator

* copy this code over to OpenKeychain
* Integrate into first time wizard

the-solipsist:
Does this use realistic password strength testing, as https://github.com/dropbox/zxcvbn does?

dschuermann:
Nope it does not. If someone wants to do it, go for it!

PR #1075, opened 2015-03-02
Passphrase strength indicator

adithyaphilip:
Refers to Issue https://github.com/open-keychain/open-keychain/issues/1050

Passphrase Strength Indicator added in CreateKeyInputFragemnt, below the passphrase EditText and above the passphrase verify EditText.

Looks like:
screenshot_2015-03-03-01-23-30

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyInputFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyInputFragment.java
@@ -32,6 +32,7 @@ import android.widget.AutoCompleteTextView;
 import android.widget.EditText;
 
 import org.sufficientlysecure.keychain.R;
+import org.sufficientlysecure.keychain.ui.widget.PasswordStrengthView;
 import org.sufficientlysecure.keychain.util.ContactHelper;
 
 import java.util.regex.Matcher;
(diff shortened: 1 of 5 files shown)


dschuermann:
Thanks for your effort.
* Please add the text from https://github.com/matt-allen/android-password-strength-indicator/blob/master/LICENSE to the files from the project. Similar to the headers in our class files.
* Please change the implementation that it uses our colors.

dschuermann:
Thinking about this further, it would be perfect to have the password meter integrated on the right side. Similar to what Twitter does on https://twitter.com/signup
This would make it less annoying and similar to what we do for the email field.

adithyaphilip:
I've changed the colors, and embedded the progress indicator in the EditText, as Twitter does. 
screenshot_2015-03-04-03-16-58
screenshot_2015-03-04-03-17-07
screenshot_2015-03-04-03-17-14
I wasn't sure if we should use absolute dp to specify the width and height of the progress indicator, so I made it cover 20% of the width of the EditText. However, this resulted in quite a bit of nesting in the xml layout, is there a better way to go about it?
Also, the passphrase might end up going behind the indicator if it's too  (extremely) long, would an implementation like below be more appropriate?
screenshot_2015-03-04-02-50-46

lezorich:
Hi, I didn't noticed that @adithyaphilip has this PR, I also worked on this patch. @adithyaphilip for the passphrase not going behind the indicator you can set padding on the edit text :) [Here](https://lh3.googleusercontent.com/-khxh3oS7SvQ/VPYvGlLgViI/AAAAAAAADhs/H65pR-q7aO4/w319-h567-no/Screenshot_2015-03-03-18-51-36.png) is my working branch for you to see how I did it (you did the PR first, so the issue is yours :p).

I thought that something like this looks nice:
![Image weak passphrasr]
(https://lh5.googleusercontent.com/-DvKQ7JlT1oE/VPYvH7OBgxI/AAAAAAAADh4/C_av353wuSQ/w319-h567-no/Screenshot_2015-03-03-18-50-59.png)

![Image medium passphrase]
(https://lh3.googleusercontent.com/-eiHntHlTezM/VPYvJAq-nLI/AAAAAAAADiE/TACy1aaO_dI/w319-h567-no/Screenshot_2015-03-03-18-51-18.png)

![Image strong passphrase]
(https://lh3.googleusercontent.com/-khxh3oS7SvQ/VPYvGlLgViI/AAAAAAAADhs/H65pR-q7aO4/w319-h567-no/Screenshot_2015-03-03-18-51-36.png)

dschuermann:
Looks already great, nice effort, @adithyaphilip and @lezorich 

* Use padding as @lezorich suggested
* Also add into SetPassphraseDialogFragment

Valodim:
wouldn't this be better placed in a standalone PassphraseEditText widget which extends EditText?

dschuermann:
@Valodim great idea!

dschuermann:
Thanks for your efforts @lezorich and @adithyaphilip . This counts as a patch to both of you! I merged the codebase of @lezorich because it was more ready and added a new issue here: https://github.com/open-keychain/open-keychain/issues/1106 .

PR #1079, opened 2015-03-03
Using ButterKnife for view injection.

ishan1604:
This can significantly reduce the amount of code we write,
and also will improve our code readability.

.gitmodules : Edit
This removes the untracked content warning in git

build.gradle : Edit
added new library and also verified the checksum with gradle-witness plugin as mentioned in the Readme.

CreateKeyFinalFragment.java : Edit
implemented butter knife in a fragment to show how neat our code would become with this library.

@dschuermann Could you please review this, I'd be keen to hear your thoughts and remarks on this.

Thanks

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,9 +19,11 @@
 [submodule "extern/minidns"]
 	path = extern/minidns
 	url = https://github.com/open-keychain/minidns.git
+	ignore = dirty
 [submodule "extern/TokenAutoComplete"]
 	path = extern/TokenAutoComplete
 	url = https://github.com/open-keychain/TokenAutoComplete
 [submodule "extern/safeslinger-exchange"]
 	path = extern/safeslinger-exchange
 	url = https://github.com/open-keychain/exchange-android
+	ignore = dirty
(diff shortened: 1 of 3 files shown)


dschuermann:
I am okay with the changes regarding git submodules, but I don't like the usage of Butterknife. I would like to keep the code Java 7 + Android annotations only without using an additional annotation processor. This is primarily to keep the code easy to understand for newcomers.
Honestly, in your example I don't see any big reduction of code ;)

Valodim:
I'm not sure. It adds complexity, has *more* lines of code than otherwise, and special cases become very unorthogonal.

Nice hint about the dirty = ignored though!

ishan1604:
Here you can find more examples and it would be clearer if it can reduce any complexity or not. However I treat it as syntactical sugar only. 
@dschuermann I'd still follow your opinion. 

@Valodim Is there any specific unorthogonal case in your mind? I'd love to see that, so that I can be prepared to handle it in my other apps where I am using ButterKnife. Thanks for the review :)

dschuermann:
Please open a new PR for git submodules :)

Issue #1077, opened 2015-03-02
Import progress sometimes reset to 0%


dschuermann:
Somehow the dialog showed 0% and no progress half way through a previously successfull import.

indrora:
When you say "many keys", is this "many private keys (e.g. a ton of RSA/DSA keypairs) or "Here's a hundred thousand public keys"?

dschuermann:
100 public keys + 5 public/private key pairs

indrora:
Huh, happens when I import ~650 keys, but not when I import ~200. According to [the documentation](http://developer.android.com/reference/android/os/TransactionTooLargeException.html), the key importer is creating a Parcel that's too big for the binder -- Bigger than 1MB.

indrora:
Found the root cause: Logs.

Android has a maximum cap on how big a Parcel/intent can be. Since the log is schlepped over the activity boundary as an extra, It counts towards the parcel weight.

I'm not sure what I can do to reduce it. All that the log contains is a bunch of integers, really. The only thing I can think of is a Singleton that just handles logs, but I'd rather not add a LogManager somewhere it doesn't belong (that'd be a hack, and I don't want that in something that's supposed to be secure).

Here's my current ideas to solve this:

* Implement a `LogManager` -- you feed it logs and it gives you tickets out. You can hand it a ticket and it returns you a log (makes keeping logs easier, but it adds a layer of complexity to logs themselves)
* In all operations that work with logs, have a singleton `OperationLog` that we carry around.
* Use the global Application to keep logs. This works, but isn't a terribly great idea, I don't think.

Fundamentally, big logs are breaking the 1MB parcel size cap, and big enough key imports are creating big logs.

indrora:
I've chosen a variation on option 2 which doesn't change the API at all.

indrora:
It appears this is two bugs in one! yay!

The `TransactionTooLargeException` was caused by a log being huge, but there's another problem that comes when you run Consolidate: the OOM killer.

This happens most when *not* running under debugging -- I cannot reproduce the issue when we are plugged in and adb is running. Possible reason? We're *slowing* something down by spewing so much into the debug log that we're avoiding a dead-ness. Or, alternatively, the GC runs differently when adb is running. 

[...]

PR #1080, opened 2015-03-03
Fix root cause of #1077

indrora:
When importing lots of keys, lots of messages about the status of keys
are generated, including some debug slime and general fluff that
isn't really needed a lot of the time. As a result, a serious bug can come
along after key imports or certain operations which cause a log to become
parceled: It gets too big (over the absurd 1Mbit size limit).

This commit implements a pool to "dehydrate" logs into: they are placed
into the pool (a ConcurrentHashMap) and a UUID assigned to them, which
is parceled along. When the OperationResult is un-parceled, it reads in
the appropriate UUID bits and rehydrates the appropriate log.

This avoids the 1Mbit size limit, reducing the parcel down to ~10 bytes,
well under that limit.

In order to avoid any memory leaks, the log pool removes a reference to
the log itself, allowing the log to die a natural death at the hands
of the GC.

(As a matter of note: This UUID is *only* used within OperationResult! It is
**NOT** made available outside!)

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/OperationResult.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/OperationResult.java
@@ -38,6 +38,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** Represent the result of an operation.
  *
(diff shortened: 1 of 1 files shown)


Valodim:
Hi!

Looks good. Do you know of any other projects which implement this method for serialization? Or rather, are you aware of any caveats to it, especially in regards to IPC?

We have a similar approach to deal with imports of large keyring files, check out [ParcelableFileCache](https://github.com/open-keychain/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/ParcelableFileCache.java). I'd like to use only one of those approaches in both places, but they are both very hacky so I'm not sure which one is better. I think I like yours better because it hides the hack in the parcel itself, rather than in the code which uses it, but we should be sure that this doesn't cause issues down the line if we go that way.

indrora:
@Valodim There aren't any caveats that I can think of. When the object is becoming a `Parcel` and when it is being unboxed out of a `Parcel` is when I'm doing this. Unless there's another thing that eats up parcels raw (which would be bad, but I don't know of anywhere it's done), this has a low impact on resources and no impact on IO.

One advantage of using this hydrate/rehydrate is that you never run the risk of leaking secrets. Ever. Writing anything to disk is slow (bad for performance) and because you're writing to flash memory on most phones, you run the risk of leaking sensitive information (bad for security).

This approach, where the object being parceled/un-parceled is handling such things means that

1. we don't spend time serializing things that are complex and will just be deserialized in just a few seconds to the user,
2. because we're using a reference to the big object being handled (in this case, an `OperationLog`) we avoid the cost of the GC coming along and reaping an object just after the parcel is bundled up (which does happen on a regular basis: the Android 5 ConcSweepGC spends an exorbitant amount of time after an object goes out of scope)
3. we never write anything to disk, which means we never risk leaking anything.

(This ticket is not the place to get into the security problems that plague writing to disk; This ticket focuses on a single place where performance is somewhat critical)

dschuermann:
Good arguments as far as I understand the problem.
As @Valodim suggested: It would be nice to see this implemented generally instead of ParcelableFileCache.
Do you know of any other projects which implement this method for serialization?

Valodim:
This technique changes the property that a parcelable is self-contained and can be reconstructed from its parcelled state more than once. That's not a requirement for us, but we should annotate it and throw an AssertionError if it happens (rather than an expected nullpointerexception). Other than that I think we can give it a shot, if there are unexpected problems later we can always revert.

indrora:
@dschuermann -> This has been the common trick within Android from what I can tell. 

@Valodim If we want to make it "many-times reconstructable", then we shouldn't invalidate out of the dehydrate pool (see the cleanup for style; one line change) once we ask or the log bac, thus our ticket is good for as long as the application is alive, and everyone's happy.

In that case, the rehydrator needs to make sure that if something goes wrong, we return null or a new Log with a message.

Not invalidating the ticket UUID actually has a bit of a memory benefit: Because we're not risking multiple copies of the Parcel kicking around, we're not leaking memory on accident because someone imported 2000 keys and turned around and encrypted a file for 1000 of them.

Valodim:
hmm yeah i considered using a weakhashmap for this, but we might just as well think of that if we ever encounter the many-times reconstructable scenario

Issue #1081, opened 2015-03-03
App Crashes when trying to change key configuration


ishan1604:
Version - 3.2beta1

Steps to reproduce
* Open Menu action "Create Key"
* Enter form details
* Click next
* Click on Change key configuration

PR #1082, opened 2015-03-03
Fixes Issue #1081

ishan1604:
The App crashed because the Adapter was trying to call a method on a
null object.

Where was the null object?
CheckBox was null because in the Layout file
view_key_adv_user_id_item.xml

There is no such widget with id = user_id_item_check_box

Fix :
I removed the two lines which were causing the bug, and now the screen
loads perfectly.


dschuermann:
thanks. This crash is a result of our latest refactorings :)

ishan1604:
Will this count towards one-patch rule? I am working on more of them however :) 

Thanks for reviewing @dschuermann 

I am intrigued to see your activity on this code base. :+1:

dschuermann:
@ishan1604 yes this counts towards the one-patch rule

Issue #1059, opened 2015-02-26
Crash on expiry change


dschuermann:
1. have key with expiry of today
2. change expiry to yesterday

```
AndroidRuntime  D  Shutting down VM
  8748         AndroidRuntime  E  FATAL EXCEPTION: main
  8748         AndroidRuntime  E  Process: org.sufficientlysecure.keychain, PID: 8748
  8748         AndroidRuntime  E  java.lang.RuntimeException
  8748         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.dialog.EditSubkeyExpiryDialogFragment$2.onClick(EditSubkeyExpiryDialogFragment.java:177)
  8748         AndroidRuntime  E  at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:160)
  8748         AndroidRuntime  E  at android.os.Handler.dispatchMessage(Handler.java:102)
  8748         AndroidRuntime  E  at android.os.Looper.loop(Looper.java:135)
  8748         AndroidRuntime  E  at android.app.ActivityThread.main(ActivityThread.java:5256)
  8748         AndroidRuntime  E  at java.lang.reflect.Method.invoke(Native Method)
  8748         AndroidRuntime  E  at java.lang.reflect.Method.invoke(Method.java:372)
  8748         AndroidRuntime  E  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:898)
  8748         AndroidRuntime  E  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:693)
```

montimaj:
@dschuermann I have compiled the latest version of the app and tested this expiry change activity. The app is crashing after doing the following:
1. Have key without expiry.
2. Set expiry to any date after today
3. Change expiry to any date between today and previous_date_set -1.
  The crash that has been mentioned here was occurring in the apk that is there in Google Play Store.  In the latest version the user cannot set the expiry date to yesterday if the previous_date_ set is today. 
Pull request: #1083

dschuermann:
More info:
It looks Android 5 again allows having older dates selected in the widget.

http://stackoverflow.com/questions/28085786/fix-datepickerdialog-setmaxdate-and-setmindate-on-android-lollipop

dschuermann:
The crash still happens but I was able to get the old spinner back (which in this case makes more sense than the big calendar view): be06c4c

dschuermann:
[...]

PR #1083, opened 2015-03-03
Fixed expiry change crash

montimaj:
Issue #1059  has been successfully resolved.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyExpiryDialogFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyExpiryDialogFragment.java
@@ -173,8 +173,14 @@ public class EditSubkeyExpiryDialogFragment extends DialogFragment {
                     long numDays = (selectedCal.getTimeInMillis() / 86400000)
                             - (expiryCal.getTimeInMillis() / 86400000);
                     if (numDays <= 0) {
-                        Log.e(Constants.TAG, "Should not happen! Expiry num of days <= 0!");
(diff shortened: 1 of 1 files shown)


montimaj:
@dschuermann The calender does not  allow the user to select any invalid date. So is there any use of 'numDays' at all? The crash can also be fixed simply removing the block if(numDays<=0) {...

dschuermann:
Obviously it somehow is possible to get numDays = 0....

montimaj:
So is there anything more to be done here? I think the issue is fixed.

dschuermann:
We don't want the user to select anything that results in an expiry date before today.

montimaj:
The calendar won't allow it 
screenshot_2015-03-03-17-33-12
screenshot_2015-03-03-17-33-37

dschuermann:
This is why we had this if clause there. This should not happen. Unfortunatly it looks Android 5 again allows having older dates selected in the widget.

dschuermann:
http://stackoverflow.com/questions/28085786/fix-datepickerdialog-setmaxdate-and-setmindate-on-android-lollipop

dschuermann:
Looks like we need to check on our own and inform the user about this.

Valodim:
@montimaj Sorry to say this, but your approach to this issue was *completely* wrong.

The check is there to make sure the variable stays in its expected domain. For good reason obviously, because with an update of android values outside this domain did in fact happen. Removing it is not only not the right solution here, but the way you initially did it (catching a runtimeexception right after it was thrown) makes no sense.

We appreciate your effort, but the goal here should be to get the actual issues fixed, not just remove a symptom and get a pull request accepted quickly.

Issue #1048, opened 2015-02-22
Replace Holo icons with Material Design icons


dschuermann:
* Replace all old holo icons with material design icons
* Remove old icons from project
* Use our download script in "Graphics" folder (see https://google.github.io/material-design-icons/ for icon list)
* Special icons on http://materialdesignicons.com/

ishan1604:
@dschuermann I am interested in taking this up. I looked at both the links. But I am not quite sure how to use the downloads script in Graphics folder.

I saw the shell script get-material-icons.sh, could you please guide me by giving one simple example on how to use it to download the icons from the given website.

What I wish to do is?
Learn from your example, implement it in the app to replace all the Holo icons with Material Design Icons and also write a Wiki Document on how this works for future reference of other developers.

Thanks

chipset95:
I have replaced most of the Holo drawbles with Material ones in [this](https://github.com/chipset95/open-keychain/commit/739917b754bb787a9ff8e51c08a55d7275394283) commit.

I wasn't able to use the scripts, but I got all the resources from [Android Asset Studio](http://romannurik.github.io/AndroidAssetStudio) and [Material Icons Index](https://google.github.io/material-design-icons/)

vectorijk:
Fetch icons by script . And I restructured icons and change icon names in layout file. 
All in [this](https://github.com/open-keychain/open-keychain/pull/1090) commit

PR #1086, opened 2015-03-03
Changed drawable resources to Material - issue #1048

chipset95:
Replaced some of the holo drawables with Material Design icons taken from https://google.github.io/material-design-icons/ as mentioned in issue #1048 

I was not able to use the provided scripts, but I managed to get most of the icons with correct sizes.


chipset95:
Missed one drawable, added it in commit 739917b

dschuermann:
* Why wasn't you able to use the script?
* Please use the original names of the icons
* Choose fitting icons not just by name

chipset95:
&bull; I ran the script but it gave me `missing res dir` with the res path of the app in the options.json files's basepath. Any help in usage of the the scripts would be appreciated.
&bull; I kept the file names same as those of holo, so that they won't have to be changed wherever they were referenced
&bull; I manualy matched the image with its purposes so that they don't loook out of place.

chipset95:
Just fixed a file name in the latest commit [3496a22](https://github.com/chipset95/open-keychain/commit/3496a22468eec62b3eba984584255bbc82168358)

dschuermann:
* Not sure why the script does not work. Are you using something different than Linux?
* Please use original file names
* Please use icons with "keys" instead of "people". (Also look at http://materialdesignicons.com/ for key icons)

dschuermann:
Done in https://github.com/open-keychain/open-keychain/pull/1090

PR #1087, opened 2015-03-03
Fixed expiry change crash(for Android L as well)

montimaj:
@dschuermann  Hopefully I have fixed this issue #1059  this time. If an invalid date is selected then the expiry date is set to the previous expiry date.
case1
//invalid date
case2
If an invalid date is selected then last expiry date is set.
case1

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyExpiryDialogFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyExpiryDialogFragment.java
@@ -19,6 +19,7 @@ package org.sufficientlysecure.keychain.ui.dialog;
 
 import android.app.Activity;
 import android.app.Dialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.os.Bundle;
 import android.os.Message;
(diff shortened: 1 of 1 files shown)


dschuermann:
After thinking about it, I would like to see the following:
If the user clicks on a date (DatePicker.OnDateChangedListener) that should be normally be disabled (but is _not_ due to the mentioned bug in Android 5), the date picker should reset itself to the old expiry date and show a Toast that this date should not be selected.

montimaj:
Ok got it the Toast part is left!

montimaj:
Hope this is what you suggested :) 
case3
case4
case5

dschuermann:
Can you do this in DatePicker.OnDateChangedListener?

montimaj:
Ok

dschuermann:
Even better: after clicking on a "disabled" date just hop back to the last selected date and show toast.

dschuermann:
The crash still happens but I was able to get the old spinner back (which in this case makes more sense than the big calendar view): https://github.com/open-keychain/open-keychain/commit/be06c4cd5839d9860485fc96e152794c68484e71

dschuermann:
One last remark for @montimaj : If you still want to work on that issue, please think about the purpose of the code before trying again.

PR #1090, opened 2015-03-03
Fix issue #1048 Material Design Icon

vectorijk:
Change name in layout file and UI java file according to name of icon file downloaded by script in Graphics directory.

Replace all old holo icons with material design icons
Remove old icons from project
Use download script to download some icons
device-2015-03-03-114314


dschuermann:
thanks for your effort!

PR #1092, opened 2015-03-03
Orange toolbar for certify and fingerprint activity added

ishan1604:


dschuermann:
This can be done in xml, no need to do it in-code.
Look into the included toolbar layout.

ishan1604:
I saw that. I did that using the code in java because if I make the changes
in the xml of the included toolbar to orange color, won't it reflect on all
the activities where it is referred to?

Or should I create a separate XML file for the orange toolbar and include
that there ?

Please advise, Thanks. :)

dschuermann:
Create a copy of the toolbar xml with an orange color and include it.

ishan1604:
Sent a new pull request with your suggested changes. Thanks for your advise

PR #1095, opened 2015-03-04
Don't dump the cursor to the log all the time

indrora:
We should avoid logging repeatedly like this.

This actually solves a horrible debugging problem where my logs were 2-3 *minutes* out of sync with reality.


dschuermann:
Hmmm, I found the logs pretty helpful. Maybe we could introduce a Constants.DEBUG_LOG_DB_QUERIES for it?

Valodim:
I like them as well, lag was never a big problem for me but it often helped for database-related debugging.

related note, what's with the three merge commits? :hurtrealbad:

dschuermann:
Please use git rebase for other pull requests to avoid the merge commits.

PR #1096, opened 2015-03-04
Show only one consolidate activity

indrora:
I'll let the commit message for 9142cdc5ba819a0cd12fdc95fb707cb2f6957f1d speak for itself: This fixes the second half of the bug described in #1077 enough that it looks right.

(now with rebase!)

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -657,6 +657,7 @@
         </activity>
         <activity
             android:name=".ui.ConsolidateDialogActivity"
+            android:launchMode="singleInstance"
             android:theme="@android:style/Theme.NoDisplay" />
         <activity
             android:name=".ui.PassphraseDialogActivity"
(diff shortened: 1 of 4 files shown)


indrora:
Git got really confused about there being two commits between those histories. I'm unsure what git wizardly I'd need to properly summon the two together.

dschuermann:
Could you exclude the changes in KeychainProvider.java and KeyListFragment.java? Then I would merge this directly.
KeyListFragment.java has this boilder plate code to allow us to see the snackbar with the log.

indrora:
I'll see what I can do. If I don't touch KeyListFragment, the debug function still has the same problem. 

I *highly* suspect that somewhere, just somewhere, the Consolidate dialog is being popped from within the consolidation. I can't confirm this though. I'll take some time and find what's invoking the activity then.

dschuermann:
@indrora Ah you are right :/

indrora:
This task is a little above me, and this isn't a real fix for the problem.

Closing this PR.

PR #1097, opened 2015-03-04
Orange toolbar for certify and fingerprint activity added

ishan1604:
Done as per your suggestion. 

Created a new layout xml and added it in the corresponding files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,11 +19,11 @@
 [submodule "extern/minidns"]
 	path = extern/minidns
 	url = https://github.com/open-keychain/minidns.git
-	ignore = dirty
+  	ignore = dirty
 [submodule "extern/TokenAutoComplete"]
 	path = extern/TokenAutoComplete
 	url = https://github.com/open-keychain/TokenAutoComplete
 [submodule "extern/safeslinger-exchange"]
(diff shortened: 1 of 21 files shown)


dschuermann:
thanks!

Valodim:
only one commit out of five in this pull request contains the actual fix for the issue, and half of that one reverts the changes made in the previous one. this is not a huge problem but please be a little more mindful of the git history.

ishan1604:
@Valodim Yes I agree. My bad, I was quit messed up with the pulls and merges. 

I will make sure this doesn't happen again. Sorry for the trouble caused to anyone who reviewed this commit of mine. :disappointed:

Valodim:
have a look at `git rebase -i`, it makes rewriting history very easy. this must not ever be done on commits which others have based their work on, but for branches only you work on it's a very helpful tool.

ishan1604:
Ohh yes. I use git rebase -i HEAD ~ <SomeNumberHere> to squash the commits when I send pull requests. I think we cannot use it for the commits that we've already pushed, what do you think?

Valodim:
that is correct. as I said, it's not big deal, just something for the future

PR #1098, opened 2015-03-04
Avoid doing bad things to adb (now with rebase)

indrora:
Dumping the cursor to the debug log over and over is a real resource hog
and tends to have the negative side effect of your logs being 2-3 minutes
out of sync with reality.


dschuermann:
Done in https://github.com/open-keychain/open-keychain/commit/7ccdd7d38ed1c7d2152b914be5d08323ebf434e4

Issue #1094, opened 2015-03-03
SHA512 is used instead of SHA256


dschuermann:
Not sure why but somehow SHA512 is used (I just saw it when testing https://github.com/open-keychain/open-keychain/issues/1093 ) but SHA256 is preferred in my settings (OpenKeychain's default).

dschuermann:
Re-produce:
1. create a key
2. Sign text with this key in Openkeychain and copy into clipboard.
3. Look at clipboard: It says it's a SHA512 cleartext signature.
-> It should be SHA256 like in our preferences

Valodim:
this is related to CanonicalizedSecretKey.getSupportedHashAlgorithms. apparently, we pay no heed to the preferred hashing algorithms specified in a keyring

montimaj:
The selected Hash algorithm will now be displayed!

dschuermann:
done in https://github.com/open-keychain/open-keychain/commit/c121657c2cf17ecd3d59809ff86f758b7b1a592c

kerlerm:
Does this commit also close https://github.com/open-keychain/open-keychain/issues/1001 ?

PR #1100, opened 2015-03-04
Changed SHA512 to SHA256(preferred)

montimaj:
Issue #1094


dschuermann:
Thanks for catching this, this should be SHA256 instead of SHA512.
But it is not exactly what I meant with #1094 . I will write more details for re-producing the problem inside the bug report.

Also you commented the preference upgrade mechanism. I commited your changes in https://github.com/open-keychain/open-keychain/commit/e312b0e675fcd81b19b95453f04b55aee927b1f3
(This also counts as a patch).

Issue #1042, opened 2015-02-15
use enums instead of int constants


Valodim:
so, not using enums is [out of date](http://stackoverflow.com/questions/4822877/why-doesnt-android-use-more-enums) at this point. we should totally get rid of int-constants wherever possible

dschuermann:
yeah, 21. century coding

danielnelz:
Hey,
I skimmed through some of the code today and replaced the magic int constants in the classes CreateKeyActivity.java and KeyFormattingUtils.java.
I created a pull request for this change.
In the next couple of days I'll be on the lookout for further int constants to replace while getting to know the codebase.
Thanks

danielnelz:
Replaced some more ints. Pull request already created.

Thanks!

PR #1101, opened 2015-03-04
#1042 replaced some int constants with enums

danielnelz:
Created enum FragAction in CreateKeyActivity.java and enum State in
KeyFormattingUtils.java and replaced int constants with them


dschuermann:
Great! Merged!

PR #1102, opened 2015-03-05
Fixed issues with preferred Hash and Encryption Algorithms

montimaj:
Issue #1094


montimaj:
Default selected Hash Algorithm is: SHA256
screenshot_2015-03-05-10-05-52
screenshot_2015-03-05-10-07-12
Changed to SHA512
screenshot_2015-03-05-10-07-53
screenshot_2015-03-05-10-08-24

dschuermann:
This is not the right solution. It required much more thinking than just replacing the variables. Maybe it was too complicated for a simple patch. I did it now in https://github.com/open-keychain/open-keychain/commit/c121657c2cf17ecd3d59809ff86f758b7b1a592c

Issue #1106, opened 2015-03-05
EmailEditText ToDos


dschuermann:
Currently the code to verify emails and show an indicator and the code to check the password strength is in CreateKeyInputFragment. The email field is also used in AddUserIdDialogFragment and the password strength indicator should also be used in SetPassphraseDialogFragment (is currently not used there).
We have some code duplication here.
It would be nice to have all code and xml related to the fields in their own classes named PasswordEditText and EmailEditText.

chipset95:
Implemented in pull request #1107. Please have a look! :)

dschuermann:
Still some leftovers:
* Move adapter code from CreateKeyInputFragment into EmailEditText
* Create NameEditText with adapter code for name autocomplete
* Move password strengthview code into PasswordEditText

dschuermann:
* Use hack from AutoCorrectAutoCompleteTextView in EmailEditText and NameEditText

chipset95:
On it!

chipset95:
Since AutoCorrectAutoCompleteTextView is not being used once NameEditText is implemented, should I remove it?

dschuermann:
yes

dschuermann:
Password strengthview code has been merged into PasswordEditText by https://github.com/open-keychain/open-keychain/pull/1111

What is left:
* Move adapter code from CreateKeyInputFragment into EmailEditText
* Create NameEditText with adapter code for name autocomplete
* Use hack from AutoCorrectAutoCompleteTextView in EmailEditText and NameEditText

chipset95:
@dschuermann I already did that in #1109

chipset95:
@dschuermann should I remove PasswordEditText, since we have PassphraseEditText now?

dschuermann:
yes

chipset95:
@dschuermann Does this count in One Patch Rule for GSoC?

dschuermann:
sure

PR #1107, opened 2015-03-06
EmailEditText, PasswordEditText issue #1106

chipset95:
Created EmailEditText extending AutoCompleteTextView with TextWatcher to validate email.
Created PasswordEditeText extending EditText with TextWatcher & Wrapper for PasswordStrengthView.
Implemented them in CreateKeyInputFragment, AddUserIdDialogFragment, SetPassphraseDialogFragment along with required XML changes.
All done in accordance with issue #1106 

EmaiEditText with Suggestions
PasswordEditText with passphrase strength indicator
PasswordEditText with passphrase strength indicator in SetPassphraseDialogFragment
EmaiEditText with Suggestions in AddUserIdDialogFragment

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyInputFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyInputFragment.java
@@ -20,9 +20,6 @@ package org.sufficientlysecure.keychain.ui;
 import android.content.Context;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.text.Editable;
-import android.text.TextWatcher;
-import android.util.Patterns;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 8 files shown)


dschuermann:
Looks great!
Please add the GPLv3 license header like done in all other OpenKeychain classes.

chipset95:
Added in commit [7c2fefe](https://github.com/chipset95/open-keychain/commit/7c2fefe2f7b9f6f7d770e37041336ae563518e93)

Valodim:
from a quick glance this code crashes with a nullpointer if a PasswordEditText is changed without having a passwordstrengthview associated to it. Is there a reason PasswordStrenthView isn't just integrated with PasswordEditText? As it is, PasswordEditText is just an EditText with a (non-optional) afterTextChanged Listener but doesn't really add functionality of its own...

codestyle notes: fields don't have an m prefix, local variables should positively never have an m prefix.

dschuermann:
@Valodim np, I merged too fast :/ should have left the PR open. My fault

chipset95:
@Valodim `PasswordStrengthView` doesn't have any code of its own in the mentioned fragments. All we have to do is create the corresponding xml in the layout, reference it in java and attach the `EditText` to it.

I could check for null and prevent the crash, or get the logic of `PasswordStrengthView` to `PasswordEditText` itself. I believe checking for null would be better because if in future a `PasswordEditText` without the strength indicator is needed, it'll be easier this way.

adithyaphilip:
I happened to have created a standalone PassphraseEditText for my earlier PR (https://github.com/open-keychain/open-keychain/pull/1075). I've raised a new PR for it at https://github.com/open-keychain/open-keychain/pull/1111
I guess that would solve the issue raised by @Valodim ?

PR #1108, opened 2015-03-06
Asks Passphrase for key delete/export in ViewKeyActivity.

tdjogi010:
Asks Passphrase for key delete/export in ViewKeyActivity.


Valodim:
Looks good

PR #1109, opened 2015-03-06
Added NameEditText, improved EmailEditText

chipset95:
As requested in #1106 , improvement in EmailEditText (created in #1107) and creation of NameEditText.

Issue #1053, opened 2015-02-23
Menu item: Import my key


dschuermann:
Re-add menu item in key list: "Import my key"

chipset95:
Import to be added in all the places where export option is present or only at one particular place?

chipset95:
@dschuermann I've sent a PR for this at #1113

PR #1110, opened 2015-03-06
Menu item: Import my key

chipset95:
Added `Import from file` and `Import all keys` option in `ViewKeyActivity` and `KeyListFragment` respectively. #1053 

Check for `null`, add `m` prefix for non static, non public objects in the new `EditText` classes. #1107 

This pull request consists of some commits from #1109 because it is yet to to merged/closed.


chipset95:
@dschuermann Please close this too, I'll put a new PR after rebase.

PR #1113, opened 2015-03-09
Added in two overflow menus

chipset95:
Added Import in KeyListFragement and ViewKeyFragment overflow menus #1053


dschuermann:
This is currently on hold because I was thinking about other concepts of managing private keys (and importing them.)

chipset95:
Okay.

dschuermann:
Implemented a different concept.

PR #1117, opened 2015-03-09
#1042: Replaced ints with enums 

danielnelz:
Created enums IOType in KeychainIntentService and MessageStatus in KeychainIntentServiceHandler. Fixed  three typos in the aforementioned classes, too.


dschuermann:
Great thanks!

Issue #1058, opened 2015-02-26
preselect key for certification if only one is available with certify flag


Valodim:
right now in the 'certify' view, a key for certification is only preselected if the user only has a single secret key. if the user has more than one key but only one is actually suitable for certifying, that one should be selected.

this is slightly complex in the class structure because it involves the CertifyKeyFragment, CertifyKeySpinner and KeySpinner. The check for certification suitability should be in CertifyKeySpinner (it's a certification thing after all), but the mechanism for pre-selection should probably be in KeySpinner.

tdjogi010:
Hey @Valodim 
I sent a pull request #1119 
Plz check it.

dschuermann:
@tdjogi010 Please read again: " The check for certification suitability should be in CertifyKeySpinner (it's a certification thing after all), but the mechanism for pre-selection should probably be in KeySpinner."

tdjogi010:
@dschuermann @Valodim 
Made changes #1119 .
I hope this is what is required.

dschuermann:
@tdjogi010 sry, but no. The check for certification suitability should be in CertifyKeySpinner. The mechanism for pre-selection should probably be in KeySpinner.

tdjogi010:
@dschuermann @Valodim 
I sent new pull request #1142 
Maybe this is what is required.

tdjogi010:
Hey, @dschuermann @Valodim 
Plz review pull request #1142 and give your advice about it.

PR #1119, opened 2015-03-09
Preselected key for certification.

tdjogi010:
Preselected key for certification.


dschuermann:
See #1058 for my comment

Issue #1120, opened 2015-03-09
System contact not updated on re-importing a key after deleting it


adithyaphilip:
Tested on Android 5.0.2 (Moto G 2nd Gen)
Importing a public key for the first time will introduce a new system contact linked to user who owns the public key. Deleting the key from the app removes the system contact. Importing the key again will not create a new system contact.

I looked into it and it seems to be due to the fact that the RawContact rows we add are read-only and hence can only be "hidden" when deleted normally. This means the row is simply marked with the "deleted" flag and left so, hiding it from the user's view, but not actually removed from the RawContacts table. When we import the public key again, we don't create a new RawContact row since we find a pre-existing one in writeKeysToContacts in ContactHelper, just marked with the "deleted" flag.

Deleting it completely from the user's device requires the CALLER_IS_SYNC_ADAPTER query parameter to be passed to the deletion operation (http://developer.android.com/reference/android/provider/ContactsContract.html#CALLER_IS_SYNCADAPTER). Would this work for us? I've tried it out and it seems to do the job, thought I'm not sure if we depend on the contact not being deleted immediately elsewhere.

PR #1121, opened 2015-03-09
Fixed contact deletion Issue #1120

adithyaphilip:
Refers to https://github.com/open-keychain/open-keychain/issues/1120
Added CALLER_IS_SYNCADAPTER query parameter, with value "true", to all deletion queries. This causes the RawContact to be actually wiped when deleted instead of just the "delete" flag being set. Solves the aforementioned issue of re-imported keys not being linked to system contacts.


dschuermann:
Wow, I searched for hours for this bug and was unable to find it! Awesome work!

Issue #1116, opened 2015-03-09
Rework "Delete after encryption" feature


dschuermann:
Delete after encryption/decryption is currently implemented very badly.
* If the user chooses to encrypt a file and share it, the dialog is shown behind the share menu of Android, thus the file will never be deleted.
* If multiple files are encrypted the dialog is shown for every file separately, also the files are deleted separately.
* The deletion code is inside the DialogFragment.

Should be:
* Show dialog before sharing the encrypted temporary file and only proceed to sharing if "Delete" is clicked
* Execute deletion after encryption process succeeded
* Allow deletion of multiple files, i.e. show a list of files in the DeleteFileDialogFragment dialog. Also show a list of files in the Toasts, collect files that failed to delete.

manojkhannakm:
Hello @dschuermann 

By "show a list of files in the DeleteFileDialogFragment dialog" do you mean to show the list of files as Checkboxes so as to allow the user to the select the files to delete or just the list of file names.

dschuermann:
Just a list of file names

manojkhannakm:
Pull request: https://github.com/open-keychain/open-keychain/pull/1124

Also while testing I've found a bug. When selecting files for encryption, if a same file is selected multiple times, while deleting it causes a crash of activity. Should this be handled while selecting the files or while deleting.

manojkhannakm:
Also while testing I've found a bug. When selecting files for encryption, if a same file is selected multiple times, while deleting it causes a crash of activity. Should this be handled while selecting the files or while deleting.

manojkhannakm:
@dschuermann Please have a look at the updated pull request https://github.com/open-keychain/open-keychain/pull/1137.

PR #1124, opened 2015-03-10
Fix for "Delete after encryption" Issue #1116

manojkhannakm:


dschuermann:
This is only a small part of the issue #1116, please read it carefully. Also please rebase.

manojkhannakm:
I think I've fixed everything that you've mentioned.

> * If the user chooses to encrypt a file and share it, the dialog is shown behind the share menu of Android, thus the file will never be deleted.

To fix this I've created a callback class OnDeletedListener to listen if the files are getting deleted and then starting the Share activity.

> * If multiple files are encrypted the dialog is shown for every file separately, also the files are deleted separately.

For this I've made the DeleteFileDialogFragment class accept multiple uris as varargs so as to delete multiple files.

> * Show dialog before sharing the encrypted temporary file and only proceed to sharing if "Delete" is clicked
* Execute deletion after encryption process succeeded
* Allow deletion of multiple files, i.e. show a list of files in the DeleteFileDialogFragment dialog. Also show a list of files in the Toasts, collect files that failed to delete.

And the above fixes are completely based on the flow which you've stated above.

If I'm still missing out something, please let me now.

dschuermann:
@manojkhannakm I annotated the line.

manojkhannakm:
My bad, thanks for pointing that out.

Also while testing I've found a bug. When selecting files for encryption, if a same file is selected multiple times while deleting it causes a crash, this is because the second time it tries to delete the file it is not found. Should this be handled while selecting the files(like restricting selection of same file) or while deleting(removing duplicate uri).

dschuermann:
Yes I think this should be handled when selecting the file. Simply don't allow the selection of the same file again. We should also catch the problem in the deletion code and handle it like a file that could not be deleted.

manojkhannakm:
Ok. So how should the users be warned when they add duplicate files, a toast or using the Notify class.

dschuermann:
I would prefer the Notify class.

manojkhannakm:
Updated pull request https://github.com/open-keychain/open-keychain/pull/1136

Issue #1126, opened 2015-03-10
Use contact name instead of user id on linked system contact card


Valodim:
see title. right now, the primary user id is shown as contact name, when it should be the merged contact's name

Sharad001:
Sir can you tell me where can i find the contact card code as it is very large code-base so please suggest how to find it.

Sharad001:
Sir can you please tell me where is this linked contact card has been used. It would be helpful if you could give some screenshot for good understanding

dschuermann:
Alread done in https://github.com/open-keychain/open-keychain/pull/1134

Issue #1125, opened 2015-03-10
Only sync confirmed keys to contacts


Valodim:
Showing key ids of non-confirmed keys in contacts provides a false sense of security. Just like contact pictures are only shown for confirmed keys, raw contacts should only be made for confirmed keys.

this implicitly solves #1122 iinm

adithyaphilip:
On it :)

PR #1130, opened 2015-03-10
Use contact name instead of user id Issue #1126

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1126, apologies for the oversight


adithyaphilip:
Opening new PR solving https://github.com/open-keychain/open-keychain/issues/1125 too

PR #1132, opened 2015-03-10
Fix #1129

### LINKED ISSUE #1129: Add link to FAQ page to in APP help

vectorijk:
Fetch FAQ page from open-keychain.github.io

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -22,6 +22,11 @@ file_filter = OpenKeychain/src/main/res/raw-<lang>/help_certification.html
 source_file = OpenKeychain/src/main/res/raw/help_certification.html
 source_lang = en
 
+[open-keychain.help-faq]
+file_filter = OpenKeychain/src/main/res/raw-<lang>/help_faq.html
+source_file = OpenKeychain/src/main/res/raw/help_faq.html
+source_lang = en
+
 [open-keychain.help-start]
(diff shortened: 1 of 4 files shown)


dschuermann:
I would agree on this pull request, but I am not sure how to handle updates to the website in the future. Do we then always manually pull the FAQ page into OpenKeychain? Maybe it is okay for now to simply add a link that opens the browser into the Help screen. Although, I prefer for a future version that help is also available offline inside OpenKeychain.

vectorijk:
Sorry reply this late. You may know, I am in the tough final week.
I thought about it. It might be pretty complicated if we fetch content from
website many time. My idea is write a script to handle this or we can find
some tool convert markdown page from static blog frame into a simple single
html page.
But I not sure what you mean adding a link into Help screen. Do you mean
open a link inline help page and open it in browser.

On Thu, Mar 12, 2015 at 5:21 AM, Dominik Schürmann <notifications@github.com

vectorijk:
I thought another solution. That is we can generate second edition of page without header and footer , just plain html when we generate static blog pages. After that we just need to copy generated pages into resource directory in project. That will be simpler.

dschuermann:
We are now using markdown for in-app help. This makes copying versions of the FAQ around much easier. Also changelogs are now easier to maintain. See https://github.com/open-keychain/open-keychain/commit/b3a599f214489ba9edce5b8c3f56b3f671499e86

Issue #997, opened 2014-11-06
Crash when clicking NFC button


dschuermann:
Reported by mail:
SGS2 crashes when clicking on NFC button (SGS2 does not have NFC)

PR #1133, opened 2015-03-11
Check if NFC is available on device

Rrego6:
Possible Fix for Issue #997 
(May revise to hide NFC button instead)


dschuermann:
Thanks. Please use our Notify class the next time. I fixed this in https://github.com/open-keychain/open-keychain/commit/27027b8ee55dd0245632bca6b647477d810ce077

Issue #958, opened 2014-10-09
Link all "my keys" to the PROFILE contact in Android


Magissia:
Own key isn't linked to "myself" in Contact app like is "myself" Google+ account on AOSP contact app. 

The "myself" contact entry do have the same mail address and name than the Open-keychain key have.

dschuermann:
@adithyaphilip You did great work on the other sync related problems. Would you like to look into this? Should be easy:  Link all "my keys" from OpenKeychain to this one "me" contact without doing email checks.

adithyaphilip:
Yup, I'm on it! :)

PR #1134, opened 2015-03-11
Sync only confirmed keys and change 

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1125 and https://github.com/open-keychain/open-keychain/issues/1126. 

TODO: cause displayed system contact to be updated on confirmation of a key from KeyViewFragment, possibly by introducing a cursor returned by loader methods for system contact information, which would allow automatic updating on change in the contacts table?


dschuermann:
Great thanks!

dschuermann:
@adithyaphilip I have another issue for you if you are interested :D https://github.com/open-keychain/open-keychain/issues/958

PR #1137, opened 2015-03-11
Updated fix for Issue #1116

manojkhannakm:
Please have a look at the updates. I've fixed the handling of adding duplicate files too.

screenshot_2015-03-11-19-30-59

screenshot_2015-03-11-19-36-38

screenshot_2015-03-11-19-29-29

PR #1138, opened 2015-03-11
Delete after encryption feature, fixes #1116

### LINKED ISSUE #1116: Rework "Delete after encryption" feature

algui91:
I think I got it. Hope it is what you want.


dschuermann:
Thanks for your work. Looks already good, but this issue has been fixed already in https://github.com/open-keychain/open-keychain/pull/1137

algui91:
Opps, didn't saw it was closed. Best regards!

Issue #1115, opened 2015-03-09
No progress shown when encrypting a file with ascii armor


dschuermann:
Hangs at 0% when encrypting a file and enabling ascii armor in the overflow menu, but after some time succeeds.

Sharad001:
Sir i would like to fix this bug and i'm working on it !!

Sharad001:
firstly i checked that enablearmor is returning false in the PgpSignEncryptOperation. !!

PR #1143, opened 2015-03-12
Fix for Issue #1115

manojkhannakm:
Please use files of size 1mb+ while testing so that the progress can be viewed clearly.

This fixes enables the change of progress for both Encryption of Files(with and without ASCII armor) and Text.

screenshot_2015-03-13-01-02-00


dschuermann:
Very good! Thanks for your work!

PR #1144, opened 2015-03-12
Own keys linked to "me" contact

adithyaphilip:
Refers to https://github.com/open-keychain/open-keychain/issues/958

Own keys have been linked to the "me" contact and the linked system contact code has been updated to handle this. Tested it and seems to be working :)

Some functions added like insertMainProfileRawContact, getMainProfileMasterKeyIds and deleteMainProfileRawContactById have resulted in a certain amount of code duplication since they only differ from their normal RawContact counterparts by the Uri being queried. Should we create a base function which these functions call by passing different Uris?


dschuermann:
Great contribution!
I think we can leave the code duplication for now, not sure if it makes things more readable when we introduce more parameters, but good suggestion!

Issue #1051, opened 2015-02-22
Show "my keys" even when no secret key is in our database


dschuermann:
Show "my keys" even when no secret key is in our database. Instead of items show a button to import your key from a file.

dschuermann:
Related: https://github.com/open-keychain/open-keychain/issues/1052

imjalpreet:
@dschuermann, in this you actually a want a change in the key_list_fragment.xml, right?

dschuermann:
I want a change in KeyListFragment.
When no secret key is available in db, but public keys are existing, the "My Keys" header should still be shown and instead of items we should show a button to import keys from a file.

imjalpreet:
So, there should be no change in the key_list_fragment.xml? I thought that when there are no keys and it shows No keys found!!, we should replace that by A button to import keys from file. Is it this way or am I understanding it a bit wrong?

Valodim:
this is about secret keys specifically. if a user has only public keys in his key list, there is no indicator that he has no secret ones because the "My Keys" section is not even shown.

not entirely sure how to handle this correctly. either use a mergecursor and put in a fake item, or use addHeaderView of the list. hmmm.

imjalpreet:
Can you tell how can I create a public key and test this?

dschuermann:
no :)

sdipendra:
Created a pull request for the above at #1145, if you can review it and tell any changes to made.

Pushpalanka:
Is this issue in-progress with you imjalpreet?

dschuermann:
For clarification: This issue higly depends on https://github.com/open-keychain/open-keychain/issues/1052
So no need to work on it until we use the RecyclerView.

dschuermann:
Done in https://github.com/open-keychain/open-keychain/commit/b2d447ab32c3e3327c0b3a89f51c46ede520a6d1 and https://github.com/open-keychain/open-keychain/commit/7004d129a8546977c450bd40931b0b774cd2025e

PR #1146, opened 2015-03-13
Implemented #1051

sdipendra:
screenshot_2015-03-14-03-59-22
If any changes regarding the aesthetics please let me know.


dschuermann:
This is not what the issue is about.
This would not be displayed when public keys are in the db.
Pleas re-read the issue.

PR #1147, opened 2015-03-15
-fixed out of bounds crash when retrieving the main profile name with se...

Hiperzone:
...cret keys

-fixed a possible crash when retrieving the main profile contactid


dschuermann:
thanks!

PR #1153, opened 2015-03-16
Add enums to int constants in AppsListFragment and ContactHelper classes

ChanakaCooray:
use enums instead of int constants to AppsListFragment.java and ContactHelper.java classes


dschuermann:
Makes no sense to use enums for query indexes, sry.

PR #1154, opened 2015-03-16
Removed unnecessary @TargetApi annotation in ContactHelper

adithyaphilip:
Removed unnecessary @TargetApi(Build.VERSION_CODES.JELLY_BEAN) by calling the query() function which does not accept a cancellation signal (API level 1), since we don't use it anyway.


dschuermann:
Great! thanks!

dschuermann:
Hmmm
Is there anything left that prevents us from removing the check in https://github.com/open-keychain/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java#L122 ?

adithyaphilip:
Nope, nothing :) tested it out a bit, seems to be working fine. I've tried it out on a Kindle Fire (ICS 4.0.3 - API level 15, the lowest we support), and it works perfectly. Submitted a PR at https://github.com/open-keychain/open-keychain/pull/1155

PR #1159, opened 2015-03-18
Switch to SuperSLiM

vectorijk:
This is first step switch to SuperSLiM.
I looked around the whole project. There will be many thing to improve during this switching.
TODO:
- [ ] Multiple Selection
- [ ] issue: transparent of header.
- [ ] more optimization.
- [ ] deploy to ViewKeyAdvCertsFragment

@dschuermann I just want to make sure whether this switch does work and whether this switching would be what you want.

output


dschuermann:
You should definitly use some sort of CursorAdapter instead of an ArrayList! For sections we would definitly like to use the SectionLayoutManager of SuperSLiM.

vectorijk:
Okay, working on it.

PR #1160, opened 2015-03-18
Show HKP server results first in key cloud search. Fixes #1066

### LINKED ISSUE #1066: Main email in search changes randomly from keybase and keyserver result
1. Do many searches for one fingerprint
2. email address randomly changes from keybase.io/dschuermann and the real email

jellysheep:
Previously, the cloud key search showed either a keyserver email or a keybase ID as the user's main email, depending on which server request was answered faster.
Now results returned from HKP keyservers will be sorted in front of keybase IDs to show the correct email address in the dialog.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/ImportKeysList.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/ImportKeysList.java
@@ -70,6 +70,7 @@ public class ImportKeysList extends ArrayList<ImportKeysListEntry> {
             modified = true;
         }
 
+        boolean incomingFromHkpServer = true;
         // we’re going to want to try to fetch the key from everywhere we found it, so remember
         //  all the origins
         for (String origin : incoming.getOrigins()) {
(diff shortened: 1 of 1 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/ImportKeysList.java
```diff
@@ -78,13 +79,19 @@ private boolean mergeDupes(ImportKeysListEntry incoming, ImportKeysListEntry exi
...
+                incomingFromHkpServer = false;
             }
         }
+
         ArrayList<String> incomingIDs = incoming.getUserIds();
         ArrayList<String> existingIDs = existing.getUserIds();
         for (String incomingID : incomingIDs) {
             if (!existingIDs.contains(incomingID)) {
-                existingIDs.add(incomingID);
+                if (incomingFromHkpServer) {
```
dschuermann: Please add a comment here why this is done

jellysheep: User IDs / email addresses returned from HKP servers are preferred over those from other (keybase.io) servers.
This prepends HKP server results to the start of the list, so that the UI (for cloud key search), picking the first list element, shows the right email address.

dschuermann: :) I mean "Add a comment to the code"

jellysheep: Sorry, it's not my day. :) Fixed.

dschuermann:
Great thanks!

Issue #1152, opened 2015-03-16
NullPointerException when signing with YubiKey


dschuermann:
Sign via API and Yubikey key: 
```
21053               Keychain  D  signImpl
 21053               Keychain  D  java.lang.NullPointerException: Attempt to invoke virtual method 'long java.lang.Long.longValue()' on a null object reference
 21053               Keychain  D  at org.sufficientlysecure.keychain.pgp.PgpSignEncryptOperation.execute(PgpSignEncryptOperation.java:492)
 21053               Keychain  D  at org.sufficientlysecure.keychain.remote.OpenPgpService.signImpl(OpenPgpService.java:290)
 21053               Keychain  D  at org.sufficientlysecure.keychain.remote.OpenPgpService.access$100(OpenPgpService.java:65)
 21053               Keychain  D  at org.sufficientlysecure.keychain.remote.OpenPgpService$1.execute(OpenPgpService.java:795)
 21053               Keychain  D  at org.openintents.openpgp.IOpenPgpService$Stub.onTransact(IOpenPgpService.java:69)
 21053               Keychain  D  at android.os.Binder.execTransact(Binder.java:446)
 20022               Keychain  E  onError getErrorId:0
 20022               Keychain  E  onError getMessage:Attempt to invoke virtual method 'long java.lang.Long.longValue()' on a null object reference
```

kentnek:
This is due to `input.getSignatureSubKeyId()` (which is a `Long`) can be `null`. Can I return a `PgpSignEncryptResult` with `PgpSignEncryptResult.RESULT_ERROR` as the `result`?

Issue #1071, opened 2015-03-02
invoke nfc in advanced key


Valodim:

dschuermann:
Add a simple button somewhere that executes the invokeNfc command from the advanced key view. So that also keys that are not yours can be shared via NFC.

sdipendra:
Do you want this button to be visible based on availability of NFC or it should be a permanent?

kentnek:
Do you mean adding a Share via NFC button to the ActionBar of the ViewKeyActivity, to the right of the Encrypt Text button?

dschuermann:
no I mean a button in the advanced key view under the first tab

kentnek:
Oh you mean something like this?

![Screenshot](http://i.imgur.com/PzzyHPh.png)

kentnek:
I have just implemented this feature and made a pull request.

However I'm wondering if both key types (owned and not owned) can be shared via NFC, why not just show the Share via NFC button in `ViewKeyActivity`. Currently its `Visibility` is set to `GONE` if a key doesn't belong to the user.

PR #1162, opened 2015-03-19
Added a blank footer view to the list of key & implemented NFC in Advanced View

kentnek:
When the list of key is long, the Floating Action Button blocks the content of the last few rows. A blank footer (with height `72dp`) is added to the bottom to prevent this (and to follow the [Material Design's Guidelines](http://www.google.com/design/spec/components/buttons.html#buttons-floating-action-button): *the content should have enough padding at the bottom so the button does not block content in last row.*)

Screenshot

I've also fixed Issue #1152 by checking if `input.getSignatureSubKeyId()` is not `null` before proceeding.

Moreover, I have implemented feature #1071 as well.

Screenshot

Okay I should start branching my work.


dschuermann:
Pleas split the pull request in separate ones :) so we can discuss them one by one

kentnek:
Okay!

PR #1163, opened 2015-03-19
Added a blank footer view to the list of keys

kentnek:
When the list of key is long, the Floating Action Button blocks the content of the last few rows. A blank footer (with height `72dp`) is added to the bottom to prevent this (and to follow the [Material Design's Guidelines](http://www.google.com/design/spec/components/buttons.html#buttons-floating-action-button): *the content should have enough padding at the bottom so the button does not block content in last row.*)

Before & After: 

Screenshot Screenshot

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
@@ -169,6 +169,22 @@ public class KeyListFragment extends LoaderFragment
         mStickyList.setDrawingListUnderStickyHeader(false);
         mStickyList.setFastScrollEnabled(true);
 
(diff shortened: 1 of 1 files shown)


dschuermann:
I am not convinced of this change. The vanilla AOSP contact app also simply has a floating button above the list. Actually I don't know any app that introduces a footer here.

kentnek:
The floating button blocks the right section (most likely confirmed/unconfirmed indicator) of the very last item if there is no bottom padding. The padding is only used when the list has a finite number of item, for example our list of keys. 

Apps with lists containing a finite number of items do this. Google Drive app is an example: there's always padding at the bottom (if not the last row would touch the bottom edge of the screen, and it would be impossible to click on the item since the floating button would have blocked it).

dschuermann:
:) You are right!

kentnek:
Thanks for merging! :D

PR #1164, opened 2015-03-19
Implemented #1071: Invoke NFC in Advanced View

kentnek:
For Issue #1071:

 - Extracted NFC code from `ViewKeyActivity` to `NfcHelper` to share code between classes
 - Changed the private anonymous `Handler` for NFC into a static private subclass, that uses `WeakReference` to avoid memory leaks
 - Added resources needed (retrieved from Graphics) for the NFC button inside `ViewKeyAdvShareFragment`.
 - Fixed the ripple boundary of the `Share With..`. button to prevent it from bleeding to other buttons on the right (UX improvement)

image

However I'm wondering if both key types (owned and not owned) can be shared via NFC, why not just keep the `Share via NFC` button in the ActionBar of `ViewKeyActivity`. Currently its `Visibility` is set to `GONE` if a key doesn't belong to the user.


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/NfcHelper.java
```diff
@@ -0,0 +1,204 @@
+package org.sufficientlysecure.keychain.util;
```
dschuermann: Please add the GPLv3 header here like in the other classes

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/NfcHelper.java
```diff
@@ -0,0 +1,204 @@
...
+import org.sufficientlysecure.keychain.provider.KeychainContract;
+import org.sufficientlysecure.keychain.provider.ProviderHelper;
+import org.sufficientlysecure.keychain.ui.util.Notify;
+
+import java.lang.ref.WeakReference;
+
+/**
+ * This class contains NFC functionality that can be shared across Fragments or Activities.
+ * <p/>
+ * Created on Mar 20, 2015.
```
dschuermann: Remove "created" and author here

kentnek:
Oh okay

kentnek:
I'm not really sure about the GNU header though, so may I put in my name inside as well like the one I have just committed?

dschuermann:
Yes exactly, the GNU header with your name is enough.

kentnek:
Ok thanks a lot!

dschuermann:
Regarding:
> However I'm wondering if both key types (owned and not owned) can be shared via NFC, why not just keep the Share via NFC button in the ActionBar of ViewKeyActivity. Currently its Visibility is set to GONE if a key doesn't belong to the user.

We don't want to encourage the sharing of keys that do not belong to you, but we want to allow it via the advanced view for advanced users. Average users should only share their own keys.

kentnek:
Oh i see.

dschuermann:
Forgot to add: Thanks for your work ;)

kentnek:
It's my pleasure :) I also need to complete the one-patch requirement for GSOC xD

PR #1166, opened 2015-03-19
Multi-threaded key import, update all feature

adithyaphilip:
Added:
1. Support for multi-threaded key import when importing from the cloud.
2. An "Update all keys" option in the overflow menu of KeyListFragment.

This was just a quick patch I tried it to see how many threads we could run at a time while importing keys from the cloud. I've used a CachedThreadPool limited to 200 threads as of now, and there seemed to be no issues refreshing 174 keys.

I've introduced the multi-threading in KeychainIntentService currently - which I guess isn't a good idea since IntentServices are not really meant to spawn more threads, as the service dies as soon as it has no more work to do. The spawned threads talk back to the activity via the callbacks, but allowing the "Cancel" intent from the activity to reach the spawned threads required making the AtomicBoolean variable holding the "canceled" state static. This is because the worker thread belonging to the service dies as soon as it finishes spawning the key import threads, and only the static variable can be shared.

The only reason I tried it out in the IntentService was so that we could immediately observe the change on importing keys from the cloud (extremely fast: 173 keys in around 30 seconds).

Should I shift the entire multi-threaded key import to a separate thread/service or leave it here for now? Is there anything else I should be doing to get it right?

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -697,6 +697,9 @@
         <service
             android:name=".service.KeychainIntentService"
             android:exported="false" />
+        <service
+            android:name=".service.CloudImportService"
+            android:exported="false" />
 
         <provider
             android:name=".provider.KeychainProvider"
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportExportOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportExportOperation.java
@@ -168,7 +168,7 @@ public class ImportExportOperation extends BaseOperation {
             return new ImportKeyResult(ImportKeyResult.RESULT_FAIL_NOTHING, log);
         }
 
-        int newKeys = 0, oldKeys = 0, badKeys = 0, secret = 0;
+        int newKeys = 0, updatedKeys = 0, badKeys = 0, secret = 0;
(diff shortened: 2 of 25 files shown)


dschuermann:
Awesome!

> Should I shift the entire multi-threaded key import to a separate thread/service or leave it here for now? Is there anything else I should be doing to get it right?

yes please. Then we can already merge this one!

adithyaphilip:
ProgressDialogFragment and KeychainIntentServiceHandler provided the required functionality but required slight modifications to work with the new Service. I've created their copies with the CloudImport prefix.

Multi-threaded import seems to be working as of now! :)

dschuermann:
I am not quite happy with the code duplication. Looking at ``CloudImportProgressDialogFragment``: The only difference is that ``ACTION_CANCEL`` is send to ``CloudImportService``? If yes, this could easily been done by using an argument in ``newInstance``. Same goes for ``CloudImportServiceHandler``: What did you change? Could this be done by an argument or by class inheritance?

adithyaphilip:
I apologise, there was absolutely no need for all that code duplication. Thankyou for pointing that out! :)
All we had to do was pass an extra parameter to `ProgressDialogFragment` telling it which Service it was working for. With `CloudImportProgressDialogFragment` gone, there's no need for `CloudImportServiceHandler` either. I've removed them both.

dschuermann:
Great work @adithyaphilip and thanks for working on the comments I made!

adithyaphilip:
Thankyou! :)

PR #1171, opened 2015-03-22
Clear passphrase cache 

### LINKED ISSUE #1169: Passphrase cache not clearing
Cached pass at 15:30
Still cached at 21:15

Setting: 3 min cache time

Maybe we are using imprecise alarm settings. This needs more investigation. At most times, it is working.

Sharad001:
Fixes #1169

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
@@ -93,7 +93,7 @@ public class PassphraseCacheService extends Service {
     public static final String EXTRA_MESSENGER = "messenger";
     public static final String EXTRA_USER_ID = "user_id";
 
-    private static final int REQUEST_ID = 0;
+
     private static final long DEFAULT_TTL = 15;
 
     private static final int NOTIFICATION_ID = 1;
@@ -314,7 +314,9 @@ public class PassphraseCacheService extends Service {
     private static PendingIntent buildIntent(Context context, long keyId) {
         Intent intent = new Intent(BROADCAST_ACTION_PASSPHRASE_CACHE_SERVICE);
         intent.putExtra(EXTRA_KEY_ID, keyId);
-        return PendingIntent.getBroadcast(context, REQUEST_ID, intent,
+        //request code should be unique for each call for pending intent so we have taken requestcode
+        //as keyId which is unique for each call as keyId is unique for each passphrase
+        return PendingIntent.getBroadcast(context, (int)keyId, intent,
                 PendingIntent.FLAG_CANCEL_CURRENT);
     }
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/CertifyResult.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/CertifyResult.java
@@ -120,7 +120,7 @@ public class CertifyResult extends OperationResult {
                     mCertifyError, mCertifyError);
         }
 
-        return Notify.createNotify(activity, str, duration, style, new ActionListener() {
+        return Notify.create(activity, str, duration, style, new ActionListener() {
             @Override
             public void onAction() {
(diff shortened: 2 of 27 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
```diff
@@ -314,7 +314,8 @@ public void onReceive(Context context, Intent intent) {
     private static PendingIntent buildIntent(Context context, long keyId) {
         Intent intent = new Intent(BROADCAST_ACTION_PASSPHRASE_CACHE_SERVICE);
         intent.putExtra(EXTRA_KEY_ID, keyId);
-        return PendingIntent.getBroadcast(context, REQUEST_ID, intent,
+        Long _id = System.currentTimeMillis();
+        return PendingIntent.getBroadcast(context, _id.intValue(), intent,
```
Valodim: Does this fix the problem? I don't see how the requestId parameter you changed is related, could you please explain?

Sharad001: Sir actually pending indent requires request-code so what happens is when two pending indents query with same request code then one which is later gets the response or i mean to say the later one is deleted from cache but not the former one because the request code has been overwritten as request code was harcoded to 0.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
```diff
@@ -314,7 +314,8 @@ public void onReceive(Context context, Intent intent) {
     private static PendingIntent buildIntent(Context context, long keyId) {
         Intent intent = new Intent(BROADCAST_ACTION_PASSPHRASE_CACHE_SERVICE);
         intent.putExtra(EXTRA_KEY_ID, keyId);
-        return PendingIntent.getBroadcast(context, REQUEST_ID, intent,
+        Long _id = System.currentTimeMillis();
```
dschuermann: please use ``long`` instead of ``Long`` and follow the code style guidelines (this means naming the var ``requestCode`` instead of ``_id``
Then also add a comment to the code indicating why this request id needs to be unique.

dschuermann:
Instead of ``System.currentTimeMillis();`` we should also use ``keyId``. This is a better indicator if a particular PendingIntent is already there.

Sharad001:
Sir, sorry for not following coding style next time i will keep in mind !!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
```diff
@@ -314,7 +314,10 @@ public void onReceive(Context context, Intent intent) {
     private static PendingIntent buildIntent(Context context, long keyId) {
         Intent intent = new Intent(BROADCAST_ACTION_PASSPHRASE_CACHE_SERVICE);
         intent.putExtra(EXTRA_KEY_ID, keyId);
-        return PendingIntent.getBroadcast(context, REQUEST_ID, intent,
+        //request code should be unique for each call for pending intent so we have taken requestcode
+        //as keyId which is unique for each call as keyId is unique for each passphrase
+        long requestcode = keyId;
```
dschuermann: this line is redundant, you can use keyId directly

dschuermann:
After fixing the last comment from me, please squash the commits via git.

Sharad001:
Sir,sorry because of some issue with my git setup i also squashed the @manojkhannakm  commit . Should i squash my changes again ?

dschuermann:
Not sure what you did, but the outcome should be one commit in this pull request. Please read some git manuals for help :)

Sharad001:
Sir, basically you want my commits to be squashed in one right !!

dschuermann:
yes

dschuermann:
I did this now in https://github.com/open-keychain/open-keychain/commit/aa3565b857d088be3d12c1fb9acf052cc6cb7720
Thanks for fixing this issue. This counts as a one-patch

Issue #1173, opened 2015-03-23
Input of passphrase for symmetric encryption causes crash


Jesperbk:
When trying to encrypt a message using symmetric encryption (passphrase), the application crashes while entering the first passphrase. 

Someone may want to verify this on a physical device, as I only have access to an emulator at the moment.

As I am looking into this project for GSoC, I will take a stab at a patch myself.

Jesperbk:
I will close the issue, as a fix has been merged.

PR #1174, opened 2015-03-23
Fixed symmetric encryption crash

Jesperbk:
This is a fix for #1173.

During comparison of the two passphrase fields, the activity passphrase is set to null, if the two fields do not match. This causes an issue when entering the second character, at which point the `removeFromMemory()` method is called on the null passphrase. 

An empty passphrase would have been an elegant solution, but the input validation logic expects the passphrase to be null if the inputted passphrases do not match. So a simple null check in the setter will have to do.


dschuermann:
thanks!

Issue #1045, opened 2015-02-20
ask for passphrase on secret key export/deletion


Valodim:
Although these operations do not require a passphrase cryptographically, it still makes sense to make sure the user is authorized if they want to use those operations. See #194 for discussion.

tdjogi010:
Hey @Valodim ,
I send a pull request #1108 .
Please check it.

Valodim:
Good so far. This is still missing other places where keys can be deleted though, like the KeyList and AdvancedKeyView.

On the long run, I'd prefer to have the check in the operation code where it really belongs, rather than just the UI. It is the UI code's obligation to inquiry the passphrase for these operations, but the Delete and Export operations should not just trust that this check was already made at an earlier point.

manojkhannakm:
Hello @Valodim

I've planned to solve this issue by creating a DeleteKeyDialogActivity(similar to PassphraseDialogActivity) which would start the PassphraseDialogActivity if the key is not cached and then display the DeleteKeyDialogFragment. Is this approach good?

But I've also found that the above approach cannot not be applied to exporting as it uses a common FileDialogFragment. Can you please suggest something on this.

And what should happen when users try to export multiple secret keys? should they be asked for the passphrases of all keys(which I believe would be a bad UX design) or should they be simply not allowed to export.

iseki-masaya:
Hello @Valodim

I send a pull request #1182.
Please take a look.

PR #1182, opened 2015-03-24
#1045: Add a loop for multiple key export.

iseki-masaya:
Please take a look.


dschuermann:
* Please don't use short if else constructs. Always use brackets.
* Can this somehow be externalized into another class? This is some code scattered around ;)

iseki-masaya:
Thank you for your review.
I modify the code : 
* Add brackets if-else constructs.
* Externalize the procedure as the function named showMultiExportDialog.

I think it is not reasonable to externalize the procedure as another class in this stage because this procedure is used in only KeyListFragment.
This procedure should be externalized as another class when it needs to be used from other class.

dschuermann:
Great thanks!

Issue #1052, opened 2015-02-22
Switch from StickyListHeaders to SuperSLiM


dschuermann:
https://github.com/TonicArtos/SuperSLiM

vectorijk:
Hi @dschuermann @Valodim 
To import SuperSLiM as an extern submodule, do we need to fork https://github.com/TonicArtos/SuperSLiM to open-keychain. 

Because I add 
```
[submodule "extern/SuperSLiM"]
	path = extern/SuperSLiM
	url = https://github.com/TonicArtos/SuperSLiM.git 
``` 
into `.gitmodules` and update with `git submodule update --init --recursive` . *Does not work*

I read the [Readme in SuperSLiM] (https://github.com/TonicArtos/SuperSLiM#how-do-i-get-it)
It should be used SuperSLiM as
 ```
dependencies {
    compile 'com.tonicartos:superslim:+'
}
```

So which way should we use? extern submodule way or the way above?

dschuermann:
https://github.com/open-keychain/open-keychain#add-new-library
Add as Maven dependency, but please use a fixed version number, not "+". Also add it to gradle witness verification.

vectorijk:
Cool! Get it

vectorijk:
@dschuermann , I got stuck. In original KeyListFragment, Its use cursor to fetch key data from database. And use override method in StickyListHeaders [getHeaderView(...)](https://github.com/open-keychain/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java#L772) to create Header View. But in sample of library SuperSLiM may use RecyclerView adapter to create key ViewHolder. And its data needed to display is stored in ArrayList in RecyclerAdapter (see [this](https://github.com/TonicArtos/SuperSLiM/blob/master/example/src/main/java/com/tonicartos/superslimexample/CountryNamesAdapter.java#L40)). Should we store all the keys fetched from cursor into a new arraylist in RecyclerAdapter first and then to create ViewHolder or something else?

Cause now I can apply library SuperSLiM into KeyListFragment and have already display key list(RecyclerView) well without sticky header. But I get confused about how to add header into key list (RecyclerView). So to store header and key view holder together into an arraylist, I think it will be easy to handle.

Valodim:
[...]

PR #1185, opened 2015-03-27
Switched to SuperSLiM and Fixed issue #1051

manojkhannakm:
Fixed issues #1052 And #1051 and also made few improvements like showing total key count during searching etc. The only thing which is missing currently is the fast scroll feature which is not available for RecyclerView.

**KeyListFragment**
screenshot_2015-03-28-00-50-46
screenshot_2015-03-27-20-58-24

**ViewKeyAdvCertsFragment**
screenshot_2015-03-28-13-55-13


dschuermann:
Looks great from looking at the UI.

The main problem with this pull request is the inefficiency of the code. Before discussing smaller details, there should be no need to maintain
```
private ArrayList<Item> mItemList = new ArrayList<>();
```
to hold all items. As done before all of this should be handled by the adapter. 
Everything else follows from this fundamental design decision.

manojkhannakm:
Actually I started maintaining an ArrayList of Items only after carefully analyzing both `StickyListHeaders` and `SuperSLiM` libraries.

The `StickyListHeaders` uses something called a `headerId` which is an integer with which the headers and its items are sorted ascendingly and also the headers are separate views which are not a part of the list.

Where as `SuperSLiM` uses a `firstPosition` for every item which is the index of its header item. The problem here is that the headers are also items of the RecyclerView which makes the `position` of the Key items different from their `cursorPositon`. This why I had to maintain an `Item` class with some minimum variables like `type`, `headerIndex`(`firstPosition`) and `cursorPosition`.

dschuermann:
Hmmh I see. This class is definitely complicated. Thanks for your explanation and your hard work on this.
Can you also look over this @Valodim ?

Valodim:
Looked it over: This looks pretty good overall, except for the cursor->arraylist hack, which is unfortunately a showstopper. I looked around a bit, there is [this library](https://github.com/twotoasters/SectionCursorAdapter/tree/52ee5a69265db095d6023f145245ce19128402fd/library/src/main/java/com/twotoasters/sectioncursoradapter/adapter) which implements (if I'm not mistaken) a sectioning recycler cursor adapter. This might be a perfect solution for this case, provided the code in that branch is mature enough.

dschuermann:
Hey, @manojkhannakm . This pull request was not completely bad, are you working on a new pull request or why did you close it?
If possible please re-push the branch.

manojkhannakm:
It got closed automatically when I reset the branch, but I've already made another branch(development-old) with these changes which I'll improve with what @Valodim has suggested later.

I'm currently learning more on unit test now so I'll fix this up a bit later.

dschuermann:
@manojkhannakm Wow great :+1:

dschuermann:
Hey @manojkhannakm ,
do you still have your branch laying around? It would be great if you could push it to github. We would then merge it into a branch on our side and use it as a starting point for further improvements.

manojkhannakm:
@dschuermann https://github.com/manojkhannakm/open-keychain/tree/development-old

Its there in the last 4 commits.

dschuermann:
Thanks. It's now here: https://github.com/open-keychain/open-keychain/tree/recyclerview-wip

Issue #1167, opened 2015-03-20
Design icon for passphrase cache


dschuermann:
We should design two icons for the passphrase cache service notification:
* one for Android < 5
* one for Android > 5
based on the design guidelines for the versions and based on our new app icon.

mrCoder256:
Good afternoon!
@dschuermann , I am going to solve this issue but I have one question: are "the icons for the passphrase cache service notification" that show after a new key is created?

dschuermann:
Yes.
The icon is also shown everytime something is signed using a key.
Please use https://github.com/open-keychain/open-keychain/blob/development/Graphics/drawables/material-launcher/vector-src.ai or https://github.com/open-keychain/open-keychain/blob/development/Graphics/drawables/material-launcher/vector-src.psd as a basis.

mrCoder256:
@dschuermann, I've designed two icons for the passphrase cache and each of them is shown for appropriate Android version (> 5 and < 5).
But the problem is that I cannot push my changes to github :( I'm still getting an error:
"Compressing objects: 100% (3156/3156), done.
error: RPC failed; result=55, HTTP code = 200"
I hope I'll be able to send my pull request to the end of this working day...

dschuermann:
No problem. Looks like github is under DDoS today.

dschuermann:
Looks linke github ddos stopped. You can try again

mrCoder256:
@dschuermann, I've sent you pull request: https://github.com/open-keychain/open-keychain/pull/1186
I've got Internet access only now :(

mrCoder256:
Please, tell me if it is necessary to make some changes in these icons. I put the original images in Graphics/drawables/passphrase-cached-icons/ folder, so it wont be difficult to change these icons.

I haven't found how to apply a proposal on the Google Summer of Code page, so I wrote a message to your organization. Was I late with my proposal?

dschuermann:
@mrCoder256 The Google Summer of Code application closed yesterday. You can apply next year and consider applying a minimum of a day before the deadline next time :)

The icon design should follow http://developer.android.com/design/style/iconography.html#notification , i.e., we need a white icon. So sry I need to close your pull request.

dschuermann:
done in https://github.com/open-keychain/open-keychain/commit/8cd7e295404fdfad5167b96fdf3a5cbeca8e4cd1

PR #1186, opened 2015-03-27
Designed icons for passphrase cache

mrCoder256:
As "one patch rule"


dschuermann:
Sry but not following the design guidelines. See https://github.com/open-keychain/open-keychain/issues/1167#issuecomment-87262940

PR #1187, opened 2015-03-27
-Fixed a situation where additional emails could be duplicated when crea...

Hiperzone:
...ting a new key -If CreateKeyEmailFragment view is recreated, the array of additional emails won't be repopulated again if its not null, avoiding new reallocations. -if CreateKeyEmailFragment view is recreated, the email adapter wont be recreated if its not null, avoiding new reallocations.


dschuermann:
Great thanks!

PR #1188, opened 2015-03-28
Corrected import error message, changed style to WARN

adithyaphilip:
If an import operation for multiple keys fails for all, we get an incorrect "Import of %d keys failed!" message. Just had to add a missing parameter.
For reference:
Old:
screenshot_2015-03-28-19-56-42

New:
screenshot_2015-03-28-21-33-35

A comment recommended switching to the warning style in case some keys succeed but others fail. Changed that too.


dschuermann:
Nice catch!

Issue #1177, opened 2015-03-23
ASCII armor encryption very slow


dschuermann:
From ml:
> If I set EXTRA_REQUEST_ASCII_ARMOR true, it take *significantly *longer to 
> to return from the async encryption service call -- maybe too long for me 
> to use that feature.

I also noticed it when encrypting a file. Choosing ascii armor slows it extremly down.

Maybe a buffer is missing somewhere.

PR #1191, opened 2015-03-30
Added Buffer to ASCII Armor Encryption

adithyaphilip:
2^16 bytes buffer is in place, solved Issue https://github.com/open-keychain/open-keychain/issues/1177

Should I add a buffer to the signature writing too?


dschuermann:
Awesome, I implemented your fix in https://github.com/open-keychain/open-keychain/commit/1923942fc97534eb86177053a29868df5223cb16

PR #1200, opened 2015-04-14
Fix for NFC signing operations: move PIN verify inside sign/decrypt operation and set correct mode.

josecastillo:
According to the OpenPGP spec, PW1 must be submitted with mode 81 for a signing operation, and mode 82 for everything else. OpenKeychain currently verifies the PW1 with mode 82 all the time. OpenPGP-compliant cards will not issue a signature when the PW1 is submitted with mode 82. 

The reason this hasn't been broken is that the only widespread OpenPGP device with NFC, the Yubikey NEO, had a bug that allowed signatures to be generated without a verified PIN. Yubico has fixed this bug as of today, and issued [a security advisory](https://github.com/Yubico/ykneo-openpgp/blob/master/doc/SecurityAdvisory%202015-04-14.txt). New Yubikeys NEO will throw an error when signing is attempted after VERIFY PW1 with mode 82. 

This patch moves the PIN verification from the ```handleNdefDiscoveredIntent``` into a new method called ```nfcVerifyPIN```, which takes as a parameter the mode of operation. The new method is called from the sign and decrypt methods, so that the PIN can be submitted with the proper mode for the requested operation. If the PIN verification fails, an IOException is thrown, which is caught and passed to ```handleNfcError```. 

I've tested this on a Nexus 7, which is the only NFC-equipped Android device I have on hand, but I would welcome additional testing.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
@@ -185,25 +185,6 @@ public abstract class BaseNfcActivity extends BaseActivity {
             throw new IOException("Initialization failed!");
         }
 
-        if (mPin != null) {
-
-            byte[] pin = new String(mPin.getCharArray()).getBytes();
-
-            // Command APDU for VERIFY command (page 32)
-            String login =
-                    "00" // CLA
-                            + "20" // INS
-                            + "00" // P1
-                            + "82" // P2 (PW1)
-                            + String.format("%02x", pin.length) // Lc
-                            + Hex.toHexString(pin);
-            if (!nfcCommunicate(login).equals(accepted)) { // login
-                handlePinError();
-                return;
-            }
-
-        }
-
         onNfcPerform();
 
         mIsoDep.close();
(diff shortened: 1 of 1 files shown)


dschuermann:
Thanks for investigating and actually fixing it also on our side!

Valodim:
Thanks for the fix! Seems I missed that part of the spec completely...

Do you happen to have some timing information on the VERIFY call? More specifically, is it worth keeping an expected state, or does the VERIFY operation take negligible time? Depending on the operation done, there may be several nfcCalculateSignature calls and we could avoid subsequent VERIFYs for those

josecastillo:
The verify call is pretty negligible in terms of overhead; I just ran a few tests and found the verify operation taking ~0.09 seconds, compared to ~.64 seconds for a decrypt operation with a 2048-bit key. Still, it might make sense to track the state if only to avoid spamming the PIN over the NFC interface. 

I'd be more than willing to take a crack at implementing this; I'm very familiar with the smart card specification at this point, and I'm actually working on an NFC OpenPGP smart card so it's quite relevant to my current work.

dschuermann:
@josecastillo Contribtions in that area are more than welcome, so if you would like to work on improving the NFC stuff, awesome!
If you have any questions while implementing don't hesitate to ask here via issues or via our mailinglist at http://groups.google.com/d/forum/openpgp-keychain-dev

Issue #1199, opened 2015-04-14
Register on keyserver URLs


dschuermann:
We could handle keyserver urls such as: https://hkps.pool.sks-keyservers.net/pks/lookup?op=get&search=0x73EE2314F65FA92EC2390D3A718C070100012282

So when users click on a webpage link pointing to a keyserver OpenKeychain should handle the search/import because we are able to really verify downloaded keys by their fingerprint.

For now it should be enough to register on sks-keyservers.net only. A perfect implementation would register on all keyserver urls configured inside OpenKeychain.

PR #1207, opened 2015-04-19
support for handling keyserver urls from browser, added documentation

adithyaphilip:
Takes care of Issue https://github.com/open-keychain/open-keychain/issues/1199.

It's been written to handle all keyserver URLs based on the presence of /pks/lookup in the path, as specified in http://tools.ietf.org/html/draft-shaw-openpgp-hkp-00#section-3, not just https://hkps.pool.sks-keyservers.net or the ones the user has added on Openkeychain.

I've enabled searching through keybase.io by default when we handle a keyserver search from a URL.

A possible addition would be:
When the user clicks on a URL and it's a keyserver he hasn't registered on Openkeychain yet, we offer the option to add it to his list of keyservers.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -39,6 +39,7 @@ import org.sufficientlysecure.keychain.ui.base.BaseNfcActivity;
 import org.sufficientlysecure.keychain.service.CloudImportService;
 import org.sufficientlysecure.keychain.service.ServiceProgressHandler;
 import org.sufficientlysecure.keychain.ui.dialog.ProgressDialogFragment;
+import org.sufficientlysecure.keychain.ui.util.FormattingUtils;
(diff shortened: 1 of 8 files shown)


dschuermann:
Besides the usage of Toast, the rest looks great!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -200,7 +207,29 @@ protected void handleActions(Bundle savedInstanceState, Intent intent) {
...
+                startListFragment(savedInstanceState, null, null, null, null);
+                break;
+            }
+            case ACTION_SEARCH_KEYSERVER_FROM_URL: {
+                // need to process URL to get search query and keyserver authority
+                String query = dataUri.getQueryParameter("search");
+                String keyserver = dataUri.getAuthority();
+                // if query not specified, we still allow users to search the keyserver in the link
+                if (query == null) {
+                    Notify.create(this, R.string.import_url_warn_no_search_parameter, Notify.LENGTH_INDEFINITE,
```
adithyaphilip: fixed :)

dschuermann:
Awesome!

Issue #1198, opened 2015-04-14
Import from QR Code: Fail gracefully if not on keyserver


dschuermann:
Currently the import from QR Codes fails if the internet connection is unstable or the key is not on a keyserver. A simple Toast is shown with "Import operation failed!". This should be handled better!

```
Keychain  D  scanned: openpgp4fpr:54de5aff80f43e882c6355efb8843ba9d9add3f6
 23048               Keychain  D  Bundle EXTRA_DATA:
 23048               Keychain  D  ------------------------------
 23048               Keychain  D  import_key_list : [org.sufficientlysecure.keychain.keyimport.ParcelableKeyRing@38dafb5b]
 23048               Keychain  D  import_key_server : hkps://hkps.pool.sks-keyservers.net
 23048               Keychain  D  ------------------------------
 23048               Keychain  D  Send message by setProgress with progress=0, max=100
 23048               Keychain  V  log: LogEntryParcel{mLevel=START, mType=MSG_IMPORT, mParameters=[1], mIndent=0}
 23048               Keychain  V  log: LogEntryParcel{mLevel=DEBUG, mType=MSG_IMPORT_KEYSERVER, mParameters=[hkps://hkps.pool.sks-keyservers.net], mIndent=1}
 23048               Keychain  V  log: LogEntryParcel{mLevel=INFO, mType=MSG_IMPORT_FETCH_KEYSERVER, mParameters=[0xb8843ba9d9add3f6], mIndent=2}
 23048               Keychain  D  hkp keyserver get: /pks/lookup?op=get&options=mr&search=0x54de5aff80f43e882c6355efb8843ba9d9add3f6
 23048               Keychain  D  hkp keyserver query: https://hkps.pool.sks-keyservers.net:443/pks/lookup?op=get&options=mr&search=0x54de5aff80f43e882c6355efb8843ba9d9add3f6
 23048               Keychain  E  query failed
 23048               Keychain  E  org.sufficientlysecure.keychain.keyimport.Keyserver$QueryFailedException: not found
 23048               Keychain  E  at org.sufficientlysecure.keychain.keyimport.HkpKeyserver.get(HkpKeyserver.java:345)
 23048               Keychain  E  at org.sufficientlysecure.keychain.operations.ImportExportOperation.importKeyRings(ImportExportOperation.java:220)
 23048               Keychain  E  at org.sufficientlysecure.keychain.operations.ImportExportOperation.importKeyRings(ImportExportOperation.java:136)
[...]

PR #1209, opened 2015-04-20
Make Logs not Toasts - Fixes Issue #1198

adithyaphilip:
Fix for Issue https://github.com/open-keychain/open-keychain/issues/1198.
It seems that `returnResult` was used to imply two things:
1. Whether the fingerprint should just be scanned and returned (not imported).
2. Whether the activity was started from inside OpenKeychain or externally.

Due to 1., importing a key via QR Code meant `returnResult` was set to false. Due to 2., it incorrectly implied that this was an external request. Since external requests can't be expected to interpret our logs, we were showing a toast.

To separate the two, an extra `EXTRA_IS_INTERNAL` is passed to `ImportKeysProxyActivity` whenever it is started from within OpenKeychain and a log is to be returned.

I've also renamed `returnResult` to `onlyReturnFingerprint` to reflect the difference between the two.

From what I've seen, `returnResult` (now `onlyReturnFingerprint`) seems to be equivalent to `ACTION_SCAN_WITH_RESULT`. Should we do away with the variable completely and  just check for the action?

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
@@ -58,12 +58,11 @@ import java.util.Locale;
 public class ImportKeysProxyActivity extends FragmentActivity {
 
     public static final String ACTION_QR_CODE_API = OpenKeychainIntents.IMPORT_KEY_FROM_QR_CODE;
+    // implies activity returns scanned fingerprint as extra and does not import
     public static final String ACTION_SCAN_WITH_RESULT = Constants.INTENT_PREFIX + "SCAN_QR_CODE_WITH_RESULT";
     public static final String ACTION_SCAN_IMPORT = Constants.INTENT_PREFIX + "SCAN_QR_CODE_IMPORT";
 
     public static final String EXTRA_FINGERPRINT = "fingerprint";
-
-    boolean returnResult;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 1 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
```diff
@@ -177,7 +184,8 @@ private void processScannedContent(Uri uri) {
     }
 
     public void returnResult(Intent data) {
-        if (returnResult) {
+        boolean isInternal = getIntent().getBooleanExtra(EXTRA_IS_INTERNAL, false);
```
dschuermann: I think this can be replaced by a check if the activity has been started by ACTION_SCAN_WITH_RESULT or ACTION_SCAN_IMPORT, because only ACTION_QR_CODE_API is used from the outside.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
```diff
@@ -540,6 +540,7 @@ private void searchCloud() {
     private void scanQrCode() {
         Intent scanQrCode = new Intent(getActivity(), ImportKeysProxyActivity.class);
         scanQrCode.setAction(ImportKeysProxyActivity.ACTION_SCAN_IMPORT);
+        scanQrCode.putExtra(ImportKeysProxyActivity.EXTRA_IS_INTERNAL, true);
```
dschuermann: this can then be removed again

dschuermann:
> From what I've seen, returnResult (now onlyReturnFingerprint) seems to be equivalent to ACTION_SCAN_WITH_RESULT. Should we do away with the variable completely and just check for the action?

yes sounds good.

adithyaphilip:
done :)

dschuermann:
great work as always, thanks for fixing!

adithyaphilip:
Thank you.

PR #1210, opened 2015-04-20
Maintain client-side state for PIN verification operations

josecastillo:
This change adds three fields to BaseNfcActivity to track PIN verification. Two booleans track whether the PIN is verified for a given operation, and a third tracks whether the PIN must be resubmitted for every signature operation (page 32 of OpenPGP smart card spec). It also adds a method to query the PW status bytes for this information. 

This pull request also incorporates a separate commit with an unrelated fix when selecting the application. A successful SELECT FILE operation is indicated by the response ending with 9000, not being equal to 9000 (page 31 of OpenPGP smart card spec). The standard OpenPGP card includes file control information in the returned data field; this change is necessary to accommodate cards that implement this part of the spec.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
@@ -57,6 +57,9 @@ public abstract class BaseNfcActivity extends BaseActivity {
     public static final int REQUEST_CODE_PASSPHRASE = 1;
 
     protected Passphrase mPin;
+    protected boolean mPw1ValidForMultipleSignatures;
+    protected boolean mPw1ValidatedForSignature;
+    protected boolean mPw1ValidatedForDecrypt; // Mode 82 does other things; consider renaming?
     private NfcAdapter mNfcAdapter;
     private IsoDep mIsoDep;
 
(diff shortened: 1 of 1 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
```diff
@@ -57,6 +57,9 @@
     public static final int REQUEST_CODE_PASSPHRASE = 1;
 
     protected Passphrase mPin;
+    protected Boolean mPw1ValidForMultipleSignatures;
+    protected Boolean mPw1ValidatedForSignature;
+    protected Boolean mPw1ValidatedForDecrypt; // Mode 82 does other things; consider renaming?
```
Valodim: is there a reason that these are Boolean rather than boolean?

Valodim:
looks awesome, thanks for your work! see the remark above, other than that I'll gladly merge

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
josecastillo: Changed the Booleans to booleans. Also realized that I neglected to commit a change to the nfcDecryptSessionKey method, so that is added as well.

josecastillo:
Sorry, I managed to push a mangled commit in there because I was out of sync with upstream. Fixed it, but the build failed due to connection issues with github...

Valodim:
done and done. thanks again.

PR #1212, opened 2015-04-21
Corrected possible null pointer with key import

adithyaphilip:
Should have been keyserver == null instead of keysever != null. Sorry.
Much thanks to @josecastillo for pointing that out.


dschuermann:
thanks!

PR #1219, opened 2015-04-23
Espresso test for CreateKeyActivity

manojkhannakm:
This is my first time writing tests so if I've gone wrong somewhere or left anything important please let me know. Also fixed few minor bugs.

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -11,6 +11,12 @@ dependencies {
     compile 'com.android.support:recyclerview-v7:22.0.0'
     compile 'com.android.support:cardview-v7:22.0.0'
 
+    // UI testing libs
+    androidTestCompile 'com.android.support.test:runner:0.2'
+    androidTestCompile 'com.android.support.test:rules:0.2'
+    androidTestCompile 'com.android.support.test.espresso:espresso-core:2.1'
+    androidTestCompile 'com.android.support.test.espresso:espresso-contrib:2.1'
+
     // JCenter etc.
     compile 'com.eftimoff:android-patternview:1.0.1@aar'
     compile 'com.journeyapps:zxing-android-embedded:2.3.0@aar'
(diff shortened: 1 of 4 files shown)


OpenKeychain/build.gradle
```diff
@@ -135,6 +143,10 @@ android {
     dexOptions {
         preDexLibraries = false
     }
+
+    packagingOptions {
+        exclude 'LICENSE.txt'
```
dschuermann: we don't have a "LICENSE.txt", only "LICENSE".

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
```diff
@@ -125,9 +125,6 @@ public void onClick(View v) {
         // initial values
         if (mAdditionalEmailModels == null) {
             mAdditionalEmailModels = new ArrayList<>();
-            if (mCreateKeyActivity.mAdditionalEmails != null) {
-                mEmailAdapter.addAll(mCreateKeyActivity.mAdditionalEmails);
-            }
```
dschuermann: Please remove these changes from this pull request, they were already fixed in development branch.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
```diff
@@ -137,6 +134,10 @@ public void onClick(View v) {
                     addEmail();
                 }
             });
+
+            if (mCreateKeyActivity.mAdditionalEmails != null) {
+                mEmailAdapter.addAll(mCreateKeyActivity.mAdditionalEmails);
+            }
```
dschuermann: Please remove these changes from this pull request, they were already fixed in development branch.

OpenKeychain/build.gradle
dschuermann: Ah, this also excludes these files from libs we use, right?
Can you explain exactly why this is exclude is added here?

dschuermann:
Looks great so far. Please fix the small issues described above.

dschuermann:
Fixed them myself and pulled. Thanks for your work!

PR #1225, opened 2015-04-25
update lib MaterialDrawer to latest(2.8.2)

vectorijk:
Since the latest version of library MaterialDrawer depends on latest com.android.support:appcompat version 22.1.0 and we switched to new support lib version 22.1 now, we can update this library to the latest.


dschuermann:
thanks

PR #1226, opened 2015-04-26
Improvements to CreateKeyEmailFragment

Hiperzone:


Hiperzone:
Will check tomorrow why the build failed

dschuermann:
Will look into this pull request later, stay tuned.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
```diff
@@ -95,9 +89,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         View view = inflater.inflate(R.layout.create_key_email_fragment, container, false);
 
         mEmailEdit = (EmailEditText) view.findViewById(R.id.create_key_email);
-        mBackButton = view.findViewById(R.id.create_key_back_button);
-        mNextButton = view.findViewById(R.id.create_key_next_button);
-        mEmailsRecyclerView = (RecyclerView) view.findViewById(R.id.create_key_emails);
+        View mBackButton = view.findViewById(R.id.create_key_back_button);
```
dschuermann: if the variables are no longer class variables, remove the prefix ``m``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
```diff
@@ -133,9 +127,10 @@ public void onClick(View v) {
                     addEmail();
                 }
             });
-        }
-        if (mAdditionalEmailModels.isEmpty() && mCreateKeyActivity.mAdditionalEmails != null) {
-            mEmailAdapter.addAll(mCreateKeyActivity.mAdditionalEmails);
+
+            if (mCreateKeyActivity.mAdditionalEmails != null) {
+                mEmailAdapter.addAll(mCreateKeyActivity.mAdditionalEmails);
+            }
```
dschuermann: Why did you remove ``mAdditionalEmailModels.isEmpty()``? I think it prevented a crash so I added it recently.
Also please fix the indent here.
Maybe you should better revert these lines to the previous version.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
```diff
@@ -143,6 +138,68 @@ public void onClick(View v) {
...
+        return !(email.length() == 0 || !emailPattern.matcher(email).matches());
+    }
+
+    /**
+     * Checks for duplicated emails inside the additional email adapter.
+     *
+     * @param email
+     * @return
+     */
+    private boolean isEmailDuplicatedInsideAdapter(String email) {
```
dschuermann: Isn't this already checked by the recycler list? I already get an error when adding a duplicated item in the current development branch.

dschuermann:
Regarding email validation:
I personally don't like validating email addresses with regexp (also https://news.ycombinator.com/item?id=5763327). We already have the small indicator on the right that indicates if something is wrong (okay looks like it's broken right now: https://github.com/open-keychain/open-keychain/issues/1231 :/ )

Hiperzone:
As we discussed earlier there was a case where duplicated emails were allowed so I decided to refactor the whole code.
The code you added was removed because when I changed the code, everything seemed to work fine for all the test scenarios that I could remember (Tested on Kitkat and lollipop devices). The tests included screen rotation and activity recreation with and without email data.

When I coded the new changes I did not have the changes that you made later and since everything was working fine I did not take it into consideration.

With the fixes I applied to createView, the bug you encountered may have been fixed, but instead of doing assumptions, I'm going to do new tests today and also include a few stability fixes to the CloudImportService :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
Hiperzone: Will be done today. my bad :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyEmailFragment.java
Hiperzone: Will do new tests today for the condition but it may be not needed since i changed the code to take null cases into consideration when adding emails from the activity to the adapter.

OpenKeychain/src/main/res/layout/create_key_email_fragment.xml
```diff
@@ -32,6 +32,7 @@
                 android:layout_marginTop="16dp"
                 android:layout_marginBottom="8dp"
                 android:imeOptions="actionNext"
+                android:inputType="textEmailAddress"
```
dschuermann: not needed due to https://github.com/Hiperzone/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/widget/EmailEditText.java#L53

OpenKeychain/src/main/res/layout/add_email_dialog.xml
```diff
@@ -21,6 +21,7 @@
         android:layout_marginTop="16dp"
         android:hint="@string/label_email"
         android:imeOptions="actionNext"
+        android:inputType="textEmailAddress"
```
dschuermann: not needed due to https://github.com/Hiperzone/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/widget/EmailEditText.java#L53

Hiperzone:
Updated, if you need me to do more changes, don't hesitate to tell me :)

PR #1230, opened 2015-04-28
Make Bluetooth sharing send plain text

### LINKED ISSUE #1223: Key shared via bluetooth is encoded in HTML
When I tried to copy my key to my Mac via Bluetooth, the Mac received a file named "bluetooth_content_share.html" with contents that looked like this:

    <html>
    <head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
    </head>
    <body>
    -----BEGIN PGP PUBLIC KEY BLOCK-----<br>
    <br>
    gVMEc6PwyOL63szYP1idG6G2k93azdbrczHK5xDNAxCjYRdZDfYsTEO7s<a href="tel:118">118</a>uEhA<br>
    jzBKVNFQD/B00js83AG97sOHsboc39liyxG8PE4jgiuHIYYSmy2BAJoG8YDxvpuy<br>
    F6E5dBlUCYMDQXF9TqluF1aXDpenYKVr6W4Rckvu1fhB6CM4LIjBuDdQju5EIfJT<br>
    ⋮
    XgwNH97I58vNgTtwueVnM96w2AXGMrfQRMPluGm22GBeF1LxcAJAUiHjvE8VQomZ<br>
    jI/BG0y5OQqsq+GzTNkqtg<br>
    -----END PGP PUBLIC KEY BLOCK-----<br>
    </body>
    </html>

The data presented here is bogus, but the html structure is the same, except that the actual file had no newlines in it, not even a trailing one.  Note that any sequence of 3 or more decimal digits got converted to a "tel" url-scheme link.

I doubt that OpenKeychain is creating this HTML version; I'm guessing it probably needs to tell the Bluetooth sharing facility in Android not to mangle the file like this.

wfaulk:
Fixes #1223

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
@@ -54,6 +54,10 @@ import org.sufficientlysecure.keychain.ui.util.QrCodeUtils;
 import org.sufficientlysecure.keychain.util.Log;
 import org.sufficientlysecure.keychain.util.NfcHelper;
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
 import java.io.IOException;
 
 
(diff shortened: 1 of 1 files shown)


dschuermann:
If we would send a key like this, does this still work with WhatsApp and other messengers? Because we are now sending a file and not text.

wfaulk:
It does not work with WhatsApp.

It turns out that if you share text with `EXTRA_TEXT`, the Bluetooth app explicitly converts it into HTML.  [BluetoothOppLauncherActivity.java, line 254](https://android.googlesource.com/platform/packages/apps/Bluetooth/+/kitkat-release/src/com/android/bluetooth/opp/BluetoothOppLauncherActivity.java#254).  The only `EXTRA_STREAM` `Uri`s that it will accept are file: and content:.  We've found that file: doesn't work for, at least, WhatsApp, which leaves implementing a ContentProvider.

I'm not sure it's worth it to work around a ridiculous Android misfeature.

dschuermann:
@wfaulk This problem can be solved, at least for Android >= 5. If you like to work on this issue even more, please read the whole articel at https://medium.com/@xXxXxXxXxXam/how-lollipops-share-menu-is-organized-d204888f606d and implement Replacement Extras.

PR #1233, opened 2015-04-28
Update README.md with new build info

wfaulk:
I found that the README was not entirely accurate with respect to the current development status and it took me a while to figure out what I needed to get and do to make it work.


dschuermann:
thanks!

PR #1235, opened 2015-04-30
Fix Bluetooth share without breaking others

### LINKED ISSUE #1223: Key shared via bluetooth is encoded in HTML

wfaulk:
Implemented Bluetooth sharing via file while leaving all other sharing alone.

Fixes #1223

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
@@ -26,7 +26,7 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Bundle;
-import android.os.FileObserver;
+import android.os.ParcelFileDescriptor;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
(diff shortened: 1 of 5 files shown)


dschuermann:
Nice work!
Instead of using another caching implementation, please use TemporaryStorageProvider (BTW, which needs to be fixed, see https://github.com/open-keychain/open-keychain/issues/1241).
Also, I would advise that we use the primary user id as a filename instead of the short key id. This is much more descriptive, short key ids serve no real purpose and should not be used anymore, see https://github.com/open-keychain/open-keychain/wiki/OpenPGP-Security#key-ids-arent-displayed .

@wfaulk Do you have time to do these changes? Would be really great!

wfaulk:
Had to rebase to get this to merge.  It should be good, though.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
```diff
@@ -213,13 +219,48 @@ private void share(Uri dataUri, ProviderHelper providerHelper, boolean fingerpri
...
+                // Bluetooth Share will convert text/plain sent via EXTRA_TEXT to HTML
+                // Add replacement extra to send a text/plain file instead.
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                    try {
+                        String primaryUserId = UncachedKeyRing.decodeFromData(content.getBytes()).
+                                getPublicKey().getPrimaryUserIdWithFallback();
+
+                        TemporaryStorageProvider shareFileProv = new TemporaryStorageProvider();
+                        Uri contentUri = TemporaryStorageProvider.createFile(getActivity(),
+                                primaryUserId + Constants.FILE_EXTENSION_PGP_ALTERNATE + Constants.FILE_EXTENSION_ASC);
```
dschuermann: only use one file ending, in this case Constants.FILE_EXTENSION_ASC

dschuermann:
The changes in KeychainApplication and KeyFormattingUtils shouldn't be required any more, please revert them.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
wfaulk: Done.

wfaulk:
Whoops.  Fixed.

dschuermann:
Great thanks!

Issue #1260, opened 2015-05-08
"Clicking on linked contact crashes the app every time"


dschuermann:
Also labled with "help-wanted". Should be feasible for a newcomer to fix this.

From G Play:
"keeps crashing when trying to link to linked in.
Chose "linked system contact" for my own contact card
As I clicked on the button linked contact it stopped the app.
Clicking on linked contact crashes the app every time"

```
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=content://com.android.contacts/contacts/-1 }
at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:1765)
at android.app.Instrumentation.execStartActivity(Instrumentation.java:1485)
at android.app.Activity.startActivityForResult(Activity.java:3736)
at android.app.Activity.startActivityForResult(Activity.java:3697)
at android.support.v4.app.FragmentActivity.startActivityForResult(FragmentActivity.java:820)
at android.app.Activity.startActivity(Activity.java:4007)
at android.app.Activity.startActivity(Activity.java:3975)
at org.sufficientlysecure.keychain.ui.ViewKeyFragment.launchContactActivity(ViewKeyFragment.java:195)
at org.sufficientlysecure.keychain.ui.ViewKeyFragment.access$100(ViewKeyFragment.java:49)
at org.sufficientlysecure.keychain.ui.ViewKeyFragment$3.onClick(ViewKeyFragment.java:167)
at android.view.View.performClick(View.java:4756)
at android.view.View$PerformClick.run(View.java:19749)
at android.os.Handler.handleCallback(Handler.java:739)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:135)
at android.app.ActivityThread.main(ActivityThread.java:5221)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:899)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:694)
```

adithyaphilip:
Reason for crash: Had assumed secret key would always have a raw contact in 'me' profile. Assumption would fail if user imported contact from file (sync was missing) or manually deleted contact.

PR #1270, opened 2015-05-08
Fixed sync on key import, crash for 'me' linked contact

adithyaphilip:
Added:
* Contact sync on key import
* Crash prevention if raw contact for secret key does not exist in 'me' contact
* Delete raw contacts marked for deletion by user

Fixes Issue https://github.com/open-keychain/open-keychain/issues/1260


adithyaphilip:
As of now, if a user does not have a 'me' contact, the raw contact for it will be added but no linked system contact will be displayed since name will be null. Is this okay?

dschuermann:
seems reasonable for me

Issue #1256, opened 2015-05-06
Enhancement: On-Card Key Generation


josecastillo:
Opening this issue because I want to implement this, but I'm unsure whether it's desired functionality. As it stands currently, users can generate an OpenPGP identity within OpenKeychain, but if they want to use a Yubikey or NFC smart card, they must first configure it on a PC. I would like to be able to support people generating their keys securely on a Yubikey or a smart card, using nothing but OpenKeychain. I've started a branch to implement the commands necessary to do this on the card, but have run into some challenges understanding how I can fit these changes into the broader application. 

https://github.com/open-keychain/open-keychain/compare/development...josecastillo:oncardkeygen

On-card key generation works by issuing a GENERATE ASYMMETRIC KEYPAIR command for the correct slot (sign/decrypt/auth). The card generates this key, and then returns TLV data with the public exponent and modulus. From there, OpenKeychain would craft a public key packet with this data, and go about making the certification and subkey binding signatures with the usual nfcCalculateSignature commands. Finally, OpenKeychain would need to set the fingerprint and key generation timestamps on the card, and optionally change the user's PINs from the default. So far I've added commands to handle all of these tasks. 

The challenge is in incorporating this into the broader architecture of identity generation in OpenKeychain. I had imagined adding a parameter to the SubkeyAdd constructor, indicating whether the key should be generated on the card, and then firing off the required NFC operations in PgpKeyOperation. But looking into it further, the PgpKeyOperation creates the key using JcaPGPKeyPair, which always seems to want a private key. With on-card key generation, we won't have that, because the card never discloses private key data. 

Anyway, apologies for the wall of text, but after looking into this for a while I'm wondering whether this is even an enhancement that you would like to see in OpenKeychain. Personally I think it would be a huge usability win — especially for NFC-only tokens that cannot be easily configured on a PC.

Valodim:
This is on our agenda, albeit not with a very high priority (#875). It is definitely a desirable enhancement, if you get it into a working state we will merge for sure.

Have you given just sending the command a shot? Is it reasonably usable via NFC, in terms of required time for key generation etc?

The data flow is indeed complicated for this issue, and I have no advice from the top of my head. I will try to think about this some more when I have the time and come back with a better answer.

josecastillo:
Oh awesome — I hadn't seen that item, sorry for not searching more thoroughly. Good to know that it's something that's on the radar screen. 

[...]

Issue #875, opened 2014-09-22
Yubikey:  key generation (on or off key)


dschuermann:

Valodim:
promotion of public keys to yubikey-attached-secret-keys is done (#1028)

generating keys on the yubikey itself or generating them then moving them to the yubikey is a whole different beast, and requires evaluation if it even works reasonably well via nfc. seems low-prio for the moment

dschuermann:
done in https://github.com/open-keychain/open-keychain/commit/a9c606d49bf7d4e7a30b7265cd1fb04ff1c3c01c

PR #1278, opened 2015-05-12
Support for moving RSA keys to Yubikey or NFC smart card

josecastillo:
This pull request implements an enhancement mentioned in #875: the ability to move a secret key from OpenKeychain to a Yubikey or NFC smart card. As discussed in #1256, on-card key generation is unacceptably slow, but the ability to generate keys in OpenKeychain and move them to a smart card is still desirable. 

Subkeys in the Edit Key activity now support a "Move Subkey to Yubikey / Smart Card" action, which will check that the key is smart card compatible, and launch the NFC Operation activity to move the key to the smart card. The command automatically determines the appropriate slot on the card based on the key flags: SIGN or CERTIFY will go in the signing slot; ENCRYPT will go in the decryption slot; and AUTHENTICATE will go in the authentication slot (unless the subkey also has flags for SIGN or CERTIFY, in which case it will go in the signing slot instead). 

Two new methods in the Base NFC Activity support the keytocard operation: ```nfcPutData```, which is responsible for setting the fingerprints and key generation timestamps, and ```nfcPutKey```, which is responsible for extracting the raw secret key parameters and sending them to the card in an appropriate format. These commands require verification of the Admin PIN, which I've hard-coded to the default "12345678" for now. This makes it simpler to put keys on a new card, but a future enhancement might involve prompting for the Admin PIN as well. I've also added a command, ```nfcModifyPIN```, to change the PIN on a smart card; I didn't add UI for this, but it will probably be necessary for a user to fully customize their smart card or Yubikey with OpenKeychain. 

I also introduce one additional UI tweak in this pull request. In the key creation activity, if the user taps a completely blank smart card (one with no key fingerprints), a dialog is displayed asking the following: "Would you like to generate a smart card compatible key?" If the user says "Yes", a flag is set, and the key generation parameters are modified to generate the following: 

1. an RSA-2048 master key for signing and key certification, 
2. an RSA-2048 subkey for encryption, and
3. an RSA-2048 subkey for authentication. 

I welcome code review and any suggestions for changes or additions. There are a few areas that I have specific questions on, which I'll call out with comments on the pull request.


Valodim:
awesome work so far! :+1:

josecastillo:
Thanks for the suggestions! Input pending results ended up being ideal for this; I was able to move the keytocard process entirely into the PgpKeyOperation, which simplifies the data flow and allows for multiple keys to be moved to the card in a single operation. The edit key UI sets a boolean flag ```mMoveKeyToCard``` in each SubkeyChange to indicate that a given subkey should be moved to a card. Once the NFC activity is complete, the operation grabs the serial number from the returned data, sets ```mDummyDivert```, and completes. 

I still have the concern that I'm using the RequiredInputParcel and CryptoInputParcel to map subkey ID's to card serial numbers, when they were intended to map input hashes to output hashes, but I wonder whether that's more of an issue of how they're named than how they're used; in the end, this works to send a list of data into the NFC operation and get a map back out.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1481,4 +1537,29 @@ private static boolean isDivertToCard(PGPSecretKey secretKey) {
...
     }
 
+    private static boolean checkSmartCardCompatibility(PGPSecretKey key, OperationLog log, int indent) {
+        PGPPublicKey publicKey = key.getPublicKey();
+        int algorithm = publicKey.getAlgorithm();
+        if (algorithm != PublicKeyAlgorithmTags.RSA_ENCRYPT &&
+                algorithm != PublicKeyAlgorithmTags.RSA_SIGN &&
+                algorithm != PublicKeyAlgorithmTags.RSA_GENERAL) {
+            log.add(OperationResult.LogType.MSG_K2C_ERROR_BAD_ALGO, indent + 1);
+            return true;
```
Valodim: true?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/OperationResult.java
```diff
@@ -736,6 +739,11 @@ public void onAction() {
         MSG_EXPORT_LOG_EXPORT_ERROR_FOPEN(LogLevel.ERROR,R.string.msg_export_log_error_fopen),
         MSG_EXPORT_LOG_EXPORT_ERROR_WRITING(LogLevel.ERROR,R.string.msg_export_log_error_writing),
         MSG_EXPORT_LOG_EXPORT_SUCCESS (LogLevel.OK, R.string.msg_export_log_success),
+
+        // NFC keytocard
+        MSG_K2C_ERROR_BAD_ALGO(LogLevel.ERROR, R.string.edit_key_error_bad_nfc_algo),
```
Valodim: these are still part of the modify operation, should be MF

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EditKeyFragment.java
```diff
@@ -415,15 +419,54 @@ public void handleMessage(Message message) {
...
+                            change.mMoveKeyToCard = false;
+                        }
+                        break;
+                    }
+                    case EditSubkeyDialogFragment.MESSAGE_KEYTOCARD: {
+                        CanonicalizedSecretKey.SecretKeyType secretKeyType =
+                                mSubkeysAdapter.getSecretKeyType(position);
+                        if (secretKeyType == CanonicalizedSecretKey.SecretKeyType.DIVERT_TO_CARD ||
+                            secretKeyType == CanonicalizedSecretKey.SecretKeyType.GNU_DUMMY) {
+                            Toast.makeText(EditKeyFragment.this.getActivity(),
```
Valodim: use Snackbar available from our Notify class instead of Toast

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -356,10 +361,16 @@ public PgpEditKeyResult createSecretKeyRing(SaveKeyringParcel saveParcel) {
...
+                                                CryptoInputParcel cryptoInput,
+                                                SaveKeyringParcel saveParcel) {
+        return modifySecretKeyRing(wsKR, cryptoInput, saveParcel, new OperationLog());
+    }
+
+    public PgpEditKeyResult modifySecretKeyRing(CanonicalizedSecretKeyRing wsKR,
             CryptoInputParcel cryptoInput,
-            SaveKeyringParcel saveParcel) {
+            SaveKeyringParcel saveParcel,
+            OperationLog log) {
```
Valodim: log should always be passed together with indent, if it's supposed to be logged as part of another action

Valodim:
I think it's fine, since the data in the RequiredInputParcel is explicitly typed by RequiredInputType. Since that data always looks the same, it's easier to use an enum there rather than subclasses, especially since parcelables require a lot of boilerplate to marshall around. It would probably make sense to change mInputHashes to something like mNfcInputData at this point though.

I added some minor comments here and there but looks great otherwise. I would be very grateful if you could add unit tests to the PgpKeyOperationTest class for the mMoveKeyToCard attribute.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyActivity.java
```diff
@@ -116,23 +122,45 @@ protected void onNfcPerform() throws IOException {
...
+
+            } catch (PgpKeyNotFoundException e) {
+                Fragment frag = CreateKeyYubiKeyImportFragment.createInstance(
+                        scannedFingerprints, nfcAid, userId);
+                loadFragment(frag, FragAction.TO_RIGHT);
+            }
+        } else {
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
+            builder.setTitle("Blank Smart Card / Yubikey Detected")
+                    .setMessage("Would you like to generate a smart card compatible key?")
```
dschuermann: These strings should be added to strings.xml
Maybe, it fits better into the wizard-like process to ask this as part of a fragment instead of a dialog, but this is not important for this pull request. I can also do this later for you.

josecastillo:
Fixed the issues mentioned above, and added a test case for the keytocard functionality. Note that in the PgpKeyOperationTest class, I changed the third test subkey from an ELG-1024 key to an RSA-2048 key, so that I would have a smart card compatible key for the test. This change does not affect any of the other test cases. 

I also changed one of the SubkeyChange constructors to take a moveKeyToCard boolean instead of a dummyDivert byte array, since mDummyDivert only ever gets set during the edit key operation. The constructor was no longer used anywhere except in test cases, so I changed it there. The actual functionality of changing a key to divert-to-card is still tested, in the new keytocard test case. 

Thanks for your help and let me know if there's anything else!

josecastillo:
I can also squash at least the first five commits to clean up the history; let me know if you'd like me to do that.

Valodim:
great! there's some open questions on how particular states of yubikey and secret key are handled which we should think about, for example when uploading a secret key to a non-empty yubikey, and how to connect the create-key flow to the upload one.

but the bulk is there and looks stable enough for testing, I think the rest can all be handled in issues. no need to squash, the history looks fine as it is. if dominik has no objections we can merge this.

PR #1293, opened 2015-05-23
highlight currently selected item in navigation drawer

Yoshi64Bit:


Valodim:
looks good! thanks for your work :+1:

Issue #1221, opened 2015-04-25
Multi-threaded key import by CloudImportService Import fails for keyrings with pub+sec key with same key id


dschuermann:
If we use multi-threaded key import as implemented by CloudImportService the following will happen:

Import fails for keyring with pub+sec key with same key id
- Have a file with separated pub and sec key (as exported by Enigmail for example)
- Import into OpenKeychain

I thought we already fixed this issue at one point in the past, but looks like we somehow re-introduced this bug. So this is a typical case were we _must_ have a unit test.

```
 32096               Keychain  D  insert(uri=content://org.sufficientlysecure.keychain.provider/key_rings/8181922323901784706/user_ids, values=is_revoked=false attribute_data=null is_primary=false master_key_id=8181922323901784706 rank=1 type=null user_id=Dominik Schürmann <d.schuermann@tu-bs.de>)
 32096              SQLiteLog  E  (787) abort at 24 in [INSERT INTO user_packets(is_revoked,attribute_data,is_primary,master_key_id,rank,type,user_id) VALUES (?,?,?,?,?,?,?)]: FOREIGN KEY constraint failed
 32096               Keychain  D  Constraint exception on insert! Entry already existing?
 32096               Keychain  D  android.database.sqlite.SQLiteConstraintException: FOREIGN KEY constraint failed (code 787)
 32096               Keychain  D  at android.database.sqlite.SQLiteConnection.nativeExecuteForLastInsertedRowId(Native Method)
 32096               Keychain  D  at android.database.sqlite.SQLiteConnection.executeForLastInsertedRowId(SQLiteConnection.java:784)
 32096               Keychain  D  at android.database.sqlite.SQLiteSession.executeForLastInsertedRowId(SQLiteSession.java:788)
 32096               Keychain  D  at android.database.sqlite.SQLiteStatement.executeInsert(SQLiteStatement.java:86)
 32096               Keychain  D  at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:1471)
 32096               Keychain  D  at android.database.sqlite.SQLiteDatabase.insertOrThrow(SQLiteDatabase.java:1367)
 32096               Keychain  D  at org.sufficientlysecure.keychain.provider.KeychainProvider.insert(KeychainProvider.java:713)
 32096               Keychain  D  at android.content.ContentProviderOperation.apply(ContentProviderOperation.java:245)
[...]

PR #1306, opened 2015-05-28
Merged KeychainIntentService and CloudImportService into KeychainOperationService

adithyaphilip:
Also fixes Issue https://github.com/open-keychain/open-keychain/issues/1221, by allowing a multi-threaded import only if there are no two keys with the same fingerprint. If there are keys with the same fingerprint, or if the keys are being read from a file, then we go with a serial key import.


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportExportOperation.java
```diff
@@ -376,8 +385,6 @@ public ImportKeyResult importKeyRings(Iterator<ParcelableKeyRing> entries, int n
             log.add(LogType.MSG_IMPORT_ERROR, 1);
         }
 
-        ContactSyncAdapterService.requestSync();
```
dschuermann: Please explain this change

dschuermann:
Did you rename KeychainIntentService to KeychainOperationService? Normally git shows this as a rename. In your PR it is handled as a new file, which makes it difficult to see the changes. Could you use "git mv"? maybe this helps.

dschuermann:
Maybe it helps if you do a new commit that renames KeychainOperationService back to KeychainIntentService and then squash the commits. I'd like to see the real changes.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportExportOperation.java
adithyaphilip: I thought it would be better to sync with contacts at the end of a batch of imports, especially when it comes to multi-threaded, rather than sync after each key. I've mentioned it at the beginning of the function and at line 343.

adithyaphilip:
Done. I've left out a level of indentation (caused by wrapping everything in a Runnable in KeychainIntentService) so the changes can be seen. I'll correct it before the final push.

Valodim:
for the record, git doesn't care about mv command, it works with the files directly. git rm is equivalent to git rm + git add

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeychainIntentService.java
```diff
@@ -61,10 +61,7 @@
 import org.sufficientlysecure.keychain.provider.ProviderHelper;
 import org.sufficientlysecure.keychain.service.input.CryptoInputParcel;
 import org.sufficientlysecure.keychain.service.ServiceProgressHandler.MessageStatus;
-import org.sufficientlysecure.keychain.util.FileHelper;
-import org.sufficientlysecure.keychain.util.InputData;
-import org.sufficientlysecure.keychain.util.Log;
-import org.sufficientlysecure.keychain.util.ParcelableFileCache;
+import org.sufficientlysecure.keychain.util.*;
```
Valodim: please don't use * imports

adithyaphilip: Sorry! Must have crept in when I reformatted code. I'll remove it before the final push.

adithyaphilip:
Should I just refactor and push or was there something wrong with the PR?

Valodim:
looks fine to me

dschuermann:
Not sure why it says that I closed it... maybe I accedentially clicked close. Could you re-open a new PR with our review integrated?

Issue #1310, opened 2015-05-29
Search cloud downloads key from keybase if available not from keyserver


dschuermann:
I just downloaded 54265E8C and it shows in OpenKeychain as been expired. After clicking the refresh button it is again orange. I suspect this happened because keybase.io is used as the download source.
We should always prioritize keyserver download if the key is also available there!

Assigned to @adithyaphilip

PR #1317, opened 2015-05-30
Merge key from Keyserver with that from Keybase instead of replacing

adithyaphilip:
Looked like the logic was in place but we were accidentally replacing the keyserver key with that from kaybase.io, before the merge. Hope I haven't misunderstood anything.

Should solve Issue https://github.com/open-keychain/open-keychain/issues/1310.


dschuermann:
thanks!

PR #1318, opened 2015-05-30
Prevent NaN Progress Dialog on import with no selected keys

adithyaphilip:
Prevents a non-cancelable  progress dialog from appearing if the "Import selected keys" button is pressed after either a cloud search or import from file is performed, but no key is selected.


Valodim:
simple change, looks good, merged.

PR #1319, opened 2015-05-31
use native Fragment class, move to v13 compat library

Valodim:
what the title says. we've been minsdk 14 for a while now, if there is no good reason not to, we might as well move to native Fragment implementation.

this commit should be tested with an older version of android. it works fine on android 5, which is unfortunately the only thing I have available for testing atm.


Valodim:
ok I read up on the subject a little more, this seems like a bad idea now. fragment implementations of older android versions had bugs which are fixed in the support library, so we should stick to that I guess

dschuermann:
Yes, better stick to the support lib fragments and replacements for other classes. I also tried this a while ago and turned out to be problematic on SDK 14.

Issue #1297, opened 2015-05-23
File decryption UI


dschuermann:
This was a GSoC 2015 task that has not been assigned. If someone wants to work on it, we are happy to guide them through the process of implementing this. It is also possible to do small subsets of this task, we are happy for every contribution!

 - [x] **General decryption UI**
Currently the file decryption screen in OpenKeychain only allows to decrypt one file at a time and only allows saving of decrypted files into the filesystem. While this is okay for simple scenarios. The decryption process could result in many files being decrypted. Either when multiple files are shared together using Android's [ACTION_SEND_MULTIPLE](http://developer.android.com/training/sharing/receive.html#handling-content) or when the OpenPGP data structure contains many parts. Thus, the screen should be extended for multiple files. A list of files should be displayed like done in the encryption screen that with the options save/open per file item. To get the right content type and decrypt the filename the new internal DECRYPT_METADATA functionality of OpenKeychain should be used. The real payload should be decrypted when save/open is clicked. Furthermore the recipient's key user id should be displayed inside the item.

 - [ ] **Hidden receipients**
A related problem are [hidden recipients](https://github.com/open-keychain/open-keychain/issues/626): The OpenPGP standard allows to prevent the disclosure for whom a file/message is encrypted for. This feature has already been implemented for encrypt but is currently commented out. It must be implemented for decryption (try decryption with all keys) to be done.

 - [ ] **Detached signatures**
Also "detached signatures" should be supported by the new decryption screens. An internal method has already been done during the work on PGP/MIME. This can be used, but needs a carefully designed UI. This method also needs a UI for the encrypt screen.

 - [ ] **Full error overlay**
If a key is not valid anymore a big overlay is shown in decrypt text UI. This should also be implemented for a new file decryption UI. See https://github.com/open-keychain/open-keychain/commit/0d6d4653b43fcf1389b470afd7364854505d96c2

Issues included in this task:
https://github.com/open-keychain/open-keychain/issues/731
https://github.com/open-keychain/open-keychain/issues/557
https://github.com/open-keychain/open-keychain/issues/950
https://github.com/open-keychain/open-keychain/issues/626

Valodim:
related implementation detail: display of the file structure and actual decryption of the files are separate cryptographic operations. the DecryptFragment should internally cache the CryptoInputParcel, to make sure the user doesn't have to input his password or yubikey twice for what is perceived as a single operation in the ui.

crosser:
[...]

PR #1324, opened 2015-06-02
multi decrypt

Valodim:
don't merge yet, work in progress. this is a follow-up to #1297.

this has become sort of a larger project than I expected, keeping this pull request around to track progress.

- [x] CachingDataDecryptorFactory, get rid of RuntimeException based control flow and allow caching of session keys
- [x] DecryptVerifyResult now includes a CryptoInputParcel which can be used to reproduce its results without further user input
- [x] split up decryptfilesfragment into input and listing
- [x] decrypted files are listed in the new fragment rather than just saved/shared
- [x] display encryption and signature status individually
- [x] icon displayed for mime-type
- [x] OpenPgpMetadata includes accurate filesize - this is not working completely for the metadata-only call yet! (not fixed, but no longer relevant because I don't use this call anymore)

todos:
- [x] backend: support using cached session keys for non-divert-to-card keys (obsolete)
- [x] cache decrypted data for send intent
- [ ] properly handle unknown signatures (requires re-decryption!)
- [x] options in file context menu (view log, save to file, delete encrypted file)

for later:
- multi-select
- uri read permission handling


Valodim:
current wip screenshot

Valodim:
- move to card layout
- drop "decrypt metadata" step. we need to decrypt to verify the signature anyways, might as well save the result and re-use it later

Valodim:
rebased on instrumentation branch (and thus recent master)

Valodim:
tons of progress!

- single pass decryption, cache decrypted data (way easier data flow, but might be up for debate)
- prefer temporary storage provider over data in parcels for most operations
- unified clipboard handling: decrypt from clipboard as from file, encrypt to clipboard stores URIs which are resolved in-time
- got rid of api < 11 clipboard compatibility layer
- thumbnails for decryted data

Valodim:
ah yeah, build fails because some of the public intents moved around. need to discuss how to handle these changes... are there apps which rely on DecryptTextActivity behavior?

dschuermann:
@Valodim I don't know of any published apps that rely on this behaviour. But if you change this, please update the wiki and the demo app at https://github.com/open-keychain/openkeychain-intents

Valodim:
- saving done & tested
- deletion done & tested
- preferences tested
- switched to indeterminate progress
- open file logic moved to encryptdecryptoverviewfragment

PR #1327, opened 2015-06-03
Prevent NullPointerException due to Activity death

adithyaphilip:
If the activity is killed, and then resumed, `savedInstanceState` will not be null, but the fragment reference will be, causing a `NullPointerException` at `mListFragment.loadNew` on search.


dschuermann:
@Valodim Could you look into this one? You did most of the recent fixes in this area.

Valodim:
Sorry, I forgot this one.

I'm fine with the change itself, but would rather see if we moved away from storing references to fragments at all, like I did in a couple of activities already. Instead, the reference should be obtained from the FragmentManager every time. This happens only in interactive places (like importKeys()), so if the FragmentManager is null, or if the Fragment can't be obtained, we can assume the Activity is not in a state where we should do anything, and would probably crash if we proceeded with the stored reference.

PR #1329, opened 2015-06-05
added proxy/tor preferences

adithyaphilip:
@dschuermann  thought I'd put up the preferences part of it for review while I finish up the rest. As of now:
* Clicking on 'Enable Tor' will check for Orbot being installed, and if it isn't, prompts the user to install it from Google Play. This is also grays out the 'Enable other proxy' option.
* Clicking on 'Enable other proxy' will similarly gray out 'Enable Tor'.
* Editing Proxy Host will check only for host being empty.
* Editing Proxy Port will check if its a number and 0 <= port <= 65535.
* To be added: Proxy type i.e SOCKS/HTTP

Screenshots:
screenshot_2015-06-05-06-01-11
screenshot_2015-06-05-06-06-19
screenshot_2015-06-05-06-08-38
screenshot_2015-06-05-06-13-16


dschuermann:
Hm, I don't really like the design in OrbotHelper and TorServiceUtils. Because we are already forking them, we could change them the way we like.
Currently thinking about alternatives...

adithyaphilip:
Thank you, will change appropriately.

dschuermann:
I looked a little bit around and came to the following conclusion:
* Let's use netcipher as a library, but only to have these two files OrbotHelper and TorServiceUtils. I forked it into this place: https://github.com/open-keychain/NetCipher so you can include this as a submodule. We should do it this way to get (security-) updates for OrbotHelper and TorServiceUtils. The small overhead by this lib is worth the easier updating.
* We could implement showDownloadDialog etc ourself to have a nicer integration.
* As discussed over Jabber, let's use OkHttp, Netcipher only for these two files
* There is a pull request at https://github.com/anthonycr/Lightning-Browser/pull/249 which is interesting. It integrates I2P besides existing Tor support in Lightning Browser. We could do the same if you like by also including the i2p client lib (see pr).

adithyaphilip:
As of now:
* Proxy support is in place in code in most places. It's currently enabled for Cloud Import. Tested Tor, HTTP proxy (couple of proxies I found online, some were too slow for us, but the ones that respond in less than 30 secs work) and SOCKS proxy (tested by manually entering `localhost` as Proxy Host and port `9050` - `SOCKS` proxy for Tor - couldn't find one that worked online), all working.
* Certificate Pinning - `OkHttp` Certificate Pinning works only if the existing `TrustManager` also trusts the certificate ("Note about self-signed certificates" at the end of http://square.github.io/okhttp/javadoc/com/squareup/okhttp/CertificatePinner.html). Since we have to define a new `TrustManager` anyway, we can make it trust only the certificate for that URL, the way we did in `openCaConnection` in `TlsHelper` earlier. This means we don't have to use `OkHttp` for pinning.
* Adding NetCipher as a submodule resulted in the spongycastle.core library being included twice, so I had to modify it at https://github.com/open-keychain/NetCipher/pull/1, thought I'm not sure if I've done it right.

Todo:
* Add our own DialogFragments to OrbotHelper, including one with three buttons for "Would you like to install/enable Orbot - `Cance`l, `Don't use Tor this time`, `Install/Enable`. Should we do this or make the middle button `Open Settings` and let him turn off Tor there instead?
* Add the new DialogFragments in OpenKeychain code where required.
* Do a final sweep of code to ensure all network operations support proxy.

I've tested the proxy support with a temporary `tempIpTest` function in `HkpKeyserver` to log our IP. 
Also, `OkHttp` was a great library to work with, didn't have any of the issues we were facing earlier.

I've checked out the PR for i2p, and it looks simple enough to integrate. We could definitely add it along with Tor and normal proxying.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeychainService.java
```diff
@@ -653,7 +665,7 @@ public void setPreventCancel() {
     }
 
     public void serialKeyImport(ArrayList<ParcelableKeyRing> keyList, final String keyServer,
-                                ProviderHelper providerHelper) {
+                                ProviderHelper providerHelper, Proxy proxy) {
         Log.d(Constants.TAG, "serial key import starting");
         ParcelableFileCache<ParcelableKeyRing> cache =
                 new ParcelableFileCache<>(mKeychainService, "key_import.pcl");
```
Valodim: not directly related to this commit, but you should use KeychainService.this in these places instead of a local variable

adithyaphilip: Thank you, I didn't know that. I think I did it just because it was shorter (how it's just four more characters strikes me now). I'll revert it to how it was earlier.

adithyaphilip:
Proxy support is active and working in most places. Things that are left:
* Modify our fork of `KeybaseLib` to accept proxies.
* To add proxy support to the `resolve` function of `HkpKeyServer`, we need to modify our `minidns` client to use a proxy. Is that the way to go or should we look at an alternative library?
* i2p support

dschuermann:
Regarding minidns: Yes, that's the way to go. Would the lib then depend on OKHttp? Could we make minidns such that it uses UrlConnection and still allow proxies without introducing a dependency? We could then get these changes upstream.
Regarding the rest: ack

dschuermann:
Regarding the design of the Orbot dialog: Feel free to propose a design. We can then try out how it feels from a UI flow and change accordingly.

dschuermann:
Please rebase your branch on to of current master. Travis CI should then go green. If not, you should investigate the problems.

adithyaphilip:
minidns uses `DatagramSocket` and I was a little confused about adding proxy support to it. 
The links: http://stackoverflow.com/questions/4569597/java-datagramsocket-and-proxy-firewall and http://stackoverflow.com/questions/18428498/sending-udp-packets-through-socks-proxy seemed most relevant out of the lot. I looked at http://www.interhack.net/projects/deschall/archive/199705/0489.html but I wasn't exactly sure what it did. I'll keep trying.

I think I could get rid of the OkHttpClient for KeybaseLib too, using the workaround at https://github.com/guardianproject/NetCipher/pull/24 for android < 19.

I2P I'm not sure how well it's working. It was timing out consistently on both OpenKeychain and Lightning Browser (it did connect once for Lightning Browser though). Also they haven't published an I2P app on the play store yet so the pressing the install button on the I2P popup just takes you to a "No results" screen on Google Play. Should we leave this for later?

Tor/normal proxy support is fully functional and active except for Certify operations because of issue https://github.com/open-keychain/open-keychain/issues/1345. The dialog we're showing right now looks like:
screenshot_2015-06-13-02-55-47

dschuermann:
Regarding proxy support: Please contact the minidns people, especially @mar-v-in (he is currently working as a GSoC student on minidns, so he should know the details).

Regarding KeybaseLin: What workarounds do you have in mind exactly? KeybaseLib uses HttpURLConnection as far as I see. So what workarounds do we need? Maybe we can reduce the amount of changes needed in KeybaseLib to a minimum by using "okhttp-urlconnection", see http://square.github.io/okhttp/javadoc-urlconnection/index.html . But this needs to be tried, I am especially skeptic if this allows proxy support.

Regarding I2P: Let's put the I2P commits in a different branch for now.

mar-v-in:
A future version of minidns will contain an API to inject arbitrary data sources. This includes the possibility to use a socks proxy. If you're interested in that API take a look at my dev branch [1], the relevant code resides in `minidns-core/src/main/java/de/measite/minidns/source`

[1] https://github.com/mar-v-in/minidns/tree/my-dev/

dschuermann:
@mar-v-in great work!
@adithyaphilip You should have access to our fork of minidns. Simply add @mar-v-in 's repo as a new remote and pull his branch into our fork. Then use it :)

settings.gradle
```diff
@@ -8,3 +8,4 @@ include ':extern:spongycastle:prov'
 include ':extern:minidns'
 include ':extern:KeybaseLib:Lib'
 include ':extern:safeslinger-exchange'
+include ':OpenKeychain-Test'
```
Valodim: is this intended?

adithyaphilip: Nope, not sure how it crept in. I'll remove it in the next commit.

adithyaphilip:
Finished cleaning everything up. Should be ready for a merge after testing now.

dschuermann:
I added two small commits.

Please fix these issues:
* No text is displayed in the dialog when tor is not installed
* When clicking on install it crashes and I get:
```
 29530             Keychain D  D  Prompting to install Tor
 29530                 Bundle  W  Key title expected String but value was a java.lang.Integer.  The default value <null> was returned.
 29530                 Bundle  W  Attempt to cast generated internal exception:
 29530                 Bundle  W  java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
 29530                 Bundle  W  at android.os.BaseBundle.getString(BaseBundle.java:921)
 29530                 Bundle  W  at org.sufficientlysecure.keychain.ui.dialog.PreferenceInstallDialogFragment.onCreateDialog(PreferenceInstallDialogFragment.java:95)
 29530                 Bundle  W  at android.app.DialogFragment.getLayoutInflater(DialogFragment.java:398)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:894)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1067)
 29530                 Bundle  W  at android.app.BackStackRecord.run(BackStackRecord.java:834)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1452)
 29530                 Bundle  W  at android.app.FragmentManagerImpl$1.run(FragmentManager.java:447)
 29530                 Bundle  W  at android.os.Handler.handleCallback(Handler.java:739)
 29530                 Bundle  W  at android.os.Handler.dispatchMessage(Handler.java:95)
 29530                 Bundle  W  at android.os.Looper.loop(Looper.java:135)
 29530                 Bundle  W  at android.app.ActivityThread.main(ActivityThread.java:5293)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Native Method)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Method.java:372)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
 29530                 Bundle  W  Key message expected String but value was a java.lang.Integer.  The default value <null> was returned.
 29530                 Bundle  W  Attempt to cast generated internal exception:
 29530                 Bundle  W  java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
 29530                 Bundle  W  at android.os.BaseBundle.getString(BaseBundle.java:921)
 29530                 Bundle  W  at org.sufficientlysecure.keychain.ui.dialog.PreferenceInstallDialogFragment.onCreateDialog(PreferenceInstallDialogFragment.java:96)
 29530                 Bundle  W  at android.app.DialogFragment.getLayoutInflater(DialogFragment.java:398)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:894)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1067)
 29530                 Bundle  W  at android.app.BackStackRecord.run(BackStackRecord.java:834)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1452)
 29530                 Bundle  W  at android.app.FragmentManagerImpl$1.run(FragmentManager.java:447)
 29530                 Bundle  W  at android.os.Handler.handleCallback(Handler.java:739)
 29530                 Bundle  W  at android.os.Handler.dispatchMessage(Handler.java:95)
 29530                 Bundle  W  at android.os.Looper.loop(Looper.java:135)
 29530                 Bundle  W  at android.app.ActivityThread.main(ActivityThread.java:5293)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Native Method)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Method.java:372)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
 29530                 Bundle  W  Key installPath expected String but value was a java.lang.Boolean.  The default value <null> was returned.
 29530                 Bundle  W  Attempt to cast generated internal exception:
 29530                 Bundle  W  java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String
 29530                 Bundle  W  at android.os.BaseBundle.getString(BaseBundle.java:921)
 29530                 Bundle  W  at org.sufficientlysecure.keychain.ui.dialog.PreferenceInstallDialogFragment.onCreateDialog(PreferenceInstallDialogFragment.java:97)
 29530                 Bundle  W  at android.app.DialogFragment.getLayoutInflater(DialogFragment.java:398)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:894)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1067)
 29530                 Bundle  W  at android.app.BackStackRecord.run(BackStackRecord.java:834)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1452)
 29530                 Bundle  W  at android.app.FragmentManagerImpl$1.run(FragmentManager.java:447)
 29530                 Bundle  W  at android.os.Handler.handleCallback(Handler.java:739)
 29530                 Bundle  W  at android.os.Handler.dispatchMessage(Handler.java:95)
 29530                 Bundle  W  at android.os.Looper.loop(Looper.java:135)
 29530                 Bundle  W  at android.app.ActivityThread.main(ActivityThread.java:5293)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Native Method)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Method.java:372)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
 29530                 Bundle  W  Key middleButton expected String but value was a java.lang.Integer.  The default value <null> was returned.
 29530                 Bundle  W  Attempt to cast generated internal exception:
 29530                 Bundle  W  java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
 29530                 Bundle  W  at android.os.BaseBundle.getString(BaseBundle.java:921)
 29530                 Bundle  W  at org.sufficientlysecure.keychain.ui.dialog.PreferenceInstallDialogFragment.onCreateDialog(PreferenceInstallDialogFragment.java:98)
 29530                 Bundle  W  at android.app.DialogFragment.getLayoutInflater(DialogFragment.java:398)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:894)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1067)
 29530                 Bundle  W  at android.app.BackStackRecord.run(BackStackRecord.java:834)
 29530                 Bundle  W  at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1452)
 29530                 Bundle  W  at android.app.FragmentManagerImpl$1.run(FragmentManager.java:447)
 29530                 Bundle  W  at android.os.Handler.handleCallback(Handler.java:739)
 29530                 Bundle  W  at android.os.Handler.dispatchMessage(Handler.java:95)
 29530                 Bundle  W  at android.os.Looper.loop(Looper.java:135)
 29530                 Bundle  W  at android.app.ActivityThread.main(ActivityThread.java:5293)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Native Method)
 29530                 Bundle  W  at java.lang.reflect.Method.invoke(Method.java:372)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
 29530                 Bundle  W  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
  742           WindowManager  V  not Base app: Adding window Window{e2029a6 u0 org.sufficientlysecure.keychain.debug/org.sufficientlysecure.keychain.ui.SettingsActivity} at 7 of 13
 29530         AndroidRuntime  D  Shutting down VM
 29530         AndroidRuntime  E  FATAL EXCEPTION: main
 29530         AndroidRuntime  E  Process: org.sufficientlysecure.keychain.debug, PID: 29530
 29530         AndroidRuntime  E  java.lang.NullPointerException: uriString
 29530         AndroidRuntime  E  at android.net.Uri$StringUri.<init>(Uri.java:470)
 29530         AndroidRuntime  E  at android.net.Uri$StringUri.<init>(Uri.java:460)
 29530         AndroidRuntime  E  at android.net.Uri.parse(Uri.java:432)
 29530         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.dialog.PreferenceInstallDialogFragment$2.onClick(PreferenceInstallDialogFragment.java:117)
 29530         AndroidRuntime  E  at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:162)
 29530         AndroidRuntime  E  at android.os.Handler.dispatchMessage(Handler.java:102)
 29530         AndroidRuntime  E  at android.os.Looper.loop(Looper.java:135)
 29530         AndroidRuntime  E  at android.app.ActivityThread.main(ActivityThread.java:5293)
 29530         AndroidRuntime  E  at java.lang.reflect.Method.invoke(Native Method)
 29530         AndroidRuntime  E  at java.lang.reflect.Method.invoke(Method.java:372)
 29530         AndroidRuntime  E  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
 29530         AndroidRuntime  E  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
  742         ActivityManager  W  Force finishing activity 1 org.sufficientlysecure.keychain.debug/org.sufficientlysecure.keychain.ui.SettingsActivity
```

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserver.java
```diff
@@ -190,36 +193,48 @@ private String getUrlPrefix() {
...
+            if (response.isSuccessful()) {
+                return responseBody;
             } else {
-                String data = readAll(conn.getErrorStream(), conn.getContentEncoding());
-                throw new HttpError(response, data);
+                throw new HttpError(response.code(), responseBody);
             }
         } catch (IOException e) {
-            throw new QueryFailedException("Keyserver '" + mHost + "' is unavailable. Check your Internet connection!");
+            e.printStackTrace();
```
dschuermann: Please use Log.e with the exception as the third parameter

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserver.java
```diff
@@ -334,13 +349,14 @@ private String query(String request) throws QueryFailedException, HttpError {
...
-    public String get(String keyIdHex) throws QueryFailedException {
+    public String get(String keyIdHex, Proxy proxy) throws QueryFailedException {
         String request = "/pks/lookup?op=get&options=mr&search=" + keyIdHex;
         Log.d(Constants.TAG, "hkp keyserver get: " + request);
         String data;
         try {
-            data = query(request);
+            data = query(request, proxy);
         } catch (HttpError httpError) {
+            httpError.printStackTrace();
```
dschuermann: Log.e, never print stacktrace

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeychainService.java
```diff
@@ -398,6 +397,7 @@ public void run() {
                         break;
                     }
                     case ACTION_IMPORT_KEYRING: {
+                        Proxy proxy = getProxyFromBundle(data);
```
dschuermann: move this a little bit down unter //Input

dschuermann:
The biggest and most important review comment comes last ;):
We need to rethink the program flow at many points in the code.
You often do something similar to this:
```
                final Preferences.ProxyPrefs proxyPrefs = Preferences.getPreferences(UploadKeyActivity.this)
                        .getProxyPrefs();
                Runnable ignoreTor = new Runnable() {
                    @Override
                    public void run() {
                        uploadKey(proxyPrefs.parcelableProxy);
                    }
                };

                if (OrbotHelper.isOrbotInRequiredState(R.string.orbot_ignore_tor, ignoreTor, proxyPrefs,
                        UploadKeyActivity.this)) {
                    uploadKey(proxyPrefs.parcelableProxy);
                }
```
This is maybe not preventable in some cases, but it puts much logic into the UI code which is scattered around.
I would like you to look into how we handle passphrase input and adopt that for cases were Orbot is not running and the user needs to decide what to do. The class CryptoOperationFragment gives much clue about how this works. When a crypto operations gives back a pendingResult (see onHandleResult() ) The corresponding activity that is required for user input is started in initiateInputActivity().
You could do the same for the "orbot is running" check. The operation itself could check the preference and if orbot is running and then return a specific pending result for this. Then an activity is executed that encapsulates the dialog (like done in PassphraseDialogActivity ) that asks the user what to do. The user decides and the process is method is repeated with the new parameters.
I hope this is clear enough. If you have trouble with this ask me or Valodim on IRC.

adithyaphilip:
Key import is now compatible with `KeychainNewService`.

I've introduced the `OperationHelper` class to deal with cases where the import operation is happening in an activity instead of a fragment. This introduced a bit of code duplication since it's pretty much a copy of `CryptoOperationFragment`, so I tried eliminating that with `NewCryptoOperationFragment` (which is used in `EditKeyFragment`). The overhead of using `OperationHelper` in fragments which extend `NewCryptoOperationFragment` is a `setOperationHelper` call in `onActivityCreated`.

The overhead in activities is maintaining a global `OperationHelper` variable and calling it's `handleActivityResult` in `onActivityResult`.

I think we might not be able to do away with the `isOrbotRequiredCheck` for actions that aren't operations, though, since they don't fit into this paradigm as of now.

I've currently implemented this in `ImportKeysActivity`, and if it's okay, I could go ahead and implement it in the remaining operations. It would also allow us to move all the operations to `KeychainNewService` since they now fit into the `execute(Parcel, Result)` format.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/ParcelableProxy.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+            case SOCKS: {
+                mProxyType = TYPE_SOCKS;
+                break;
+            }
+        }
+    }
+
+    public Proxy getProxy() {
+        if (mProxyHost == null) return null;
```
Valodim: mind the code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/ParcelableProxy.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * used to simply transport java.net.Proxy objects created using InetSockets between services/activities
+ */
+public class ParcelableProxy implements Parcelable {
+    private String mProxyHost;
+    private int mProxyPort;
+    private int mProxyType;
+
+    private final int TYPE_HTTP = 1;
```
Valodim: use an enum

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeychainService.java
```diff
@@ -403,23 +387,15 @@ public void run() {
...
-                                keyFingerprintSet.add(keyList.get(i).mExpectedFingerprint);
-                            }
-                            if (keyFingerprintSet.size() == keyList.size()) {
-                                // all keys have unique fingerprints
-                                multiThreadedKeyImport(keyList.iterator(), keyList.size(), keyServer);
-                            } else {
-                                serialKeyImport(keyList, keyServer, providerHelper);
-                            }
-                        }
+                        ImportExportOperation importExportOperation = new ImportExportOperation(KeychainService.this,
```
Valodim: sweet!

dschuermann:
please rebase on current master. I will then review again.

adithyaphilip:
Orbot enable/install dialog has been tested in the following places: 
* Key search 
* Key import
* Key "refresh" in `ViewKeyActivity`
* Key upload in `UploadActivity`
* Key upload during certification

Verified that calls are being routed through Orbot when it is running, using NoRoot Firewall.

This PR also includes `AddKeyserverException` in `uploadKey` being replaced with an appropriate `OperationLog`.

adithyaphilip:
Rebased on latest master.

dschuermann:
You have some unrelated code style changes in there.
Please at least make sure that OperationResult and ViewKeyTrustFragment don't end up having these code style changes and only your programmatic changes in there.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserver.java
```diff
@@ -190,36 +193,48 @@ private String getUrlPrefix() {
...
         }
-        if (conn == null) {
-            conn = (HttpURLConnection) url.openConnection();
-        }
-        conn.setConnectTimeout(5000);
-        conn.setReadTimeout(25000);
-        return conn;
+
+        client.setProxy(proxy);
+        // TODO: PHILIP if proxy !=null increase timeout?
```
dschuermann: Maybe solve this TODO before merge

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserver.java
```diff
@@ -398,6 +410,7 @@ public String toString() {
      * Tries to find a server responsible for a given domain
      *
      * @return A responsible Keyserver or null if not found.
+     * TODO: PHILIP Add proxy functionality
```
dschuermann: Make an issue for this that we don't forget to do it later

adithyaphilip: Created an issue here: https://github.com/open-keychain/open-keychain/issues/1403

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserver.java
adithyaphilip: Sorry, it was already solved, just forgot to remove the todo

adithyaphilip:
I've removed the code style changes from OperationResult and ViewKeyTrustFragment. I'll make sure not the mix them again.

dschuermann:
Can you squash the new commit "undid code style changes" into "ensuring code style is preserved" so that your back and forth with the code change is not preserved in git history.
There should be tutorials available online how to do git squash.

Mabye I should explain more ;) The reason I don't like the code style changes in the first place is that they make it more difficult to use "git blame". Now every line has been touched by you, before you could easily see how did what changes.

After that we can merge.

adithyaphilip:
Done. And ah, I see what you mean. I'll be more careful from now on. Thankyou for explaining that to me, I hadn't considered it.

dschuermann:
Wohoo :) :+1:

PR #1331, opened 2015-06-07
replace messenger hack with eventbus 

Valodim:
So, I think we should move from our messenger hack service communication data flow to [eventbus](https://github.com/greenrobot/EventBus). reasons:

- simplifies code: fragments are longer involved in progress or pendinginput, but can hook into it if they want
- better performance, because it saves a lot of parcelling and message passing all around. 
- hopefully helps getting rid of IllegalStateException problems, which are notoriously difficult with the messenger hack

in order to not turn this into a "someone ought to" deal, I went ahead and implemented it for all compliant operations (certify, sign/encrypt, decrypt/verify, edit key)


Valodim:
forgot to mention: this also allows us to get rid of ParcelableFileCache and workarounds like hydrate/dehydrate

dschuermann:
I am okay with this, but this needs to be tested on older Android versions. Looking at EventBus' issue reports, we could also run into all kind of new problems with this library.

Valodim:
(following discussion with dominik, also relevant for @adithyaphilip)

Here's the plan for this: I will backport all non-eventbus related changes. There are quite a lot of those, because more logic moved into the CryptoOperationFragment, the actual eventbus communication is abstracted from there. I will maintain the eventbus change in a separate branch, which we will put in a beta release to see how it goes, before merging it into stable.

Valodim:
changes backported, will merge if noone objects for a while

dschuermann:
merge! ;)

PR #1333, opened 2015-06-08
Explicit mention of dependency on 3rd party tool

VPhantom:
I had the darnest time remembering my old APG days where something was said in passing about, unlike any other application I know, the need here for OI File Manager (or others, later on) to be able to select local files.  I'm sure I'm not the only person who could benefit from this reminder. :)


dschuermann:
Great thanks

Issue #1303, opened 2015-05-27
descriptive nfc error messages


Valodim:
this is a follow-up of issue #1301

nfc error messages are at this point just "bad nfc response code NNNN". IOException is an appropriate class, since it represents a failed or interrupted IO operation, but it should be parameterized with the error response code.

the exception's getMessage() can return the same message as displayed right now for compatibility, but an additional getMessage(Context context) method could be introduced to return localized and better descriptive error messages. the mapping can probably not be done much nicer than a large switch statement, but maybe I'm missing something.

error messages on p51 of the [openpgp card spec](http://g10code.com/docs/openpgp-card-2.0.pdf)

dschuermann:
closed with https://github.com/open-keychain/open-keychain/pull/1338

Issue #1282, opened 2015-05-15
yubikey workflows


Valodim:
Thanks to this great pull request #1278  by @josecastillo, there are now mechanisms to a) create yubikey-compatible keys, which is suggested in the manage keys dialog when an empty yubikey is held against the device and b) support in the edit key dialog for moving those keys to the yubikey.

what's missing is a connection between those two. most users should almost never use the edit key dialog, and after a user created a yubikey-compatible key we can assume he wants to move it to his yubikey, so we should offer that option.

this is a little tricky because there are many intermediate states a user might end up in which must be thought about. in particular, what happens if the user creates such a key then stops the procedure?

Valodim:
labeling this discussion for now, we should take a step back and think where we are and where we want to go with yubikeys first. here's workflows we should cover for obtaining yubikey-bound key:

1. user has yubikey with a key on it, wants to import it to openkeychain
2. user has empty yubikey, wants to get it working with openkeychain
3. user has yubikey with a key on it, but wants to create a new one

I think we have 1) covered bar some bugs with special subkey constellations, although dominik had some doubts about the ui flow.

point 2) is almost there with the pull request, except for the missing steps in the wizard (see above).

for 3), my first idea would be to add a "clear key" operation, which does exactly that and leaves the yubikey in a state where flow 2) can be applied again.

there is also the matter of where to ask the admin pin for the yubikey. I'm not sure if the password-cache-flow which is used otherwise is a good idea here.

finally, there is currently no indication that a key is connected to a yubikey in the current ViewKeyActivity (#1252)

btw, does anyone know if yubikeys are shipped in this hybrid pgp+otp state already? I remember I had to change that on mine before I could use pgp features, which was a pretty user unfriendly operation.

josecastillo:
If the user aborts during the key-to-card operation (say, by pulling the Yubikey away after only one or two keys have been processed), their keyring shouldn't be stripped since an IOException will be thrown, preventing the NFC operaton from completing. The user might, however, end up with a Yubikey that has one or two of their keys, but not all three. 

Right now, they can just tap again and it will retry, but that's only because I haven't yet implemented a safety check. Ideally, we should throw an error if the user tries to move keys onto a Yubikey with existing keys, to prevent them clobbering an existing Yubikey with secret keys on it. My suggestion is the following enhancement (pseudocode) in the onNfcPerform case NFC_KEYTOCARD: 

    getData(oldFingerprint)
[...]

PR #1338, opened 2015-06-10
Improved smart card error handling

josecastillo:
This commit addresses #1303, adding descriptive error messages for the status codes returned by a Yubikey or smart card. A new ```CardException``` class extends ```IOException``` with a field to store the status word returned by the card. Some error conditions should be exposed to the user, like a missing key or a bad PIN; these are implemented with localizable strings that are shown in a notification. Other error conditions are only of interest to developers, like an invalid header byte or command chaining failure. These error conditions should not occur if the application is implemented correctly, and non-localized AssertionErrors are thrown in these cases.

This commit also takes a whack at something I mentioned in #1282: a safety check before the PUT KEY operation. If a user attempts to put a key into a slot that contains a key with a different fingerprint, an IOException is thrown so that the key already on the smart card will not be overwritten.


dschuermann:
Travis is failing due to my adventure into Android unit tests.
@josecastillo Can you rebase on top of current master? Travis is working again.

josecastillo:
No prob, done and done.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
```diff
@@ -105,6 +107,83 @@ public void handleNfcError(IOException e) {
...
+            case 0x6A80:
+                throw new AssertionError("Card returned 'Wrong Data' status; this is a programming error!");
+            case 0x6883:
+                throw new AssertionError("Card expected last command in a chain; this is a programming error!");
+            case 0x6B00:
+                throw new AssertionError("Card reported invalid P1/P2 parameter; this is a programming error!");
+            case 0x6D00:
+                throw new AssertionError("Instruction (INS) not supported by smart card; this is a programming error!");
+            case 0x6E00:
+                throw new AssertionError("Class (CLA) not supported by smart card; this is a programming error!");
```
Valodim: can all of these really never happen due to bad card status or anything? this is a hard crash, some of them sound like they could also happen from a misbehaving smartcard, or simply one that doesn't support openpgp?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
```diff
@@ -105,6 +107,83 @@ public void handleNfcError(IOException e) {
...
+
+            // These errors might be encountered in everyday use, and should display a localized
+            // error message to the user.
+            case 0x6285:
+            {
```
Valodim: code style: brace on case line

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
```diff
@@ -105,6 +107,83 @@ public void handleNfcError(IOException e) {
...
+                Notify.create(this, getString(R.string.error_nfc,
+                        getString(R.string.error_nfc_terminated)), Style.WARN).show();
```
Valodim: no need for getString here, can pass the resource ids directly

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseNfcActivity.java
josecastillo: This is a good point; I made this assumption from reading through the ykneo-openpgp source code, where they shouldn't happen unless we messed up; but a misbehaving card could force a crash by returning one of these statuses. Plus, someone developing a new NFC-compatible thing could mistakenly crash the app when they're testing it, and an alert would be more useful in that case. I'll turn them into snackbar notifications.

Valodim:
done! thanks for your work! :+1:

PR #1343, opened 2015-06-11
Update copyright info to 2015

thialfihar:
I took the liberty to update the copyright info to 2015. I left the MIT license (is that right? it's a new file) and any external copyrights untouched.

Commit contains the command used, perhaps useful next year.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/Keyserver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/Keyserver.java
@@ -1,7 +1,7 @@
 /*
- * Copyright (C) 2012-2014 Dominik Schürmann <dominik@dominikschuermann.de>
- * Copyright (C) 2011-2014 Thialfihar <thi@thialfihar.org>
- * Copyright (C) 2011 Senecaso
+ * Copyright (C) 2012-2015 Dominik Schürmann <dominik@dominikschuermann.de>
+ * Copyright (C) 2011-2015 Thialfihar <thi@thialfihar.org>
+ * Copyright (C) 2014-2015 Senecaso
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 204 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/Keyserver.java
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright (C) 2012-2014 Dominik Schürmann <dominik@dominikschuermann.de>
- * Copyright (C) 2011-2014 Thialfihar <thi@thialfihar.org>
- * Copyright (C) 2011 Senecaso
+ * Copyright (C) 2012-2015 Dominik Schürmann <dominik@dominikschuermann.de>
+ * Copyright (C) 2011-2015 Thialfihar <thi@thialfihar.org>
+ * Copyright (C) 2014-2015 Senecaso
```
dschuermann: Wrong

dschuermann:
:-1: 
I don't think this is important. There are also some wrong changes in it (see comment as example).
I am even not sure if we should do this. My understanding is that the year is only changed when a new bigger change is done to a file. I think this is also how AOSP handles this. Also see http://stackoverflow.com/questions/3487007/when-to-update-the-year-in-open-source-copyright-notice

Valodim:
ralted note, the MIT licensed file is a short snippet that is useful on its own, I put it on my gist and felt it didn't make much sense GPL'ing it

see: https://gist.github.com/Valodim/d741193322cfc24d722e

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/Keyserver.java
thialfihar: Good catch. Didn't update the single year replacement when I generalized it to years other than 2014.

thialfihar:
@Valodim: Ah, I didn't see that one, and since it was 2015 it wasn't touched. But the password strength indicator came from elsewhere and has an MIT license from 2014.
@dschuermann: Ok, I had a different understanding of it, but that makes sense.

PR #1349, opened 2015-06-17
Support for Marvin's minidns branch

adithyaphilip:
They've changed the library structure and done a little refactoring, so it required a few changes to our OpenKeychain project too. Just a quick PR in case I've done something wrong.


dschuermann:
@adithyaphilip Travis fails compiling so there need to be more fixes to get this working.

Valodim:
ah, indeed. looks like the library isn't actually updated in the commit itself

adithyaphilip:
I'm not sure what's wrong, it fails if I try a `gradlew build`, right after cloning the project or doing a `clean`, but it works from inside studio. Also, once it builds using studio, further builds from the terminal work unless I do a `clean`. I guess this means studio is somehow compiling `extern:minidns:minidns-core` but I haven't configured it correctly for the wrapper to build from terminal?

dschuermann:
Ah I see. This pr merges into  tor-support from proxy-minidns. So it's on hold until we resolve the tor-support problems

dschuermann:
@adithyaphilip Can you rebase this on top of master so that we only have the two related commits in here?

adithyaphilip:
Yup, I thought I'd actually try doing this over from the start. I don't think I added the library right. I'll just close it for now?

dschuermann:
jup np

PR #1352, opened 2015-06-17
instrumentation branch

Valodim:
I cleaned this one up and removed all v/multi-decrypt commits. this should be ready to merge, but maybe someone wants to review it. no features added, but includes a bunch of patches for problems encountered during testing, and a ton of test coverage.


Valodim:
possible regressions:
- keyspinner now uses the keyadapter, and consequently the key_list_item layout. this may lead to incorrect display of key status
- removed DialogFragmentWorkaround for PassphraseDialogFragment, and moved fragment logic into onResume() + onPause() instead. I think this is the better way to deal with this, but it may introduce platform-specific problems

Valodim:
oh also, regarding instrumentation coverage: running instrumentation tests on travis is not feasible at the moment, so I ran them and committed the resulting .ec coverage file into the repository. that's not great, but seems a viable solution to the problem for the moment.

dschuermann:
looks good to me, nothing obvious wrong here.

Issue #1356, opened 2015-06-19
Fix crash when using icon theme with mask for unknown icons


twouters:
I've experienced crashes in openkeychain when using the icon theme provided by [TwelF](https://gitlab.com/xphnx/twelf_cm12_theme).

```
[snip]
W/StatusBarIconView(2244): Icon not found in 2130837623: 7f020077
W/StatusBarIconView(2244): No icon for slot org.sufficientlysecure.keychain/0x1
W/StatusBar(2244): removeNotification for unknown key: 0|org.sufficientlysecure.keychain|1|null|10101
D/NotificationService(954): onNotification error pkg=org.sufficientlysecure.keychain tag=null id=1; will crashApplication(uid=10101, pid=27442)
D/AndroidRuntime(27442): Shutting down VM
E/AndroidRuntime(27442): FATAL EXCEPTION: main
E/AndroidRuntime(27442): Process: org.sufficientlysecure.keychain:passphrase_cache, PID: 27442
E/AndroidRuntime(27442): android.app.RemoteServiceException: Bad notification posted from package org.sufficientlysecure.keychain: Couldn't create icon: StatusBarIcon(pkg=org.sufficientlysecure.keychainuser=0 id=0x7f020077 level=0 visible=true nu
m=0 )
E/AndroidRuntime(27442):     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1484)
E/AndroidRuntime(27442):     at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime(27442):     at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime(27442):     at android.app.ActivityThread.main(ActivityThread.java:5289)
E/AndroidRuntime(27442):     at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(27442):     at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime(27442):     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
E/AndroidRuntime(27442):     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
E/AndroidRuntime(27442):     at de.robv.android.xposed.XposedBridge.main(XposedBridge.java:117)
```

This seems to be a rather common issue on mostly - if not only - cyanogenmod devices:
- [CYAN-6506](https://jira.cyanogenmod.org/browse/CYAN-6506)
[...]

PR #1360, opened 2015-06-23
New all-white status bar icon to fix issue #1356

moparisthebest:
This implements the fix proposed and used by mirakel, to solve issue #1356, I did take a function from mirakel, but both projects are GPLv3 so there are no problems at all.

I'm not an artist, so I took a creative commons icon from the noun project and gave the required credit in the readme, I of course couldn't care less what icon is actually used, as long as it doesn't cause a crash like the current icon does.  I converted the SVG using the inkscape commands from update-drawables.sh

Let me know what else if anything is required to get this merged.


dschuermann:
Thanks for your work. I added a new notify icon based on our launcher icon here: https://github.com/open-keychain/open-keychain/commit/8cd7e295404fdfad5167b96fdf3a5cbeca8e4cd1

PR #1367, opened 2015-06-25
Fixed fingerprint mismatch on web download of keys

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1365. Key query had fingerprint in caps but we use lower case internally.


dschuermann:
Great! Thanks for fixing.

PR #1373, opened 2015-06-26
Add debug app name

thialfihar:
That way debug builds can be distinguished from the release version by name as well.

Also a minor refactoring, explicitly importing "R" an "BuildConfig" by FQN where that wasn't the case.


dschuermann:
thanks

PR #1374, opened 2015-06-26
Add theme setting with light and dark theme

thialfihar:
This tries to refactor the theming a bit, moving away from fixed colours (except where it makes sense, such as key flags) and using attributes instead that the theme can override.

Relevant changes:
* a theme setting (light and dark, light is default)
* ~~while working on the settings I noticed that the Advanced preferences section was misnamed (it's not only "password caching"), so renamed it to "Advanced", reflecting the internal naming of the respective fragment~~
* BaseActivity now will re-create the activity if it notices a theme change `onResume()`
* the light theme is almost identical to OpenKeychain theme, but for these adjustments:
  * the pure white toolbar in the Encrypt/Decrypt view is slightly darker, because the full white drowned out the system's notification bar, now it is still close to white but the bar is visible on top of it
  * removed the superfluous divider in the decrypt files view
  * reduced the large double-separator in the import keys view to be a normal thin divider, consistent with the rest of the GUI, also adjusted the divider at the bottom to match it
  * the background in the import keys view now also honours the light theme, previously it used white directly


OpenKeychain/src/main/res/xml/preference_headers.xml
```diff
@@ -1,8 +1,11 @@
 <preference-headers xmlns:android="http://schemas.android.com/apk/res/android">
     <header
+        android:fragment="org.sufficientlysecure.keychain.ui.SettingsActivity$GuiPrefsFragment"
+        android:title="@string/section_gui" />
+    <header
         android:fragment="org.sufficientlysecure.keychain.ui.SettingsActivity$CloudSearchPrefsFragment"
         android:title="@string/section_cloud_search" />
     <header
         android:fragment="org.sufficientlysecure.keychain.ui.SettingsActivity$AdvancedPrefsFragment"
-        android:title="@string/section_passphrase_cache" />
```
Valodim: section_passphrase_cache is actually more accurate, all unrelated settings were removed from that dialog at some point

thialfihar: Ah, I only now understand that the YubiKey PIN usage setting is a kind of password cache as well, but even given that, the "Use number keypad for YubiKey PIN" still seems out of place, no?

Do you think it would make sense to rename this back to "Password Cache" section and pull out the YubiKey ones into their own section? It's a very specific feature after all.

thialfihar:
Rebased onto recent multi-decrypt changes, also adjusting the new GUI stuff.

thialfihar:
Rebased onto recent change of the password cache section.

dschuermann:
* After upgrading no theme is selected by default in settings, but Light should be selected

Problems with Light theme:
* selected items in drawer menu are blue
* Background of Toolbar in ViewKey is blue

Problems with Dark theme:
* Some parts of the design are not perfect, but maybe we can accept this for now
* Many icons should be changed to the white version, but again we could accept this for now and do this in a later pull request.

Also please rebase again :)

dschuermann:
Regarding preferences update: There is Preferences.updatePreferences() for this

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -202,7 +202,15 @@ public boolean getTextUseCompression() {
         return mSharedPreferences.getBoolean(Pref.TEXT_USE_COMPRESSION, true);
     }
 
+    public String getTheme() {
+        return mSharedPreferences.getString(Pref.THEME, "light");
```
dschuermann: Please put "dark" and "light" into public static final...

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SettingsActivity.java
```diff
@@ -95,7 +101,40 @@ public boolean onPreferenceClick(Preference preference) {
...
                     (CheckBoxPreference) findPreference(Constants.Pref.USE_NUMKEYPAD_FOR_YUBIKEY_PIN));
 
+        } else if (action != null && action.equals(ACTION_PREFS_GUI)) {
+            addPreferencesFromResource(R.xml.gui_preferences);
+
+            initializeTheme((ListPreference) findPreference(Constants.Pref.THEME));
+        }
+    }
+
+    @Override
```
dschuermann: The following code is repeated in BaseActivity. Could you externalize this into it's own class?

thialfihar:
Aye, the colours should work now. I upgraded MaterialDrawer in the process. The light theme doesn't derive from a MaterialDrawerTheme, because the drawer icon then becomes black, unfortunately. But this now should look like before.

The constants and upgrading of preferences is in there as well. I renamed `updatePreferences()` to `upgradePreferences()`, because I think that's the correct name. Let me know if you feel strongly about the former name.

I'm working on the refactoring of the theme change now. I tried to unify that originally, but it got ugly. I think I have a better idea now.

thialfihar:
I think that works well.

dschuermann:
I am fine with this so far.
@Valodim could you also look into this?

Valodim:
looked the code over, looks good so far! I'll compile it and check out how it looks in a while

Valodim:
crashes in main activity for me, probably some incompatibilities with recent changes in the coloring code there?

```
android.content.res.Resources$NotFoundException: Resource ID #0xff000000
at android.content.res.Resources.getValue(Resources.java:1310)
at android.content.res.Resources.getValue(Resources.java:1290)
at android.content.res.Resources.getColor(Resources.java:948)
at org.sufficientlysecure.keychain.ui.adapter.KeyAdapter$KeyItemViewHolder.setData(KeyAdapter.java:181)
at org.sufficientlysecure.keychain.ui.adapter.KeyAdapter.bindView(KeyAdapter.java:224)
at android.support.v4.widget.CursorAdapter.getView(CursorAdapter.java:256)
at org.sufficientlysecure.keychain.ui.KeyListFragment$KeyListAdapter.getView(KeyListFragment.java:713)
at se.emilsjolander.stickylistheaders.AdapterWrapper.getView(AdapterWrapper.java:164)
at se.emilsjolander.stickylistheaders.AdapterWrapper.getView(AdapterWrapper.java:26)
at android.widget.HeaderViewListAdapter.getView(HeaderViewListAdapter.java:220)
at android.widget.AbsListView.obtainView(AbsListView.java:2346)
at android.widget.ListView.makeAndAddView(ListView.java:1864)
at android.widget.ListView.fillDown(ListView.java:698)
at android.widget.ListView.fillFromTop(ListView.java:759)
at android.widget.ListView.layoutChildren(ListView.java:1673)
at se.emilsjolander.stickylistheaders.WrapperViewList.layoutChildren(WrapperViewList.java:193)
at android.widget.AbsListView.onLayout(AbsListView.java:2150)
at android.view.View.layout(View.java:15595)
at android.view.ViewGroup.layout(ViewGroup.java:4966)
at se.emilsjolander.stickylistheaders.StickyListHeadersListView.onLayout(StickyListHeadersListView.java:257)
at android.view.View.layout(View.java:15595)
at android.view.ViewGroup.layout(ViewGroup.java:4966)
at android.widget.FrameLayout.layoutChildren(FrameLayout.java:573)
(...)
```

Valodim:
workaround-fixed it and checked out the theme. Maybe I'm just not used to dark themes, but I really can't say I'm convinced of the overall look and feel. besides the general feeling, here are a couple of problems I noticed:
- dialogs (progress, passphrase, ...) aren't dark and flash up brightly
- snackbars are dark in the regular theme as well, so their background is now the same as the regular one making them difficult to distinguish
- toolbar is now blue for confirmed keys and show a blue checkmark in the background, which dilutes the "green checkmark" concept
- generic toolbar is blue as well, but drawer header isn't

thialfihar:
Cheers for the review.

* Yeah, that's ugly. That theme is set specifically in the code. I'll try to make it honour the setting.
* Will look into that.
* Aye, that probably should be green as well, since the other header is orange also.
* Not sure what you mean there. But I just noticed the toolbar for extended info is still green, do you mean that one?

dschuermann:
Regarding the dialogs: We should probably use AppCompatDialog to get the right design.

dschuermann:
Partly done here: https://github.com/open-keychain/open-keychain/commit/56d599dd7648c4eafd87f8802f10cc86bcd67a4a
You can specify a theme in CustomAlertDialogBuilder's constructor that is designed differently for light and dark.

OpenKeychain/src/main/res/values/colors.xml
```diff
@@ -1,67 +1,28 @@
...
+    <color name="password_strength_medium">#ff9800</color>
+    <color name="password_strength_high">#7bad45</color>
 
     <!--
+        Standard Android colors:
         http://www.google.com/design/spec/style/color.html#color-color-palette
-        light = normal color
-        dark = 900
-
-        exception: green
```
dschuermann: why have you removed these comments? they are still valid.

or: At the time I had removed the green. Failed to restore them afterwards... will do now, tho.

or:
Apart from rebasing (and the re-adding of the comments above) the last two commits deal with the issues mentioned above:

* Dialogs should be pretty now (@dschuermann: let me know if you disagree with the move of that code to `ThemeChanger`)
* Fixed the toolbar and status bar for key views... in fact, the previous code coloured them both the same, but now it should follow style guides with a darker version for the status bar.

@Valodim I'm still unsure about your fourth point, but I think that might have been fixed with the key view status bar? Otherwise please let me know how I can reproduce it.

I personally think the dark snackbar works well with the dark theme. I propose to leave it at that for now and maybe adjust it later, if prolonged usage uncovers some issues.

dschuermann:
I merged the pull request because our old design still works as far as I tested and this is a great start for new designs :+1:

Although, I am still not fully convinced of the proposed dark design, but we can sort that out in later pull requests/commits. I will also discuss this again with the other contributors.

Issue #1357, opened 2015-06-22
Visual element to verify if the rewritten password matches.


ishit:
What are your thoughts on this?

dschuermann:
Please provide a clear description about what you have in mind. I will then reopen this issue.

ishit:
During the signup phase, when the user is asked to choose a password and retype that password, we can provide a visual indication(like a green tick) that the retyped password matches the original password.

dschuermann:
Great idea. I tagged this with help-wanted, if anyone wants to work on this.

ishit:
Shall I send a pull request for this issue? Seems like workable for first contribution.

dschuermann:
yes

ishit:
@dschuermann Can we close this?

Issue #1382, opened 2015-06-30
YubiKey creation: PIN/Admin PIN


dschuermann:
Create a screen for the create key wizard that shows a random PIN + Admin PIN (first numbers of Admin PIN = PIN).
Next screen: User needs to type in these PINs to prove that she has remembered them.

Add PIN/Admin PIN back end functionality for SaveKeyringParcel to apply them after moving the keys to the card.

Assigned to myself.

Valodim:
no reason to move the pin/admin pin through the backend via SaveKeyringParcel, the dataflow is a lot simpler if this is just done in the activity.

dschuermann:
Not sure if I can get a nice program flow when I want to execute this together with moving the keys to the card. That's the reason I thought to do it this way.

dschuermann:
done in https://github.com/open-keychain/open-keychain/commit/195508ed92434197d0d6ab2d3ef6e0b4bd0780b6

PR #1387, opened 2015-06-30
Add visual element to verify the re-typed password. Refer #1357.

ishit:


dschuermann:
Great idea.
Maybe we should use the same visuals as the ones used for the email text field for a unified experience ;)
I would also like to see this in it's own class similar to EmailEditText so I can also reuse this for https://github.com/open-keychain/open-keychain/issues/1382
Do you have time to do this?

ishit:
Yes, how do you want the data to flow between PassphraseEdit and PassphraseEdiitAgain? Something like a BroadastManager? Also, shall I add another icons for this(copy of email text field icons) or reuse and rename the existing email text field icons?

dschuermann:
* I see, the program flow would be too difficult. Then let's leave it integrated, don't put it in a new class.
Reuse
* Reuse and rename existing email text field icons

ishit:
Done. 
rsz_device-2015-07-02-201036

dschuermann:
Great work, much appreciated!

Issue #1288, opened 2015-05-21
GSoC 2015: Reworked Preferences UI


adithyaphilip:
The current keyserver preferences UI needs a rework. We would have to rewrite the list using RecyclerView.

This is what it looks like right now:

![screenshot_2015-05-21-19-51-08](https://cloud.githubusercontent.com/assets/5712241/7750608/d6eb1676-fff2-11e4-896a-6703c3a8abd7.png)

We need to highlight that the keyserver at the top of the list is the preferred one. Also we'd have to replace the "Rotate" button with something like a drag-and-drop interface (like https://github.com/h6ah4i/android-advancedrecyclerview). 
Alternatively, a radio button might help with both highlighting and choosing the preferred keyserver.

EDIT:
Design guidelines: http://www.google.com/design/spec/components/lists-controls.html#lists-controls-types-of-list-controls

Yoshi64Bit:
It is also possible to use [this](https://github.com/bauerca/drag-sort-listview) library. It's a ListView which supports drag and drop and item removing.

dschuermann:
Thanks for the other library, unfortunately it is unmaintained and not based on recyclerview. I would only considered "future proofed" libs ;)

dschuermann:
Google just announced new version of support lib should support RecyclerView with drag and drop. At least they say it just this minute in Google IO live stream.

PR #1395, opened 2015-07-01
Reworked Keyserver Preferences

adithyaphilip:
Corresponds to https://github.com/open-keychain/open-keychain/issues/1288.
This is what it looks like now:
screenshot_2015-07-02-01-16-44
screenshot_2015-07-02-00-25-43
Notable difference from earlier implementation - there is no save button. It saves to preferences every time a change is made.


dschuermann:
Great work! Immediately merged!

Issue #1161, opened 2015-03-19
Clean up keyserver request threads in keyimport/CloudSearch.java


jellysheep:
When searching for keys in the cloud, in `keyimport/CloudSearch.java` threads are created (usually 2, [see here](https://github.com/open-keychain/open-keychain/blob/development/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/CloudSearch.java#L60)) to send requests to HKP and Keybase keyservers. After a maximum duration of 10 seconds, the request results are collected.
The created threads are not cleaned up, instead they are expected to terminate themselves. However if the request to a keyserver needs a long time (>10s, e.g. due to slow network connection), the threads remain, even if a new search is started, although the threads should be cleaned up by `Thread.interrupt()`.

dschuermann:
Is this still valid @adithyaphilip ? You are currently familiar with this code ;)

PR #1399, opened 2015-07-01
Kill search threads that have not returned

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1161.


dschuermann:
Thanks

Issue #1398, opened 2015-07-01
Keyserver preference: Crashes on empty list


dschuermann:
1. Delete all keyservers
2. Go back and open keyserver preference again -> crash

Assigned to @adithyaphilip

dschuermann:
Maybe we simply shouldn't allow that the user can delete all keyservers.

PR #1400, opened 2015-07-02
Fixe Keyserver Preference crash on empty list

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1398.
I've made it accept already existing empty keyserver preferences (in case someone updating has 0 keyservers), but it won't let you delete if you have only 1 keyserver. Having 0 keyservers was unchecked in all places where `getPreferredKeyserver` was being used. Since we were shifting to "at least 1 keyserver", I've left them that way, is that okay?


dschuermann:
Yeah I think it's okay this way.

Issue #1406, opened 2015-07-06
Proxy settings improvement


dschuermann:
Currently users can change Proxy Host, port and type even when "Enable other proxy" is disabled.
These 3 settings should depend on "Enable other proxy".

adithyaphilip:
Closed via https://github.com/open-keychain/open-keychain/commit/6baf8163deb045f8a31fea7ad853b156502f018e

PR #1414, opened 2015-07-07
Linked other proxy settings to "Enable other proxy"

adithyaphilip:
Solves https://github.com/open-keychain/open-keychain/issues/1406


extern/KeybaseLib
```diff
@@ -1 +1 @@
-Subproject commit 0b0a60533c5f76b60e43895f3a0342bb0be68539
```
dschuermann: Please revert this change and squash. I think you forgot git submodule update

Issue #1412, opened 2015-07-07
NullPointerException: ParcelableProxy.getProxy()


dschuermann:
Attach YubiKey with existing key to import this key from a keyserver -> crash

```
  8072         AndroidRuntime  E  FATAL EXCEPTION: ModernAsyncTask #1
  8072         AndroidRuntime  E  Process: org.sufficientlysecure.keychain.debug, PID: 8072
  8072         AndroidRuntime  E  java.lang.RuntimeException: An error occured while executing doInBackground()
  8072         AndroidRuntime  E  at android.support.v4.content.ModernAsyncTask$3.done(ModernAsyncTask.java:137)
  8072         AndroidRuntime  E  at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
  8072         AndroidRuntime  E  at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
  8072         AndroidRuntime  E  at java.util.concurrent.FutureTask.run(FutureTask.java:242)
  8072         AndroidRuntime  E  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
  8072         AndroidRuntime  E  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
  8072         AndroidRuntime  E  at java.lang.Thread.run(Thread.java:818)
  8072         AndroidRuntime  E  Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.net.Proxy org.sufficientlysecure.keychain.util.ParcelableProxy.getProxy()' on a null object reference
  8072         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListCloudLoader.queryServer(ImportKeysListCloudLoader.java:103)
  8072         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListCloudLoader.loadInBackground(ImportKeysListCloudLoader.java:67)
  8072         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListCloudLoader.loadInBackground(ImportKeysListCloudLoader.java:35)
  8072         AndroidRuntime  E  at android.support.v4.content.AsyncTaskLoader.onLoadInBackground(AsyncTaskLoader.java:242)
[...]

PR #1424, opened 2015-07-08
Shifted proxy logic to ImportKeysListCloudLoader

adithyaphilip:
Should fix https://github.com/open-keychain/open-keychain/issues/1412. Tried making it similar to how CryptoOperationFragment handles it. Should we have something like that for loaders?


dschuermann:
What do you mean with "have something like this for loaders"?
Is there a use case missing in this pull request that still needs to be handled?

adithyaphilip:
Nope, haven't left out any use cases I know of. Just meant something like CryptoLoaderHelper with callbacks to hide the OrbotHelper check like we do in CryptOperationHelper. Not too necessary, since we only seem to use it in `ImportKeysListCloudLoader`.

Issue #1425, opened 2015-07-08
Scanning with Barcode Scanner hangs at import


dschuermann:
1. Scan with Barcode Scanner
2. Gets URI with fingerprint and Imports in OpenKeychain
3. Import hangs

Needs more investigation.

adithyaphilip:
Looking into it.

Valodim:
we'd like to push out a bugfix release today, should we wait for this one?

adithyaphilip:
I'm almost done, just testing. Half an hour?

dschuermann:
sure np

Valodim:
no rush, I'll need more time than half an hour myself for a bunch of nitpickings before the release ;)

adithyaphilip:
Does this issue still persist?

PR #1429, opened 2015-07-08
Fixed QR code import and serialKeyImport progress

adithyaphilip:
I wasn't able to exactly reproduce https://github.com/open-keychain/open-keychain/issues/1425 - my issue was related to when Tor was selected for proxy but wasn't running. A related issue is that selecting "Don't Use Tor" will result in the key being imported and the user taken to the certify screen, but no progress dialog being displayed. Looking into that too.


adithyaphilip:
Shows progress dialog after choosing to not use Orbot now.

Issue #1252, opened 2015-05-06
Always show YubiKey card view


dschuermann:
The YubiKey card view should be always visible for keys bound to YubiKeys

Valodim:
This is not so simple actually. The Yubikey serial is stored in the s2k dummy data of the secret key, but it is not available from the content provider at this point, and neither do we store other information from the yubikey such as owner name. This reduces available information for a yubikey card to pretty much only its presence, not what is usually shown.

Valodim:
with the key status card, this is no longer necessary

Issue #1449, opened 2015-07-16
show more information after secret key import


Valodim:
missing secret key capabilities seem to be a constant source of confusion for users. it might be a good idea to show a dialog to the user after import of a secret key if there are any special circumstances about it (missing/divert/stripped subkeys for certify/sign/decrypt)

Valodim:
Here's a mockup fragment:

![](http://valodim.stratum0.net/layout-2015-07-16-215801.png)

Valodim:
see #1450

Issue #1436, opened 2015-07-10
Confirm key + upload broken?


dschuermann:

Valodim:
only the upload is broken, the key is correctly confirmed before that. thinking about this, the upload should probably not be a part of the operation itself, but rather run in the background afterwards, there is no reason the ui flow needs to block until the upload is finished.

mohlerm:
I just tried confirming a friends key. It doesn't matter if we use fingerprint comparison / QR scan or Safeslinger: confirming key doesn't work, the button does nothing.
We didn't tick 'upload' cause we don't want that.

dschuermann:
@mohlerm If you don't check "Synchronize with the cloud" it should work. At least I can't reproduce your problem with 3.4.1.
What version of Openkeychain and Android are you using

dschuermann:
@Valodim I also just tried to reproduce confirm together with upload and it works for me. At least OpenKeychain don't hang or crashes. What exactly have you encountered?

mohlerm:
@dschuermann Sorry for not being specific in the first post.
I'm using 3.4.1 from fDroid on a 5.0 Xperia Z3C. I'm using a masterkey+3 subkey setup and the subkeys are stored on a Yubikey NEO.

Steps to reproduce: 
1. click on a person with a orange question mark symbol
2. Confirm via fingerprint comparison (can't do the others due to lack of 2nd phone atm)
3. Press "Fingerprints match"
4. Select my Key
5. Untick 'Synchronize with the cloud'
6. Confirm key -> Nothing happens
Sometimes a (progress bar?) window flashes for a milisecond, can't read anything on it though.

Valodim:
masterkey+3subkey? can you elaborate on that? or tell us the key id, if it's on keyservers?

mohlerm:
Sure, KeyID: 0x4d90384e53a54380
I've created a signing master key and 3 subkeys: one authentication, one signing and one encryption (sub)key. These three keys are transfered to the Yubikey while the master key is kept airgapped.

dschuermann:
@mohlerm Sry, but lol.
Have you followed a tutorial? If yes can you please send us the URL?
The OpenPGP standard defines that only the master key can certify other keys, thus you cannot certify with OpenKeychain if your master key is airgapped.
However this should show a notification with an error that this is not possible.

mohlerm:
Oh, I actually thought about this being the issue.
[...]

PR #1450, opened 2015-07-17
secret key status activity

Valodim:
I started working on what I mentioned in #1449. The idea is that a user can see what operations are available for a (secret) key, but not in terms of subkeys. This translates into a simple activity which displays a status each for confirm, sign and decrypt.

I covered all obvious cases, but currently only the subkey status itself is considered and not others, like when the entire key is revoked. I reckon there is also some room for improvement in the way of understandable wording and design.

The activity is currently accessible via the context menu, but I also thought about displaying it right after the user imports a secret key, which will hopefully help for scenarios like #1436 where the user isn't entirely aware of what operations a secret key can or cannot do.

screenshot from the fragment, it looks a little better in action but I can't make a very good screenshot at the moment:

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
@@ -121,6 +121,29 @@ public class CanonicalizedSecretKey extends CanonicalizedPublicKey {
             return this != UNAVAILABLE && this != GNU_DUMMY;
         }
 
+        /** Compares by "usability", which basically compares how independently usable
+         * two SecretKeyTypes are. The order is roughly this:
+         *
+         * empty passphrase < passphrase/others < divert < stripped
+         *
+         */
+        public int compareUsability(SecretKeyType other) {
+            // if one is usable but the other isn't, the usable one comes first
+            if (isUsable() ^ other.isUsable()) {
+                return isUsable() ? -1 : 1;
+            }
+            // if one is a divert-to-card but the other isn't, the non-divert one comes first
+            if ((this == DIVERT_TO_CARD) ^ (other == DIVERT_TO_CARD)) {
+                return this != DIVERT_TO_CARD ? -1 : 1;
+            }
+            // if one requires a passphrase but another doesn't, the one without a passphrase comes first
+            if ((this == PASSPHRASE_EMPTY) ^ (other == PASSPHRASE_EMPTY)) {
+                return this == PASSPHRASE_EMPTY ? -1 : 1;
+            }
+            // all other (current) cases are equal
+            return 0;
+        }
+
     }
 
     /** This method returns the SecretKeyType for this secret key, testing for an empty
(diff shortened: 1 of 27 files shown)


dschuermann:
I like the idea and design, but I am not sold on the idea of having another activity.
What if we show the red items (and maybe yubikey ones) in a card view inside the key view activity?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -117,6 +117,26 @@ public boolean isUsable() {
             return this != UNAVAILABLE && this != GNU_DUMMY;
         }
 
+        /** Compares by "usability", which basically compares how independently usable
+         * two SecretKeyTypes are.
+         */
+        public int compareUsability(SecretKeyType other) {
```
dschuermann: Why not call it capabilities?

dschuermann:
Or maybe even displaying all items in a cardview, not only the read ones. I think it's okay to do this as it only clutters the key view of your own keys.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
Valodim: ok

dschuermann:
If we simply display this in a card view it would also fix https://github.com/open-keychain/open-keychain/issues/1252

Valodim:
I'm not sure. The 95% case here should be that the key can do all three, any other scenario is already special in some way. So for most users this would be useless information at best, confusing at worst because it suggests that any of these points can not be taken for granted and/or might change over time.

dschuermann:
Point taken. So then I would go back to my first proposal: Show only the red items in a card view.

Valodim:
I changed the fragment into a custom view, which is now displayed as a card in secret keys.

dschuermann:
I like really like the new design!
* As you said in your initial comment, it should consider the revocation/expiry status of the master key.
* The menu item should be removed

Long term thoughts:
We can introduce buttons to guide the user to the right action, like extending the expiry

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
Valodim: on second thought, I chose the term usability because this is explicitly *not* capability. it compares two keys which have the same capabilities (sign/encrypt/certify) and orders them by how indepently usable they are, i.e.: "empty passphrase < passphrase/others < divert < stripped"

maybe the word "usability" isn't perfect because it is used in a different sense than UX, but "capability" is defined more sharply in the context of keys, so I want to avoid that word.

Valodim:
Soo now that we are going to do the master password thing, and we'll be needing dialogue on import ("to import, please insert passphrase"), I'm back to thinking this status thing should be shown right there as part of the import process, in a sort of "key successfully imported, here's what you can do with it" manner.

Thoughts?

dschuermann:
ack, sounds reasonable.

Issue #1304, opened 2015-05-28
Change keybase strings


dschuermann:
Instead of post-pending the username the strings should use a variable. This is true for all "keybase_narrative_*"

xSooDx:
I've been trying to work on this for a while, but i can't figure out exactly what to do.
Initially i thought to add '%s' to the keybase_narrative_* strings and use the getString() method to add the usernames but this wouldn't work because it is possible to have multiple usernames.

The existing code seems to work very well as it is. Any change would make little to no improvements.

Could you elaborate so i can better understand what is to be done?

dschuermann:
@xSooDx your approach is right. Use %s instead of prepending the usernames. This should be changed because in some languages the usernames are not at the end of the sentence but in the middle.

PR #1456, opened 2015-07-21
ISSUE #1304: Change keybase strings 

xSooDx:
-No longer appends the links to the keybase_narratives
-Now uses '%s' to insert the names in the keybase_narratives
-added '%s' to keybase_narrative_* in the string.xml file
-- NOTE:Have not modified for other languages

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyTrustFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyTrustFragment.java
@@ -269,8 +269,12 @@ public class ViewKeyTrustFragment extends LoaderFragment implements
         private SpannableStringBuilder insertLinks(SpannableStringBuilder proofLinks,String proofType){
             SpannableStringBuilder ssb = new SpannableStringBuilder();
             ssb.append(proofType);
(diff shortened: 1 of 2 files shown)


Valodim:
the build failure is not your fault. no idea why that happens, rerun should fix it

dschuermann:
@xSooDx Thanks for your pull request.
Is there a reason for using your own method "insertLinks()" instead of Android's formatting methods, described at http://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling ?

xSooDx:
@dschuermann I did it because the i couldn't directly format a spannablestringbuilder without loosing the links in the string. string.format removed the links

dschuermann:
@xSooDx Please rename the method to something like "formatSpannableString" and add a comment above the method why String.format cannot be used.

timbray:
Why is the %s approach better?

dschuermann:
@timbray In some languages the usernames are in the middle of a sentence and not at the end. The %s allows to have a dynamic position. See https://github.com/open-keychain/open-keychain/issues/1304

xSooDx:
@dschuermann Made the changes you requested.

dschuermann:
Great work

PR #1461, opened 2015-07-27
Periodic Keyserver Sync

adithyaphilip:
Still a work in progress. Currently:
* A new table UPDATED_KEYS keeps track of the last updated times of keys. This values is set every time a key is imported from cloud, but not for import from file. The value is preserved during consolidate and the key deletion that takes place when you re-save an existing public key, but is lost on just key deletion.
* Every time the SyncAdapter is run, it will update all keys that are older than a day.
* All updates are cancelled when the user is on KeyListFragment.

Todo:
* Send a new username/pass to Orbot using SOCKS proxy for a new circuit each time.
* Finish implementing notification that appears if user has checked Tor in proxy settings but it is not enabled at time of sync.
* Implement the periodic interval at which the SyncAdapter runs.
* Preference to turn it off, and checkbox to enable it in FirstTimeActivity


OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -800,6 +800,20 @@
                 android:resource="@xml/custom_pgp_contacts_structure" />
         </service>
 
+        <service
+            android:name=".service.KeyserverSyncAdapterService"
+            android:exported="true"
```
Valodim: exported?

OpenKeychain/src/main/res/xml/keyserver_sync_adapter_desc.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<sync-adapter xmlns:android="http://schemas.android.com/apk/res/android"
+    android:contentAuthority="org.sufficientlysecure.keychain.debug.provider"
```
dschuermann: Make sure that this also works for non-debug builds. Maybe you need to introduce a string in build.gradle

OpenKeychain/src/main/AndroidManifest.xml
adithyaphilip: The guide at https://developer.android.com/training/sync-adapters/creating-sync-adapter.html said it was necessary so the system (the Android SyncAdapter framework, I assume) can access it?

Valodim: makes sense, thanks for clearing that up :+1:

OpenKeychain/src/main/res/xml/keyserver_sync_adapter_desc.xml
adithyaphilip: Done in https://github.com/open-keychain/open-keychain/commit/d461b424d2ddb7b71a73568fcf006c2a09829768

adithyaphilip:
Just pushed the latest version. Current sync rules are:
1. A sync will not be initiated if the screen is on. Instead it will be postponed by 5 min.
2. A sync once initiated (the screen was off) will be cancelled only if the user opens the OpenKeychain application. A cancelled sync is again postponed by 5 min.
3. If the user has chosen Orbot for proxy AND disabled silent starts in Orbot, a notification is displayed asking him whether he wants to Start Orbot or connect Directly (should we leave the latter option in at all, since the update is pretty sensitive?). The sync will resume 10 secs after the user presses "Start Orbot" (currently there is no reliable way of saying when Orbot has started if silent starts are disabled). In this case, the sync will happen even if the screen is on, but will be cancelled if the user opens the OpenKeychain app.
4. If Orbot is chosen as proxy, a staggered sync as performed by `parcimonie` is currently implemented. This unfortunately takes too long, and I'll be trying to get around it via https://github.com/open-keychain/open-keychain/issues/1469 as soon as possible.

Also, I've left the testing values for time in as of now:
* A sync is triggered every 1 minute instead of every few days (at https://github.com/open-keychain/open-keychain/commit/3530646d0dff057af3b22ca4c6a51d19df014bf4#diff-42bfea51b1889fb19f510328f027e88eR160)
* A postponed sync is delayed by 30s not 5 min (at https://github.com/open-keychain/open-keychain/commit/3530646d0dff057af3b22ca4c6a51d19df014bf4#diff-ef1fc82f02e79c5bb66b295a03418865R208)
* Keys that are older than 1s are updated, instead of 1 week (at https://github.com/open-keychain/open-keychain/commit/3530646d0dff057af3b22ca4c6a51d19df014bf4#diff-ef1fc82f02e79c5bb66b295a03418865R338)
* Staggered sync time variation is reduced to max 2 min (https://github.com/open-keychain/open-keychain/commit/3530646d0dff057af3b22ca4c6a51d19df014bf4#diff-ef1fc82f02e79c5bb66b295a03418865R275)

dschuermann:
Great.
Can you add a preference into our SettingsActivity that is doesn't persists another state but simply enableds/disables the sync for the keyserver sync adapter in Android?
While we are at it: We could add a similar preference to enable/disable linking into the contacts. Should be easy if you did the first pref.
Some people look into our preferences and don't find the account settings :)

adithyaphilip:
Yup, I remember the issues about them. Was already on it. Done in https://github.com/open-keychain/open-keychain/commit/dca6de579f1deb7dc445e294afd6f367c689eed8 ^^

dschuermann:
Can you put the discussed time constants in a central file, maybe on top of KeyserverSyncAdapterService as public statics? This makes it easier to adjust them later.

dschuermann:
* If I disable, then re-enable the sync and then turn of the screen, it is properly postponed and I see the update in logcat.

* However, I don't see the periodic sync every 1 minute.
Not sure if it has something to do with https://commonsware.com/blog/2015/06/12/psa-another-alarmmanager-regression.html and https://commonsware.com/blog/2015/03/23/alarmmanager-regression-android-5p1.html . I am running 5.1.1 CM nightly

* We should stop spamming the logcat with errors when a key is not found in the keyserver, we should change this from error to debug log.
```
Keychain D  E  Failed to get key at HkpKeyserver
  1370             Keychain D  E  org.sufficientlysecure.keychain.keyimport.HkpKeyserver$HttpError: 404: <?xml version="1.0" encoding="utf-8"?>
  1370             Keychain D  E  <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd" >
  1370             Keychain D  E  <html xmlns="http://www.w3.org/1999/xhtml">
  1370             Keychain D  E  <head>
  1370             Keychain D  E  <title>No results found</title>
  1370             Keychain D  E  <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
  1370             Keychain D  E  <style type="text/css">
  1370             Keychain D  E  /*<![CDATA[*/
  1370             Keychain D  E  .uid { color: green; text-decoration: underline; }
  1370             Keychain D  E  .warn { color: red; font-weight: bold; }
  1370             Keychain D  E  /*]]>*/
  1370             Keychain D  E  </style></head><body><h1>No results found</h1>No results found: No keys found</body></html>
  1370             Keychain D  E  at org.sufficientlysecure.keychain.keyimport.HkpKeyserver.query(HkpKeyserver.java:236)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.keyimport.HkpKeyserver.get(HkpKeyserver.java:361)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:222)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:119)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:476)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:468)
  1370             Keychain D  E  at java.util.concurrent.FutureTask.run(FutureTask.java:237)
  1370             Keychain D  E  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
  1370             Keychain D  E  at java.util.concurrent.FutureTask.run(FutureTask.java:237)
  1370             Keychain D  E  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
  1370             Keychain D  E  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
  1370             Keychain D  E  at java.lang.Thread.run(Thread.java:818)
  1370             Keychain D  E  query failed
  1370             Keychain D  E  org.sufficientlysecure.keychain.keyimport.Keyserver$QueryFailedException: not found
  1370             Keychain D  E  at org.sufficientlysecure.keychain.keyimport.HkpKeyserver.get(HkpKeyserver.java:364)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:222)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:119)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:476)
  1370             Keychain D  E  at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:468)
  1370             Keychain D  E  at java.util.concurrent.FutureTask.run(FutureTask.java:237)
  1370             Keychain D  E  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
  1370             Keychain D  E  at java.util.concurrent.FutureTask.run(FutureTask.java:237)
  1370             Keychain D  E  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
  1370             Keychain D  E  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
```

dschuermann:
Another idea: Could you implement a boolean in Constants that allows to switch between a set of time values for debugging/testing and set of values for production release?

adithyaphilip:
Finally figured out what the issue was and why I couldn't reproduce it - I was adding the periodic sync along with the Contact sync in the account creation code. If we were just updating the application, that bit wouldn't run. Fixed in https://github.com/open-keychain/open-keychain/commit/ac66471f1179c2d264e9fb641707463f7eae191b, along with:
* Constants which depend on debug/release build introduced in KeyserverSyncAdapterService
* Error logs changed to debug for key not found
* Fixed Todos from https://github.com/open-keychain/open-keychain/issues/1463

Valodim:
random thought, did you test what happens when a consolidate operation is triggered, particularly after the database is upgraded?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
```diff
@@ -136,20 +136,28 @@ public boolean checkConsolidateRecovery() {
...
                     ContentResolver.setSyncAutomatically(account, ContactsContract.AUTHORITY, true);
+                    KeyserverSyncAdapterService.enableKeyserverSync(context);
                 } else {
                     Log.e(Constants.TAG, "Adding account failed!");
                 }
+            } else {
+                // users updating to the version with sync will already have the account created
+                if (!Preferences.getPreferences(context).isKeyserverSyncInitialized()) {
+                    KeyserverSyncAdapterService.enableKeyserverSync(context);
+                }
```
dschuermann: Instead of using this check and introducing SYNC_KEYSERVER_SETUP, I would prefer that you increase Constants.default.PREF_VERSION and do ``KeyserverSyncAdapterService.enableKeyserverSync(context);`` in Preferences.upgradePreferences.
I know, strictly speaking, enabling the sync adapter is not really a new preference, but because we already have an upgrade mechanism, it fits there.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
```diff
@@ -0,0 +1,512 @@
...
+import java.util.GregorianCalendar;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class KeyserverSyncAdapterService extends Service {
+
+    // how often a sync should be initiated, in s
+    public static final long SYNC_INTERVAL =
+            Constants.DEBUG ?  TimeUnit.MINUTES.toSeconds(2) : TimeUnit.DAYS.toSeconds(3);
```
dschuermann: Please introduce something like "Constants.DEBUG_KEYSERVER_SYNC", similar to DEBUG_LOG_DB_QUERIES/DEBUG_SYNC_REMOVE_CONTACTS, and use this as a toggle. I don't like to have my debug build sync the whole time :P

dschuermann:
After implementing the comments, please remove/change your "PHILIP" debug logs.
It works fine so far. Future problems can be fixed with new pull requests.

dschuermann:
:tada:

adithyaphilip:
yay ^^
@Valodim yup, seems to work fine. Was there something specific I should have checked for? The bit at https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/ProviderHelper.java#L1275 retains the updated key times during a consolidation, though this is empty when the database gets upgraded. 
Just upgrading from v11 to v12 should be sufficient to test this, I guess, since a consolidate is triggered right after the upgrade?

PR #1465, opened 2015-07-27
Work in progress: Min crypto requirements

dschuermann:
WIP: don't merge

This is my branch that soon (tm) implements checks for insecure cryptography.
Consider the following attack: Your communication partner uses a 512 bit RSA key to sign a message. A three letter organization can forge such signatures due to their small bit length.
Currently there is no indication in OpenKeychain that this is a problem.
This branch should fix this and rework some result handling and API problems on its way.

I made a pull request so that everyone can see the progress here.


Valodim:
sweet :+1:

dschuermann:
I think the most important checks are in place now. Will now work on tests...

Issue #1470, opened 2015-08-05
App crashes when click "skip setup"


YodaTheMagistr:
App crashes when click "skip setup" or "configure key". If i click "skip setup", app will never run again (only after reinstall).
App builded from source (latest). Build without errors. What kind of information i need to provide? 
 
Error log
'''
08-05 19:02:19.289  13241-13241/org.sufficientlysecure.keychain.debug E/AndroidRuntime﹕ FATAL EXCEPTION: main
    Process: org.sufficientlysecure.keychain.debug, PID: 13241
    java.lang.RuntimeException: Unable to start activity ComponentInfo{org.sufficientlysecure.keychain.debug/org.sufficientlysecure.keychain.ui.EditKeyActivity}: android.view.InflateException: Binary XML file line #12: Error inflating class TextView
            at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2192)
            at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2241)
            at android.app.ActivityThread.access$800(ActivityThread.java:141)
            at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1204)
            at android.os.Handler.dispatchMessage(Handler.java:102)
            at android.os.Looper.loop(Looper.java:136)
            at android.app.ActivityThread.main(ActivityThread.java:5049)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:515)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:793)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:609)
            at dalvik.system.NativeStart.main(Native Method)
     Caused by: android.view.InflateException: Binary XML file line #12: Error inflating class TextView
            at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:716)
            at android.view.LayoutInflater.rInflate(LayoutInflater.java:758)
            at android.view.LayoutInflater.rInflate(LayoutInflater.java:761)
            at android.view.LayoutInflater.inflate(LayoutInflater.java:495)
            at android.view.LayoutInflater.inflate(LayoutInflater.java:400)
            at android.view.LayoutInflater.inflate(LayoutInflater.java:353)
            at org.sufficientlysecure.keychain.ui.EditKeyFragment.onCreateView(EditKeyFragment.java:131)
[...]

PR #1474, opened 2015-08-05
Fixed crash on api < 21 due to drawable using attrs

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1470
According to http://stackoverflow.com/questions/8041537/how-to-reference-style-attributes-from-a-drawable/13471695#13471695, drawables can't refer to an attribute. The workaround implemented from there involves having a separate drawable xml for each theme. 
Changed all the drawables I could find that do this (as tracked down from a usage search of items in attrs.xml). Hope that covers everything.


dschuermann:
great catch!

Issue #1300, opened 2015-05-24
Redesign Create Key UI


Hiperzone:
The Create Key UI needs to be redesigned in order to accommodate the new "Passphrase Alternatives" unlocking methods.

Serveral layouts are already planned which can be consulted in the following link:
https://cld.pt/dl/download/f7b037a7-bc13-46a5-8fb4-4cfb15b62b7a/openkeychain_ui_sample.pdf

These layouts will be implemented by using the following components.
-ViewPager where the user can swipe between each unlocking method.
-Adapter to give support to the ViewPager.
-Fragments for each unlocking method with the corresponding layout seen in the pdf. Various techniques may be used to avoid duplicate the same fragment (one with a image and one where the user actually uses the unlocking method).

The needed functionality of each fragment will be described on different issues as also the re-factoring needed for CreateKey Activity and the communication between the Activity and its fragments.
I'm open to suggestions as always.

Issue #1299, opened 2015-05-24
Passphrase Alternatives


Hiperzone:
Currently OpenKeyChain only supports one kind of key unlock method which is the Passphrase. By taking advantage of new security features currently present in many smartphones, OpenKeyChain will allow users to choose new security options which will be available by displaying a Key Unlocking Method Wizard to the user.

Security methods may range from the already known Passphrase method to NFC, Fingerprint, QRCode, LockPattern  or combinations of two security methods.

A complete redesign of the Create Key UI flow will be done to accommodate the new changes with a new UI design.
Implementation of Android UI components for each unlock method so they are reusable for all situations.

thetaeo:
It would be interesting to work on fingerprint sensors using the Android 6.0 API (I have a Nexus 5X for testing, and have gotten used to the extremely painless unlock procedure). It appears that the API is based off of creating a private key, then requiring a fingerprint to perform operations with it (see [this sample](https://github.com/googlesamples/android-FingerprintDialog/blob/master/Application/src/main/java/com/example/android/fingerprintdialog/MainActivity.java#L103-L138)). I would be willing to work on this.

dschuermann:
@thetaeo Our new plans are to replace individual key protection with a full app lock mechanism with sqlcipher. Thus, before adding protection mechanisms, we need to implement app lock. Please read and comment at https://github.com/open-keychain/open-keychain/issues/1541 for more information.

Valodim:
Superseded by #1541

PR #1481, opened 2015-08-13
Passphrase alternatives feature

Hiperzone:
This pull request includes all the passphrase alternative features like the Wizard  (#1300)
, NFC, Pattern and Pin Passphrase support (#1299).
This is the definitive pull request and the first candidate for a merge.
Any future corrections will be updated here ( no more new branches :P )

There are two conflits with the latest OpenKeyChain changes so I will wait to see what you want me to do about them.

Another thing, I don't have experience with git sub modules so I'm not sure if I have to commit the sub module (which I have yet to discover how) or if just adding it to git is enough plus the gradle changes to use it. I will wait feedback about it here. So for now do not merge the code yet.


dschuermann:
I forked the pattern git into openkeychain for you: Please use https://github.com/open-keychain/android-patternview.git

Hiperzone:
Previous crashes are fixed. I would like to talk you tomorrow if possible :)

dschuermann:
Haven't looked into the details of this pr, a small comment already:
We decided against the word "passphrase" as explained here: https://github.com/open-keychain/open-keychain/wiki/Language-Conventions#openkeychain-terminology

Hiperzone:
Alright, reverting back then. Sorry i couldn't reach you today, I only woke up a few hours later.

Hiperzone:
I already talked to Valodim so you don't need to worry :)

Valodim:
- [ ] pin: there is no back button
- [ ] pin: I wonder if inline keys are really a good idea. there is a pin-mode soft keyboard which could be used, check out the pin input in android settings
- [ ] pattern: confirmation should give more feedback about the current state, check out the pattern input in android settings
- [ ] pattern unlock: unlock button is unnecessary, just check if pattern is valid when it's finished drawing
- [ ] pattern: extra step with just the checkmark seems unnecessary
- [ ] nfc: the indeterminate progress bar is confusing, it is usually used when the user is supposed to wait for a process to finish, not when action is required
- [ ] nfc: green exclamation point icon looks wrong
- [ ] nfc: what happens when I use a tag which has data on it? is it overwritten?
- [ ] nfc: openkeychain crashes when I hold a tag to it in other dialogs
- [ ] nfc unlock: wrong tag gives no proper feedback
- [ ] nfc unlock: "please wait while tag is configured" shows up while processing
- [ ] "enter" button on the keyboard does nothing in the wizard
- [ ] wizard fragments should be layouted so the "next" button isn't hidden by the keyboard
- [ ] unlock: layout is a little misaligned, it seems slightly too wide
- [ ] wizard: hitting back forgets all data input, e.g. typing a mail address then moving back to name. not sure if it should remember tho, @dschuermann?
- [ ] wizard: please add password option during testing, we can always remove it again later

dschuermann:
I think it's okay that the wizard only remembers values when pressing "next".

Hiperzone:
Alright,
-About the pin changes, while initially I preferred the keyboard on screen style, I understand Valodim reasons for seeking a different input method style. In this case using the android keyboard could work and would then replace the current one with an edit text instead. This should solve all the pin issues that were mentioned. This change should be quite fast and won't take much time.

-About the pattern confirmation feedback, I'm not sure yet where I can improve it. I could rephrase the texts to be more clear but comparing to the android one I don't see any additional advantages in trying to add more states just to detect when the user is drawing and when the finger is out.

-About the confirmation mark icon, I find it to be necessary in my honest opinion because it indicates that the pattern/pin was input correctly instead of just advancing to the next step. I could hide it if that's if you want. Currently it seems in some android versions the tint is not applied even if it's in compact mode so my solution is to tint it myself in Photoshop.

-For the removal of the unlock button while on pattern dialog, that's a easy fix.

-About the green exclamation mark, I did not find any suitable icon for "information", if you could point one to me that would be great :)

-For the nfc data overwrite, if there is data present, it will be overwritten if Mifare tech is used. Currently I don't know how I can check if the card is blank. While i could check if the page has only zeros, I wonder how reliable is that since that could mean valid information for other applications.

-I cannot seem to produce any of the mentioned nfc crashes on other dialogs which does not make even sense to me since the other dialogs don't use the nfc dispatcher. I just tested here and it does not crash on pin/patttern or even on the wizard.

-Will fix the lack of feedback on nfc dialog.

-"please wait while tag is configured" does no seem to show up here when being configured since its a textview replacement. Can you take a screenshot for me?

-The enter button never did anything before. I'm going to take a good guess here on the reason for that. That button is not a reliable source of triggering the "next" button because there are wizard screen steps with multiple textviews, so my question is, which one triggers the next button? should it jump to the next textview? I'm not sure if I will have time today to attempt to fix this issue since it needs some medium functionality implementation on the wizard and I'm on a deadline.

-About the wizard navigation buttons not being visible when the keyboard opens, I will look into it, should be a simple configuration adjustment.

-Will fix the unlock dialog layout misalignments, this as caused by a fix i did on the wizard layouts. I'm probably going to separate the layouts into two files because this is causing more headaches than it is needed.

-Will add the password back to the unlock choice step :)


Going to start fixing the minor things listed here while I wait for your feedback.

Valodim:
- pin keyboard in activity is fine, but it's text input with no (password-) text field. there should be a textfield
- with "other dialogs" I meant keylist or public key view
- the enter button works for me as expected in current release: switches to next input field, or advances wizard
- "please wait while tag is configured" shows up during *unlock*

Hiperzone:
Going to look at these issues starting today.

Hiperzone:
-""please wait while tag is configured" shows up during unlock " is already fixed on the last commit.
Fixing the keyboard enter button and the wizard navigation buttons at the moment. Seems I've to change the way they were implemented because i can't seem to get them above the keyboard no matter what.

Currently I've been busy with my personal life and also took some days off because I will be moving to England really soon so expect these fixes till saturday if that is fine :)

Hiperzone:
Also can you rephrase or explain "pin keyboard in activity is fine, but it's text input with no (password-) text field. there should be a textfield"
 I'm having some trouble in understanding what you mean by that.

Valodim:
superseded by #1541. code from this pr might still prove useful if we go ahead with an implementation there

PR #1486, opened 2015-08-17
WIP: hidden recipients

dschuermann:
Work in progress

idea:
- If 0L try with all secret keys
- Do decrypt operation with key 0, if not successful, restart operation
- Then try with key 1, and so on.

problems:
- The current index is not properly persisted between restarts


Valodim:
rebasing this would be a complete rewrite at this point, and this feature is not very high on our list of priorities.

Issue #1128, opened 2015-03-10
Transifex client:  'ascii' codec can't encode characters


dschuermann:
transifex client stops with "Error compiling the translation file: 'ascii' codec can't encode characters in position 33-37: ordinal not in range(128)"

I can skip with ``tx pull -af --skip``: Then compiling works.
```
Pulling translations for resource open-keychain.strings (source: OpenKeychain/src/main/res/values/strings.xml)
 -> fr: OpenKeychain/src/main/res/values-fr/strings.xml
 -> bg: OpenKeychain/src/main/res/values-bg/strings.xml
 -> zh: OpenKeychain/src/main/res/values-zh/strings.xml
 -> is: OpenKeychain/src/main/res/values-is/strings.xml
 -> de: OpenKeychain/src/main/res/values-de/strings.xml
 -> it: OpenKeychain/src/main/res/values-it/strings.xml
 -> cs: OpenKeychain/src/main/res/values-cs/strings.xml
 -> et: OpenKeychain/src/main/res/values-et/strings.xml
 -> eu: OpenKeychain/src/main/res/values-eu/strings.xml
 -> fi: OpenKeychain/src/main/res/values-fi/strings.xml
 -> ja: OpenKeychain/src/main/res/values-ja/strings.xml
 -> es: OpenKeychain/src/main/res/values-es/strings.xml
 -> ru: OpenKeychain/src/main/res/values-ru/strings.xml
 -> nl: OpenKeychain/src/main/res/values-nl/strings.xml
 -> pt: OpenKeychain/src/main/res/values-pt/strings.xml
 -> sr: OpenKeychain/src/main/res/values-sr/strings.xml
Error compiling the translation file: 'ascii' codec can't encode characters in position 33-37: ordinal not in range(128)
 -> tr: OpenKeychain/src/main/res/values-tr/strings.xml
 -> sv: OpenKeychain/src/main/res/values-sv/strings.xml
 -> uk: OpenKeychain/src/main/res/values-uk/strings.xml
 -> sl: OpenKeychain/src/main/res/values-sl/strings.xml
 -> zh_TW: OpenKeychain/src/main/res/values-zh-rTW/strings.xml
 -> ro: OpenKeychain/src/main/res/values-ro/strings.xml
 -> pl: OpenKeychain/src/main/res/values-pl/strings.xml

```

vectorijk:
[...]

PR #1495, opened 2015-08-20
Fix pulling Serbian translation from Transifex

pejakm:
[History review of translation file for Serbian language](https://github.com/open-keychain/open-keychain/commits/master/OpenKeychain/src/main/res/values-sr/strings.xml) shows that Serbian translations aren't updated since February 28th. This should fix things.

##### Changes #####

* added `sr: sr` to the `lang_map` array
* removed `sr_CS: sr-rCS` from `lang_map` array (it's deprecated long ago)


dschuermann:
thanks!

dschuermann:
Hmm, I am not sure this is the real problem here. sr: sr shouldn't be necessary in this map as it should come automatically with the transifex tool. de: de is also not included for example.
I think it's more this problem: https://github.com/open-keychain/open-keychain/issues/1128

dschuermann:
@pejakm Sry, but I reverted your commit. It is definitly #1128

pejakm:
Oh, sorry, I wasn't aware of this. Is there a way to figure out which character causes this? (Although I'm suspecting on `ô`.)

dschuermann:
@pejakm np. I am not sure, I think someone needs to debug transifex-client.

Issue #1480, opened 2015-08-10
Crash on "update all": RuntimeException in ImportOperation.java:483


dschuermann:
OpenKeychain from git master crashes when I hit "update all"

Assigned to @adithyaphilip 

I got the following log, looks like a key is not on the keyserver:

```
[remved, see comment]
```

dschuermann:
Ah sry this wasn't the right log. This is the right one:
```
 13107             Keychain D  V  query(uri=content://org.sufficientlysecure.keychain.debug.provider/key_rings/-4266035947009897926/unified, proj=[has_any_secret, verified, privkey_data])
 13107             Keychain D  V  query(uri=content://org.sufficientlysecure.keychain.debug.provider/key_rings/unified, proj=[_id, master_key_id, user_id, is_revoked, is_expired, verified, has_any_secret, has_duplicate_user_id, fingerprint, creation])
 13107             Keychain D  D  insert(uri=content://org.sufficientlysecure.keychain.debug.provider/key_rings/4165106460550059802/keys, values=can_certify=false is_revoked=false can_authenticate=false master_key_id=4165106460550059802 fingerprint=[B@11c0b71b rank=1 can_sign=false key_curve_oid=null expiry=1451624722 can_encrypt=true key_id=-7185159641852011494 key_size=4096 algorithm=1 creation=1314575929)
 13107             Keychain D  D  Found class: class org.spongycastle.openpgp.PGPPublicKeyRing
 13107             Keychain D  V  log: LogEntryParcel{mLevel=DEBUG, mType=MSG_IMPORT_FETCH_KEYSERVER_OK, mParameters=null, mIndent=3}
 13107             Keychain D  V  log: LogEntryParcel{mLevel=DEBUG, mType=MSG_IMPORT_FINGERPRINT_OK, mParameters=null, mIndent=2}
 13107             Keychain D  V  log: LogEntryParcel{mLevel=OK, mType=MSG_IP_SUCCESS, mParameters=null, mIndent=5}
 13107             Keychain D  V  log: LogEntryParcel{mLevel=START, mType=MSG_IP, mParameters=[0xb540b362b202db50], mIndent=5}
 13107             Keychain D  V  query(uri=content://org.sufficientlysecure.keychain.debug.provider/key_rings/-5386107917860414640/unified, proj=[has_any_secret, verified, pubkey_data])
[...]

PR #1501, opened 2015-08-24
Fixed crash during "update all"

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1480 and provides a cleaner solution to https://github.com/open-keychain/open-keychain/issues/1221. 

Issue was due to `mProviderHelper.clearLog()` being called in parallel with `mProviderHelper.savePublicKeyring()` during multi-threaded key import. No wonder it was so hard to reproduce.


dschuermann:
nasty bug :D thanks for fixing!

PR #1534, opened 2015-09-16
Hidden recipient

bashrc:
So here's a proposed implementation for hidden recipient. If the key ID cannot be obtained then each subkey for each allowed key is tried.

Admittedly for the user this could potentially be laborious if they have multiple keys, but for common use cases in which there is only a single keypair it should be reasonably convenient.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedKeyRing.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedKeyRing.java
@@ -106,6 +106,14 @@ public abstract class CanonicalizedKeyRing extends KeyRing {
         return result;
     }
 
+    public Set<Long> getKeyIds() {
+        HashSet<Long> result = new HashSet<>();
+        for(CanonicalizedPublicKey key : publicKeyIterator()) {
+            result.add(key.getKeyId());
+        }
+        return result;
+    }
+
     public long getEncryptId() throws PgpKeyNotFoundException {
         for(CanonicalizedPublicKey key : publicKeyIterator()) {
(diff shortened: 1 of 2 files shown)


Valodim:
this code does not actually check if the content can be decrypted with a particular key, it effectively only uses a random one from the user's available keys.

the actual problem with this feature (which isn't solved in this PR) is that we can only know whether a private key is suitable for decrypting a message after we obtained its passphrase, which leads to a lot of complexity because the decryption routine may need to request many passphrases, and keep the state inbetween.

Even with a working implementation of the above mentioned data flow, I'm not sure the added complexity to our decryption routine is even worth it - hidden recipients are a rarely used feature. 

On the up side, support for hidden recipients will become trivial if we decide to use app lock over passphrases (check out #1541), so I'm postponing this until then.

dschuermann:
@Valodim Unfortunatly, it will not become trivial due to Yubikeys. I started https://github.com/open-keychain/open-keychain/pull/1486 to do something like:
* If 0L try with all secret keys
* Do decrypt operation with key 0, if not successful, restart operation
* Then try with key 1, and so on.

bashrc:
I have tried it with two different keys with different passphrases, and that does appear to work. Once it arrives at a decode using the given passphrase then it exits from the loop.

Anyway, having a single passphrase for all the user's keys would make sense, although it still wouldn't get around the need to try all available keys.

Valodim:
like I said, your code does not check if the key can actually decrypt the data, it only selects a key from the list of available ones (presumably by database order, so something like smallest key id).

@dschuermann it's a lot easier still, we can simply try all keys and if none works ask the user for his yubikey. this will succeed if the user holds the right yubikey (regardless of bound key), or fail as expected otherwise

PR #1540, opened 2015-09-20
Mousehints

rohands:
I've implemented them for just four views. Kindly tell me if there are more so that I can implement the same. 
Also, I am a beginner to Open Source World. I hope you guys will help me in my further contributions. Thank you

### DIFF
--- a/OpenKeychain/src/debug/res/values/strings.xml
+++ b/OpenKeychain/src/debug/res/values/strings.xml
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="app_name">"OpenKeychain (Debug)"</string>
+    <string name="nfc_not_supported">"Seems like NFC is not supported in your device."</string>
+
 </resources>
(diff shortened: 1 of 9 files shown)


Valodim:
Hi, thanks for contributing!

Did you see that the hint texts can be placed in the contentDescription attributes directly in the xml files? That way, the setup() call doesn't need to include the getString() everywhere.

Also, where did you get the LongClick code from? It looks exactly the same as [here](https://gist.github.com/hastoukopsaro/7910233), but without all the comments and, more importantly, license? It's apache license so we can just include it, but removing license on other people's code is a no-go I'm afraid.

Valodim:
seems you added an extra commit which isn't related to this pull request? please keep pull requests clean, makes it harder to merge otherwise

rohands:
Ah yes Vincent. I added ab extra commit.
Will note this point and make sure I don't in my further contributions :)
I can switch to an earlier commit and make a pull request. That should do
right?

Valodim:
yup! closing this one then

PR #1542, opened 2015-09-21
Issue better nfc handling

rohands:
Hi,
I've to tried handle NFC being supported by making my own Exception Class called NFCNotSupportedException and alerting the user by raising an alert Dialog. Hope it helps.
I've also changed the return type of invokebeam() to suit my requirements. Pls inform me if I shouldn't be doing such things. I'll avoid in my future contributions


dschuermann:
Thanks for your pull request and sry that it took so long.
I opted for a more unobtrusive way of handling this by not showing the button when NFC is not available: https://github.com/open-keychain/open-keychain/commit/62918c571899855bc4a16a77c9f423ff24c7d036

PR #1543, opened 2015-09-22
Issue mouseover hints

rohands:
Changed class and string names.


Valodim:
wow I can't believe I forgot to merge this one for the last release. I tested it at some point and it was ready to go. A-mergin' right now. Sorry for that, thanks for your work :+1:

Issue #1541, opened 2015-09-20
switch from passwords to app lock mechanism


Valodim:
We had a couple of feature requests in the past for an "app lock" mechanism like many apps have. The idea is to encrypt all application data and require the user to enter a password to access the app at all, with a caching-mechanism just like we have for key passwords now.

We rejected this feature, arguing that:
- the key list isn't really secret, similar to the contacts app the user is better off using FDE if he wants to protect it
- the secret keys are password-encrypted already
- after import from gpg or other tools, the user can use their regular password, which seems like the natural thing to do

At this point however, other arguments have come up which favor app lock over key passwords. The implication of an app lock mechanism is that there is one password to access the app, and after that individual keys are not encrypted again.
- protecting the list of contact keys is actually a very valid use case, especially in repressive countries
- using s2k encrypted keys was adopted from established workflows in pgp, and I feel they stand more on tradition than merit. different passwords for different keys are a feature of questionable use
- moving the password input mechanism from cryptographic operations up to database unlock should significantly reduce code complexity in the crypto routines, which is always a good thing
- hidden recipients are very difficult to implement if we potentially have to ask the user for all key passwords to find the right key to use (#626)
- the s2k mechanism is slow by design, and this is actually becoming a problem for keys which were encrypted on a fast desktop machine (#1530). there are ways to remedy this, but they involve re-encrypting the key so would require password on import

Valodim:
Thought about this some more:

Both "encrypt everything" and "encrypt only my keys" are valid modes of operation, so we will need support for individual encryption of keys. Consequentially, we will need to provide the user with two options: Encrypt the entire database, or encrypt only secret keys. I'm not sure if we should make these independent options, i.e. encrypt database *and* secret keys with different passwords, I tend towards a "no", but leaving this up for discussion for now.

Another decision we will have to make is how to encrypt the secret keys for storage in the database. I think it's a little easier to encrypt them in a container instead of using s2k, because this moves all encryption logic into the database layer, essentially making it part of the key retrieval rather than a property of the key object itself as it is now.

What's nice is that we can split up the implementation into two parts which are almost entirely independent: Firstly, encryption of all secret keys with a (single) master password rather than individual passphrases, and secondly, encrypt the database itself. These are still rather large tasks!

For the former, there are a couple of sub-tasks - in no particular order:
- ask for master password as part of first time wizard
- ask for master password after update if none is set and re-encrypt all keys!
[...]

PR #1546, opened 2015-09-25
Encrypted export

Valodim:
in recent versions, we have taken some steps toward making "key export" more similar to "backup" features from other apps. the next step in this direction is to encrypt backups using a "backup code". this is a randomly generated (not user-chosen!) string with which the data will be symmetrically encrypted.

so far, this pull request introduces a new activity which displays a backup code, and asks the user to write it down then enter it to proceed, and finally shows "save" and "share" buttons when the code was input correctly.


Valodim:
spent some time cleaning up the export operation code, and externalizing the upload operation. next step: actually encrypted export

Valodim:
slowly getting there... some todos for later

- [ ] show save dialog post kitkat
- [ ] need to re-add option to hand bytes directly to uploadoperation, this is required for revoke&upload to work atomically
- [ ] the way the backup code is built from the EditText fields is a little silly, change this to a simple method

other thoughts:
- the decryptverify dialog should probably be able to recognize pgp data, right now it's considered mime data and parsed as text/plain.
- what about file name conventions? we need one filename for the encrypted file itself, and one for the encrypted data inside.
- should we still strip local signatures? this is a backup, not an export after all...

dschuermann:
proposal for...  
file name: openkeychain-backup-2015-09-28.gpg  
encrypted file name: keys.asc

Local sigs: For now I would say: yes, lets strip the local sigs. People will also use this backup to transfer keys to other computers so at least for now, we should strip them.

Valodim:
save file done, localization done, worked on screen size compatibility, and also handling onSaveInstanceState for orientation changes

I think this is pretty close now, at least on the "export" side of things. Code reviews and testing much appreciated. For testing, any backup code starting with "ABC" is accepted, and the passphrase actually used for encryption is always just "abc" instead of the full backup code.

Valodim:
Import works fine with the recent changes, at least for a first iteration.

The biggest problem I have left with the export is how the user is asked for all his keys' passphrases... it's not *terrible*, but I don't like the ui flow there very much either. any ideas?

I also meddled with the ImportKeysActivity a bit, (cleaning the code up and) hiding the search bar if the intent gives fixed input like a uri to list keys from.

dschuermann:
I think asking for all the passphrases is good enough for now. This is something that we can tackle later with https://github.com/open-keychain/open-keychain/issues/1541

* filename not consistent between >= kitkat and < kitkat. maybe "backup_" is enough as it gets too long
* use file ending from Constants.FILE_EXTENSION_PGP_MAIN
* symmetric passphrase encoding: We should discuss how to encode the backup code as the symmetric passphrase for being compatible with gpg. I would propose using it with uppercase and dashes (reason: WYSIW we use)
* this should be placed into our FAQ
* I looked at whiteout's backup function and they add a literal data packet before the encrypted keys: https://github.com/whiteout-io/mail-html5/wiki/Secure-OpenPGP-Key-Pair-Synchronization-via-IMAP#packet-structure-version-1 Not sure if we want to do something similar, just an observation.
* I couldn't find out if they include the dashes inside the backup code, but they use uppercase: "random high-entropy uppercase alphanumeric string of 24 characters". Maybe we can ask them or look into their code if they include dashes or not.

Restore should be easy (at least to get it to a "good for now" state):
* rename drawer item to "Backup/Restore"
* add new button that opens file selection (same as in decrypt activity).
* check if it's a backup file based on the filename or on literal data packet?

Valodim:
I would definitely encode passphrases including the dashes, no reason to do anything else. We *could* add compatibility by trying different formats at decrypt times, but that could turn out quite confusing if the user tries the same passphrase in gpg then.

I have also considered adding a flag to the data about the passphrase. this would allow us to show a nicer passphrase prompt, and also inform an attacker about the passphrase format. However, if the attacker knows the origin of the file he can already know this, and we are talking `log_2(26^24) = 112` bits of entropy so that's pretty safe either way. But that's not a priority.

About whiteout's extra literal data packet, I wonder if that works with gpg?

Valodim:
checked out gpg's code, pretty sure `gpg --import` skips all non-key-related packets

Valodim:
making this one official

PR #1553, opened 2015-10-06
hidden recipient v2

bashrc:
Here is another variant of the hidden recipient patch. Some conditions have been added to avoid unneccessary checking or null values. It was tested with two email accounts having different keys and different passphrases.


Valodim:
this approach doesn't work for the general case. it relies on the allowedKeyIds parameter to provide the right key to use for decryption, and will fail if the first one in that set (which isn't even well-defined since it's a set) isn't the right one.

bashrc:
Yes you may be right.

Issue #903, opened 2014-09-29
Verify Yubikey signatures


Valodim:
Currently, we trust that the signed hash we get from a yubikey is correct and deliver it without further checks. I'd much rather throw an error than give an invalid signature here.

Validating is an expensive operation and we'd have to go through the data a THIRD time to do that - we however already have the data digest hash! Unless I get my crypto wrong here, we should be able to do "signature^d mod n" and if the signature is correct receive the hashed digest. I'm not sure if there is a way to coerce PGPSignature to just verify() without init() and update(), should look into that.

This is "nice to have" priority at best, just leaving this here for maybelater.

dschuermann:
@Valodim Is this still true?

Valodim:
Well it's true in principle, but I reckon a check for the key fingerprint on the yubikey is just as good...

PR #1564, opened 2015-10-23
key lookup improvements

Valodim:
so far, this contains working key lookup after decryption.

to actually verify the signature, the decrypt operation is re-executed after successful import, this *should* work in all relevant cases but maybe I'm missing something. it would be possible to store the signature hash, and re-verify without further input using a loader style refresh mechanism here - but that is a lot of work for relatively little gain.

\\ edit

looked at bouncycastle, and after wading through stuff like the `JcaPGPContentVerifierBuilderProvider` for a while I found that it is *not* easy to initialize a onepasssignature without having the public key in the first place. It might be possible to compute the digest entirely independently, but that seems error-prone. I'll leave it for now.


Valodim:
ok I wasn't happy with that previous state of things~

I managed to create a proof of concept `PGPNoKeyOnePassSignature` class which can be used in place of the `PGPOnePassSignature` class and requires the actual public key only at verification time, rather than initialization time. the bad news is, this is not compatible with the `java.security.Signature` api and requires meddling a level deeper than I'd usually prefer (namely, encoding the hash with algorithm identifier using ASN1/DER). the PoC code right now is extremely ugly and can be improved obviously, but I'm not sure it will ever be pretty.

on the up side, we can use this to cache message digests in operation results, which allows re-verification at a later point without going through the data again.

other thoughts:
* we could ditch one of the code paths used for OnePassSignature/Signature! that's pretty neat actually, if you look at PgpSignatureChecker there is a lot of unnecessary code duplication because those two signature classes just don't allow for shared code
* we can use this to improve handling of multiple signatures? given that nobody noticed that multiple signatures were broken until very recently (f6de2712d3edef9837a53da5d78a9daa28639af4), that's probably not too useful
* we could provide a "check signature hash" api call. I'm not sure if it's useful for anything, but the more interesting operation would be the converse "sign some hash" one, which could be used for authentication purposes
* we can use the technique to pick up on #903, I guess

Valodim:
moved work from the previous comment into a different branch (v/PGPNoKeyOnePassSignature)

Valodim:
This works for me as it is, mergin'

Issue #1573, opened 2015-10-27
NullPointerException in KeySyncAdapterService


Valodim:
```
java.lang.RuntimeException: Unable to start service org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService@21f63140 with null: java.lang.NullPointerException
	at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:2759)
	at android.app.ActivityThread.access$2100(ActivityThread.java:138)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1333)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:149)
	at android.app.ActivityThread.main(ActivityThread.java:5061)
	at java.lang.reflect.Method.invokeNative(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:515)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:788)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:604)
	at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.NullPointerException
	at org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService.onStartCommand(KeyserverSyncAdapterService.java:80)
	at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:2742)
	... 10 more
```

the intent parameter seems to be null in `onStartCommand`. docs on the parameter:
> intent - The Intent supplied to Context.startService, as given. This may be null if the service is being restarted after its process has gone away, and it had previously returned anything except START_STICKY_COMPATIBILITY.

adithyaphilip:
Looking into it

dschuermann:
@adithyaphilip great :)

adithyaphilip:
We return `START_NOT_STICKY` from `KeyserverSyncAdapterService`, and the docs for it say:
>if this service's process is killed while it is started (after returning from onStartCommand(Intent, int, int)), and there are no new start intents to deliver to it, then take the service out of the started state and don't recreate until a future explicit call to Context.startService(Intent). The service __will not receive a onStartCommand(Intent, int, int) call with a null Intent__ because it will not be re-started if there are no pending Intents to deliver.

So I'm not sure what exactly is causing it. Will keep investigating. Is there a way to reproduce it?

[...]

PR #1578, opened 2015-10-30
Reworked KeyserverSyncAdapterService

adithyaphilip:
In `KeyserverSyncAdapterService`:
* Ignore null Intents and Intents with null actions to fix https://github.com/open-keychain/open-keychain/issues/1573.
* Introduced `START_REDELIVER_INTENT` to better handle service process being killed.
* Modified staggered Orbot update so that the first key is updated immediately. A large time difference between when Orbot is started and when the key is updated could result in starvation if the user turns Orbot off before even a single key could be updated (besides repeated Orbot starts and a lengthier update).

In `ImportOperation`:
* Corrected case where an early cancellation of multi-threaded import may be interpreted as `RESULT_FAIL_NOTHING`.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
@@ -540,7 +540,9 @@ public class ImportOperation extends BaseOperation<ImportKeyringParcel> {
 
             // adding required information to mResultType
             // special case,no keys requested for import
-            if (mBadKeys == 0 && mNewKeys == 0 && mUpdatedKeys == 0) {
+            if (mBadKeys == 0 && mNewKeys == 0 && mUpdatedKeys == 0
+                    && (mResultType & ImportKeyResult.RESULT_CANCELLED)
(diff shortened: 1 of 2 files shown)


Valodim:
I'm not so sure about REDELIVER_INTENT. the service is already executed periodically, isn't it? when does the redelivery happen?

adithyaphilip:
The redelivery is for when the system kills the service due to memory constraints. If we use `START_NOT_STICKY`, the system will not restart the service if it's killed in the middle of the sync and we will have to wait until the next time the sync is run (every three days). On the other hand, if we return `START_REDELIVER_INTENT`, the service will be restarted when the system feels it has sufficient memory.
http://stackoverflow.com/questions/9093271/start-sticky-and-start-not-sticky

Valodim:
I have no strong feelings one way or the other, but three days for the next try also seems ok to me. Care to tie-break here @dschuermann?

dschuermann:
I think it's good to switch to ``START_REDELIVER_INTENT``. We can then really promise that an update is done every 3 days. With ``START_NOT_STICKY`` we would claim something we don't abide to.

Valodim:
Cool, feel free to merge then

PR #1584, opened 2015-11-03
Inline Identity Edit

Valodim:
I added a way to edit identities inline in the extended information view as an action mode. It's a little rough around the edges still - icons aren't pretty, and "save" doesn't actually save anything.

This is more of a design suggestion than a definitely planned feature. Once we get rid of per-key passphrases, editing of user ids will be the only non-advanced use case left for key editing, maybe having that inline is a little neater? Not sure.

edit redesign screenshot


Valodim:
added subkey editing the same way.

I also moved menu handling into the activity, which allows for neato animation of the edit icon

dschuermann:
Rebased on current master. Will do some work on this branch during 32c3

Valodim:
Please give an advance notice when you merge larger changes like this in the future. No big deal, but it increases the chance that code gets reviewed, which is good for our codebase quality

PR #1597, opened 2015-11-15
Select time to remember password in dialog

Valodim:
Using different caching times at different times is quite the valid use case. If I know I'll be doing a lot of keysigning or mail reading in the next hour or so, for example, it should be in the few hours range. Otherwise, couple minutes is fine. But it really varies.

Another advantage is that it raises awareness that passwords are cached. Users are sometimes confused why they aren't asked their password on a second operation, if they havn't noticed the cache notification before.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
@@ -27,8 +27,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.os.Binder;
-import android.os.Build.VERSION;
-import android.os.Build.VERSION_CODES;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
@@ -349,14 +347,14 @@ public class PassphraseCacheService extends Service {
             case ACTION_PASSPHRASE_CACHE_ADD: {
                 long masterKeyId = intent.getLongExtra(EXTRA_KEY_ID, -1);
                 long subKeyId = intent.getLongExtra(EXTRA_SUBKEY_ID, -1);
-                long timeoutTime = intent.getIntExtra(EXTRA_TTL, DEFAULT_TTL);
+                long timeoutTtl = intent.getIntExtra(EXTRA_TTL, DEFAULT_TTL);
 
                 Passphrase passphrase = intent.getParcelableExtra(EXTRA_PASSPHRASE);
                 String primaryUserID = intent.getStringExtra(EXTRA_USER_ID);
 
                 Log.d(Constants.TAG,
                         "PassphraseCacheService: Received ACTION_PASSPHRASE_CACHE_ADD intent in onStartCommand() with masterkeyId: "
-                                + masterKeyId + ", subKeyId: " + subKeyId + ", ttl: " + timeoutTime + ", usrId: " + primaryUserID
+                                + masterKeyId + ", subKeyId: " + subKeyId + ", ttl: " + timeoutTtl + ", usrId: " + primaryUserID
                 );
 
                 // if we don't cache by specific subkey id, or the requested subkey is the master key,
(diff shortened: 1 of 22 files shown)


Valodim:
Screenshots of the current implementation:





I gotta say, I don't like it much. But I'm not sure how to improve the layout either.. I thought about something like a horizontal wheel picker type thing? Or maybe just have a `+` and `-` button next to the word? Ideas would be appreciated

dschuermann:
I like it :)
What about remembering the last selection for the next time the dialog is shown?

Valodim:
Yeah. Either that, or leave the default as a preference. I also considered changing "Five Minutes" to "Just Now" and remember the password until the next `Intent.ACTION_SCREEN_OFF` broadcast intent. Also, maybe change the granularity to One Hour/Day/Week? Really difficult to tell which granularity makes sense...

dschuermann:
Maybe leaving the default as a preference makes more sense in this case. It's more an exception than the rule to use long caching times. Remembering until screen goes off would be nice. I don't a specific opinion regarding granularity...

Valodim:
I would assume that most users have a notion of what constitutes short term and long term for them, but don't need more than two or three options. I'm thinking of a preference menu with a "pick two (or three?), and one default" semantic, and the user ends up with his chosen options plus "Just Now" in the dropdown. This would also make it relatively easy to try to reduce the set of default options, but also allow flexibility to the users

dschuermann:
Sounds great

Valodim:
created a settings activity with the above choice flow:

Valodim:
Q: in the current implementation, we "refresh" the passphrases when they are used. should we still do this?

I find this problematic for a couple of reasons: firstly, we can't even do it consistently anyways, for instance, apps shouldn't be able to refresh all the time to artificially extend the duration. secondly, this makes the durations quite dynamic and intransparent for the user - he picked one hour, why is this still in there after two?

Valodim:
added "until screen off" option.

I wonder if using a shorter cache time "than usual" is actually a valid use case. As in, do users ever choose "this time, cache *shorter* than usual"? If it isn't, we can ditch the default setting altogether and use the first as default

dschuermann:
Hmmm yeah, maybe using the first one as default is good enough. People who really prefer long cache times should just deactivate short times in the preferences.

Valodim:
removed the default option. this might be good to go now :+1:

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/PassphraseCacheService.java
```diff
@@ -285,10 +284,12 @@ private Passphrase getCachedPassphraseImpl(long masterKeyId, long subKeyId) thro
 
         }
 
+        // TODO do we still want this?
```
dschuermann: no we don't want this any more. Remove it.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SettingsCacheTTLActivity.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (C) 2010-2014 Thialfihar <thi@thialfihar.org>
```
dschuermann: wrong copyright header

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SettingsCacheTTLFragment.java
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright (C) 2012-2015 Dominik Schürmann <dominik@dominikschuermann.de>
+ * Copyright (C) 2015 Adithya Abraham Philip <adithyaphilip@gmail.com>
```
dschuermann: wrong copyright header

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/widget/CacheTTLSpinner.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (C) 2014 Dominik Schürmann <dominik@dominikschuermann.de>
```
dschuermann: wrong copyright header

OpenKeychain/src/main/res/layout/import_yubikey_new.xml
```diff
@@ -0,0 +1,241 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dschuermann: Why has this file been added?

dschuermann:
I did a small review, looks great! Just some minor comments.

Issue #1321, opened 2015-06-01
Support Facebook public key download


dschuermann:
https://www.facebook.com/notes/protecting-the-graph/securing-email-communications-from-facebook/1611941762379302
https://news.ycombinator.com/item?id=9637649

## Simple support (maybe this already works)
Register intent filter to process asc returned from
https://www.facebook.com/*/publickey/download
, e.g., https://www.facebook.com/steve.weis/publickey/download
(only works when logged in)

## Extended support:
* Use API of Facebook to search for keys?
* Maybe in additon to HKP and keybase.io?
* Use linked identities (see branch) to link back to Facebook?

dschuermann:
Externalized linked identity feature into https://github.com/open-keychain/open-keychain/issues/1623

PR #1610, opened 2015-11-25
Basic Facebook Support

adithyaphilip:
Related to https://github.com/open-keychain/open-keychain/issues/1321
This PR adds:
* Support for importing keys from links of the form:
https://www.facebook.com/*/publickey/download.
e.g. https://www.facebook.com/adithya.abraham/publickey/download <br>
Requirement: The visibility of the key must be set to "Public" on Facebook.
* Reloading of `ImportKeysActivity` if the activity was already open and a new Intent is received.

Testing Facebook search:
* Clicking on above link in browser (after which you may search for any Facebook username) (or)
* Add https://facebook.com as preferred keyserver (this works because of the URL inspection mentioned below) and searching for a person's Facebook username.

A query to Facebook is currently identified in `CloudSearch` and `ImportOperation` by URL inspection. I'm guessing this should be changed to something more centralized like adding a new boolean `searchFacebook` to `CloudSearchPrefs`, similar to `searchKeybase`. Working on that right now. Is everything else fine?

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -30,6 +30,7 @@ import android.view.ViewGroup;
 import org.sufficientlysecure.keychain.Constants;
 import org.sufficientlysecure.keychain.R;
 import org.sufficientlysecure.keychain.intents.OpenKeychainIntents;
+import org.sufficientlysecure.keychain.keyimport.FacebookKeyserver;
 import org.sufficientlysecure.keychain.keyimport.ImportKeysListEntry;
 import org.sufficientlysecure.keychain.keyimport.ParcelableKeyRing;
 import org.sufficientlysecure.keychain.operations.results.ImportKeyResult;
@@ -41,6 +42,7 @@ import org.sufficientlysecure.keychain.ui.util.Notify;
 import org.sufficientlysecure.keychain.util.Log;
 import org.sufficientlysecure.keychain.util.ParcelableFileCache;
 import org.sufficientlysecure.keychain.util.ParcelableFileCache.IteratorWithSize;
+import org.sufficientlysecure.keychain.util.Preferences;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -50,6 +52,8 @@ public class ImportKeysActivity extends BaseNfcActivity
 
     public static final String ACTION_IMPORT_KEY = OpenKeychainIntents.IMPORT_KEY;
     public static final String ACTION_IMPORT_KEY_FROM_KEYSERVER = OpenKeychainIntents.IMPORT_KEY_FROM_KEYSERVER;
+    public static final String ACTION_IMPORT_KEY_FROM_FACEBOOK
+            = Constants.INTENT_PREFIX + "IMPORT_KEY_FROM_FACEBOOK";
     public static final String ACTION_IMPORT_KEY_FROM_KEYSERVER_AND_RETURN_RESULT =
             Constants.INTENT_PREFIX + "IMPORT_KEY_FROM_KEY_SERVER_AND_RETURN_RESULT";
     public static final String ACTION_IMPORT_KEY_FROM_KEYSERVER_AND_RETURN_TO_SERVICE = Constants.INTENT_PREFIX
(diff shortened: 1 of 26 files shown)


adithyaphilip:
Search by username for Facebook is now working :) Added a preference below keybase.io for it.

adithyaphilip:
Reworked remaining preferences in https://github.com/open-keychain/open-keychain/commit/66ec7f17a6f89e8747e98f958eeed5c065621c33. Will squash appropriately when it's ready to be merged.

adithyaphilip:
I've added a base `PresetPreferenceFragment` in https://github.com/open-keychain/open-keychain/commit/cdca80806ad687aeca99e179f6897b3ceb69410c and cleaned up the remaining code. We didn't need the `PreferenceActivity` compatibility in `ProxyPrefsFragment` anymore so I removed that bit too (if I recall, it had something to do with letting `SettingsActivity` display the proxy preferences right away, which we don't seem to use anymore (or ever did?)). If this is okay, I'll go ahead and squash everything.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -113,6 +117,12 @@ protected void initLayout() {
         setContentView(R.layout.import_keys_activity);
     }
 
+    @Override
+    public void onNewIntent(Intent intent) {
```
Valodim: nice catch, I just recently wondered about our handling of singleTop

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -113,6 +117,12 @@ protected void initLayout() {
...
+        super.onNewIntent(intent);
+        handleActions(intent);
```
Valodim: there *is* `setIntent()` in Activity, I feel like the clean solution here is to set the activity's intent to the new one and call handleActions (which mostly manages fragment state) in `onResumeFragments` any time the intent changed. this also puts it in the correct lifecycle for the fragment manager, which is not yet initialized during `onNewIntent` (or `onResume`) but is initialized in `onResumeFragments`

see also [FragmentActivity](http://developer.android.com/reference/android/support/v4/app/FragmentActivity.html#onResume()) and [commit](http://developer.android.com/reference/android/app/FragmentTransaction.html#commit()) android docs

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -254,21 +271,26 @@ protected void onRestoreInstanceState(Bundle savedInstanceState) {
...
+    private void startListFragment(byte[] bytes, Uri dataUri, String serverQuery,
+                                   Preferences.CloudSearchPrefs cloudSearchPrefs) {
         Fragment listFragment =
-                ImportKeysListFragment.newInstance(bytes, dataUri, serverQuery, false, keyserver);
+                ImportKeysListFragment.newInstance(bytes, dataUri, serverQuery, false,
+                        cloudSearchPrefs);
         getSupportFragmentManager().beginTransaction()
                 .replace(R.id.import_keys_list_container, listFragment, TAG_FRAG_LIST)
-                .commit();
+                .commitAllowingStateLoss();
```
Valodim: I'm really not a fan of `commitAllowingStateLoss`, and the general android advice is to use it as a last resort only. see above on handling `onNewIntent`.

Valodim:
While this is a nice feature indeed, we should probably not have Facebook enabled by default. I can imagine that many users would be upset if they found out all of their search queries in OpenKeychain are submitted to Facebook (of all companies!)

Valodim:
do you have a username I can use to test this? I just tried some random names and none gave me any results

adithyaphilip:
Try adithya.abraham and amoghbl1. Isn't *exactly* a search though, since the usernames need to match completely.

As for the `commitAllowingStateLoss`, thankyou, that does sound a lot better. Will change and push soon.

Valodim:
Hmm. So it matches only by exact name? Does it match by e-mail address?

adithyaphilip:
Nope, matches by exact Facebook username. It's more a side effect of supporting Facebook URLs, "search" here just fills in your query in the place where the username should be in the URL for a public key. Search by e-mail, or a person's actual name, would require us integrating the Facebook SDK.

dschuermann:
I like the support of Facebook URLs, but given the limitations of the search and the fact that facebook is contacted everytime someone searches for a key I would propose removing the search feature. Can link handling be decoupled from search?

adithyaphilip:
Yup, search is pretty limited right now. Removed and squashed with https://github.com/open-keychain/open-keychain/commit/f29280bbb268d112426c6662e1227118819fb904. Also taken care of the `commitAllowingStateLoss`. Good to go?

Valodim:
didn't get to test it yet, but I looked over the code and found no issues

dschuermann:
Tested it and works great, nice work!

PR #1615, opened 2015-12-02
provider for OpenPgpApi

Valodim:
To display contact info in K-9, I needed some way to get data per email address. The straightforward approach is adding a Provider in addition to the OpenPgpService we offer, so that's what I went with.

This is still missing a permission check. I checked and it's possible to check the calling uid just like with other service calls, so it's easy to restrict to apps which are already allowed.

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -806,7 +806,12 @@
             android:exported="false"
             android:label="@string/keyserver_sync_settings_title" />
 
-        <!-- Internal classes of the remote APIs (not exported!) -->
+        <provider
+            android:name=".remote.KeychainExternalProvider"
+            android:authorities="${applicationId}.provider.exported"
+            android:exported="true" />
+
+        <!-- Internal classes of the remote APIs (not exported) -->
         <activity
             android:name=".remote.ui.RemoteCreateAccountActivity"
             android:exported="false"
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainExternalContract.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainExternalContract.java
@@ -1,0 +1,35 @@
+package org.sufficientlysecure.keychain.provider;
+
+
+import android.net.Uri;
+import android.provider.BaseColumns;
+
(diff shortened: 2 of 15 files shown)


dschuermann:
Because this is API only, I would propose moving it into the "remote" subpackage.

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -806,7 +806,12 @@
             android:exported="false"
             android:label="@string/keyserver_sync_settings_title" />
 
-        <!-- Internal classes of the remote APIs (not exported!) -->
+        <provider
+            android:name=".remote.KeychainExternalProvider"
+            android:authorities="${applicationId}.provider.exported"
```
dschuermann: what about using ``org.openintents.openpgp.provider`` as the authority to keep the package naming consistent for the API?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainExternalContract.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static class EmailStatus implements BaseColumns {
+        public static final String EMAIL_ADDRESS = "email_address";
+        public static final String EMAIL_STATUS = "email_status";
+
+        public static final Uri CONTENT_URI = BASE_CONTENT_URI_EXTERNAL.buildUpon()
+                .appendPath(BASE_EMAIL_STATUS).build();
+
+        public static final String CONTENT_TYPE
+                = "vnd.android.cursor.dir/vnd.org.sufficientlysecure.keychain.provider.email_status";
```
dschuermann: changing to ``"vnd.android.cursor.dir/vnd.org.openintents.openpgp.provider.email_status"`` ?

dschuermann:
finished looking at it, if everything still works as before, it should be good besides the authority naming.

PS: some license headers would be nice :)

Valodim:
The provider isn't a unique service - if I change the provider to openintents, api clients can no longer choose between providers.

see also: https://github.com/k9mail/k-9/blob/master/k9mail/src/main/java/com/fsck/k9/activity/compose/RecipientLoader.java#L259

dschuermann:
Ah that makes sense, then leave it like this.

Issue #1622, opened 2015-12-10
Automatic Key Updates Forced Set


RichJeanes:
Since the most recent update, Automatic Key Updates is enabling itself every time I turn it off. I have Enable Tor turned on, too, so every time the key update triggers, OpenKeychain connects to the Tor network but then I can't shut Tor down again until I force close OpenKeychain. Every time I try to shut it down, it reconnects immediately. Don't know if the two issues are related.

OpenKeychain 3.7 via Play Store
Nexus 5X on Verizon, Android 6.0 (MDB08L)

Valodim:
@adithyaphilip this one looks like one of yours :)

adithyaphilip:
Don't seem to be able to reproduce it (the preference refusing to turn off) on mine (5.1). Can anyone confirm if it's 6.0 specific (possibly related to 6.0 permissions)?

One reason for Tor repeatedly connecting is that we employ a staggered update strategy where keys are updated with an interval of 10-20 min, though it shouldn't cause it to immediately reconnect.

Will take a closer look at both issues soon.

chinkung:
I have this issue too.

Automatic key update turning on automatically even I try to turn it off.

OpenKeychain 3.7
Android 4.4.4

adithyaphilip:
@chinkung thank you, means it isn't 6.0 specific then. Do you also happen to have Tor enabled?

adithyaphilip:
Fixed with https://github.com/open-keychain/open-keychain/pull/1633

PR #1633, opened 2015-12-20
Fixed Keyserver and Contact Sync turning itself on

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1622. Looks like at some point keyserver sync and contact linking was enabled whenever the app started.
Also cancels a postponed/staggered sync if it was scheduled, but was disabled in preferences before it could trigger.


dschuermann:
Great work as always!

adithyaphilip:
Thank you! :)

PR #1639, opened 2015-12-22
Fixed #1636

### LINKED ISSUE #1636: EncryptKeyCompletionView: Deleting a token shows class name
1. Add 4 keys to EncryptKeyCompletionView
2. Focus other textview, now it is shown as "Test +3"
3. Focus EncryptKeyCompletionView. delete token:

device-2015-12-22-142501

rohands:
Overridden toString method() of KeyItem Class.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
@@ -65,7 +65,6 @@ public class EncryptModeAsymmetricFragment extends EncryptModeFragment {
         args.putLong(ARG_SINGATURE_KEY_ID, signatureKey);
         args.putLongArray(ARG_ENCRYPTION_KEY_IDS, encryptionKeyIds);
         frag.setArguments(args);
-
         return frag;
     }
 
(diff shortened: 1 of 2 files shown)


dschuermann:
What is the behavior after this patch?

rohands:
Whenever you delete a token, instead of getting the class name (KeyItem)
which is totally unacceptable, you get just his user name.
However, this happens on deleting every token.
A better solution could be to display the user name on deleting the most
recent token and on further deletions of tokens, we could just delete them
without showing the user name. This behaviour also is the default case.
What do you think would be the right choice?

dschuermann:
I think it would be great if pressing the delete key always completely deletes a token. This is also the current behavior when the field is not collapsed.

rohands:
Okay sure, I'll make changes to it right away.
What I am planning to do is to just return an empty string from toString().

On Wed, Dec 23, 2015 at 5:46 PM, Dominik Schürmann <notifications@github.com

dschuermann:
I just tested this fix and it does not solve the problem outlined in #1636 . I suspect that the issue in #1636 is a bug in the library itself because it works when TokenAutoComplete has not been collapsed.

rohands:
Please correct me if I'm wrong 
But I think the problem to be addressed was to not get class name on trying to delete the token after the focus had been changed atleast once. 
After quite some debugging I figured that it was calling KeyItem's toString() method. Whatever text is returned from toString() would appear when the token is deleted. 
The workarounds are 
1. You can set a certain deletion style by which only the text that was entered by the user before the token had been matched would remain on deleting the token OR 
2. The more obvious way by which you can make no text appear by deleting the token. (Default behaviour)

However I'd like to bring to your notice that there's a flaw in my code. 
You might be wondering why I've returned " "( string with whitespace and not an emptry string) from toString() of KeyItem class.
Well that's because on returning an empty string I'm not sure how but it seemed to have been conflicting with the functionality of the library. The " token matching" wasn't happening. 
But then, leaving things at a whitespace pretty much solves the problem right? The user will have to delete an extra character in contrast to the several characters which would appear in case class names appeared. 
I'd like to know where I went wrong in understanding
Thank you

Issue #1664, opened 2016-01-06
Key import from file broken since Facebook support pr


dschuermann:
Importing keys from files no longer loads the keys inside the import activity's list. Reverting https://github.com/open-keychain/open-keychain/commit/f29280bbb268d112426c6662e1227118819fb904 solves the issue.

@adithyaphilip Can you look into this?

adithyaphilip:
Figured it out - side-effect of having `handleActions` in `onResumeFragments`. `onActivityResult` is called before `onResumeFragments`, so the fragment it loads with the key data from the file, ends up being replaced by the latter.

This also meant that if a user minimizes the app after a cloud search the results would be wiped and he'd have to search again.

Fixed with https://github.com/open-keychain/open-keychain/pull/1666.

PR #1666, opened 2016-01-08
Fixed File Import issue due to onResumeFragments

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1664. Caused due to `handleActions` in `onResumeFragments`.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -21,6 +21,7 @@ import android.app.Activity;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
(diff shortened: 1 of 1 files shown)


Valodim:
looks good. remind me though, why do we need to handle `onNewIntent` in `ImportKeysActivity` again?

adithyaphilip:
Since we added `launchMode="singleTop"` to `ImportKeysActivity` in the manifest, we need `onNewIntent` to handle new Intents to the activity if it's already open e.g. if it's in the background and we try to import keys from a file using an external file manager.

Valodim:
next question then, why is it singleTop? From git history it's been that way for a long time, remember why we use that way @dschuermann?

dschuermann:
According to git blame, it has been singleTop since 2010 (sic!), so this has been introduced by Thialfiar o.O
So I think we can safely change that.

dschuermann:
@adithyaphilip Can you remove singleTop and change the activity accordingly? I think this is the last blocker before a new release :) *excited*

adithyaphilip:
@dschuermann yup, on it :)

dschuermann:
@adithyaphilip sry was faster :D

Issue #1536, opened 2015-09-16
Display UID comments in "Keys" list


nazir-cz:
Please add in the list of available keys UID comments next to UID name.
I have 11 keys here with the same UID name with different comment (and expiration dates), and at first sight it is not clear which key do I choose...

Would help display in the form:

```
Name (Comment)
Email
```

dschuermann:
We currently show the creation time in addition to the name and email if the user ids of keys are equal. Unfortunately, that is not true in your case because the comments inside the user ids are different.

I think we could change this and show the comment if there are more than one key with the same name/email combination but that requires more internal work. Because it is a corner case and not high prio, this issue has been tagged "help-wanted".

Valodim:
There is two individual problems here:

Firstly, *how* to disambiguate. At the moment, we show the creation date for disambiguation. This works well for the case where a user has multiple keys, and one of them is outdated while the other isn't. It does not work well for multiple valid keys for the same e-mail address which are intended for different purposes. This is a thing which requires further discussion, my gut thought though is that we don't want disambiguation by comments.

Secondly, *when* to disambiguate. This is a problem which is currently solved in the database with the HAS_DUPLICATE_USER_ID flag, and I think that's the right place for it. The problem is that we don't show comments in the key list, so we should not use them for disambiguation. While it's certainly possible to solve this simply by using a more complex query which ignores the comment part for disambiguation (`WHERE a.user_id LIKE CONCAT(SUBSTRING(b.user_id, INDEXOF(b.user_id, "(")), "%")` or something like that. ew!), my proposal would be to save name, email, and comment individually in the database *in addition to* the untouched user id string, and perform the query on those.

rohands:
So, what's the solution? 
Is it making changes to the database or displaying comments for user ids which return <b>1</b> on <code>HAS_DUPLICATE_USER_ID</code> ?
Also, I'd like to point out that when keys are imported from PC with User IDs which have the same name, email but different comments, counting on what value is stored in <code>HAS_DUPLICATE_USER_ID</code>  to solve the problem will fail because it will not show the creation date for these keys (obviously because user ids are different!).
It just displays the name and email id which could be difficult for the user to disambiguate at first glance.
Also, is @nazir-cz talking about the keys imported from PC or the ones imported through other methods?

rohands:
@Valodim are you waiting for more discussion on this issue regarding "How to disambiguate"?

Valodim:
[...]

PR #1667, opened 2016-01-09
Fix for 1536

rohands:
#1536 
Tested by importing keys from PC


rohands:
Why does it say some checks were unsuccessful?

Valodim:
That's test coverage. It just means the ratio of tested vs untested lines decreased, so don't worry about it.

We're currently working on a release, will look at this PR after we're done with that :+1:

rohands:
Sure sure.

Valodim:
Looked this over, and I think it solves the actual problem at the wrong level. I'll put details in the related issue since it might require more discussion

PR #1683, opened 2016-01-13
allow "panic button" apps to clear all passphrases

eighthave:
We've been developing "panic button" features in our apps for years, and now we are working to make a more flexible and open ecosystem of apps that can both trigger other apps, or respond to triggers. Since OpenKeychain already caches passphrases and includes a "clear" feature, it is a natural panic responder app. This setup allows for easy configuration of a single trigger action which can then trigger multiple apps.

The first commit adds support for receiving the panic trigger. The second commit wipes OpenKeychain from the Recent Apps when it is responding to a panic trigger.

I made a video to demonstrate the user experience, this pull request offers the same user experience as the demonstration with Orweb:
https://www.youtube.com/watch?v=mS1gstS6YS8

The first panic button app that sends this kind of trigger is called Ripple, and it is available here:
* https://github.com/guardianproject/ripple
* https://play.google.com/store/apps/details?id=info.guardianproject.ripple
* https://guardianproject.info/fdroid

I've added support to this app, and it'll be included in the next release:
* https://panicbutton.io/
* https://play.google.com/store/apps/details?id=org.iilab.pb

This pull request just implements the most basic panic response.  I could see OpenKeychain offering more options, like disguising itself, deleting all files (especially if the user has an off-device backup), etc.  That would require adding a "panic setup" Activity like [Zom](https://zom.im) has for the user to opt-in.  Read all about that here:
https://guardianproject.info/2016/01/12/panickit-making-your-whole-phone-respond-to-a-panic-button/

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -782,6 +782,9 @@
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
+        <activity
+            android:name=".ui.ExitActivity"
+            android:theme="@android:style/Theme.NoDisplay" />
 
         <!-- Internal services/content providers (not exported) -->
         <service
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -772,6 +772,17 @@
             android:name=".ui.HelpActivity"
             android:label="@string/title_help" />
 
+        <activity
+            android:name=".ui.PanicResponderActivity"
+            android:noHistory="true"
+            android:theme="@android:style/Theme.NoDisplay">
+            <intent-filter>
+                <action android:name="info.guardianproject.panic.action.TRIGGER" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
+        </activity>
+
         <!-- Internal services/content providers (not exported) -->
         <service
             android:name=".service.PassphraseCacheService"
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ExitActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ExitActivity.java
@@ -23,27 +23,30 @@ import android.content.Intent;
 import android.os.Build;
 import android.os.Bundle;
 
-import org.sufficientlysecure.keychain.service.PassphraseCacheService;
-
-public class PanicResponderActivity extends Activity {
-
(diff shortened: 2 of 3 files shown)


Nekroze:
+1 for this. As a panic responder maybe we could have openkeychain wipe all public keys including stubs.

The public keys are basically a contact list so might be worth while to wipe that.

Wiping the public keys and stubs would be great because that is all that openkeychain stores for security tokens like the Yubikey. And the Yubikey can just be reimported if the crisis is averted.

dschuermann:
@eighthave I think you are missing an import for PassphraseCacheService, travis is not building.

eighthave:
@dschuermann  yeah, sorry, `./gradlew` building was failing for me, so I just pushed to see what travis would say. I updated it.

@Nekroze maybe just wipe all app data just to be sure.  The PanicKit library provides a method for that: `PanicResponder.deleteAllAppData(Context context)`.  But maybe it makes sense to keep the preferences, and just wipe the public/private keys and the NFC config.

dschuermann:
Looks good to me :+1: 

If we implement some kind of app lock in the future (see https://github.com/open-keychain/open-keychain/issues/1541), we could then improve upon this.

@Nekroze If we want to implement a responder to wipe keys we would need Trusted Intents (lib by guardian project) as this is a very destructive method, which needs better protection. Currently, simply clearing the cache should be okay to expose without verification of the Intent sender.

@Valodim what's your opinion on this?

Valodim:
I'm a little confused by the choice of mechanism. Wouldn't a Broadcast Intent be better suited for this purpose?

eighthave:
It is possible to send/receive the `ACTION_TRIGGER Intent` as a Broadcast, but then the sender cannot be verified by the receiver. The only way that to tell which app sent an `Intent` is for:

1. the `Intent` to be sent with `startActivtyForResult()` from an `Activity` that has a `launchMode` of `standard` or `singleTop`.
2. the receiving `Activity` then can get the sender using `Activity.getCallingActivity()`.

I've been sketching out an alternate technique that will work within senders and receivers that are instances of `Activity`, `Service`, or `BroadcastReceiver`, but that will require that the `Intent` is crafted by the sender in a specific way using code from a library: https://dev.guardianproject.info/issues/6260

If you can find another way to verify `Intent` senders, I'd love to hear it.

eighthave:
Also, it is important to consider the panic trigger app in this, it might contain a private message that includes location or other sensitive information.  So trigger apps might want to verify which app it is sending the trigger to via a whitelist of APK signing keys or just `packageName`.  While it is possible to send a broadcast to a single app, it doesn't add anything over sending it with `startActivityForResult()` and just increases the risk that data might leak due to an implementation error.

dschuermann:
thanks! If you want to provide more features open a new pull request. In general I am open for the idea of using a Panic button :+1:

PR #1693, opened 2016-01-18
Always revoke subkeys with a revocation signature.

kbsriram:
Unlike UID revocations which are "reversible" by newer UID
self-signatures, a subkey revocation should be "permanent" even if
followed by a newer self-signature.

The RFC is ambiguous on this, but this is the convention used by (e.g.)
GnuPG. The rationale for this behaviour is available as comments within
the GnuPG source.

UID signatures:
https://github.com/mtigas/gnupg/blob/50c98c7ed6b542857ee2f902eca36cda37407737/g10/getkey.c#L1668-L1674
Subkey signatures:
https://github.com/mtigas/gnupg/blob/50c98c7ed6b542857ee2f902eca36cda37407737/g10/getkey.c#L1990-L1997

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 	path = extern/safeslinger-exchange
 	url = https://github.com/open-keychain/exchange-android
 	ignore = dirty
+[submodule "OpenKeychain/src/test/resources/openpgp-interop"]
+	path = OpenKeychain/src/test/resources/openpgp-interop
+	url = https://github.com/google/openpgp-interop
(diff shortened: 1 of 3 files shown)


dschuermann:
Looks good to me :+1 for merge.

Valodim:
Makes perfect sense. I implemented it like this because that's how uids are done, but the other choice is just as reasonable and de facto established.

I havn't had time yet to look at the interop testing addition, will try to get to that tomorrow :+1:

Valodim:
Looked it over, I have no grievances with this code. Mergin'

PR #1695, opened 2016-01-20
Decrypt operation wrongly returns a PendingIntent to show the signature key in the OpenPgpSignatureResult, even though there is no signature at all

mgeier63:
Please ignore the commits regarding the submodules, couldn't figure out how to exclude them. The only relevant change is in "OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java"


dschuermann:
Thanks for your work. Unfortunately, you changed the submodules and did not conform to our coding style, but as the change was small I did it here: https://github.com/open-keychain/open-keychain/commit/681a201a69141d61aeb09b89dfbe6fd429e68304

PR #1705, opened 2016-01-27
Better transaction animation

Franzaine:
This PR adds a transaction animation similar to what is found in the Fidesmo app to all NFC transaction views. The animation comes straight from the latest version of Nordpol.


Franzaine:
@dschuermann what do you think about this? :)

dschuermann:
In general: The design is great!

* On taking away the card too early a cross icon is displayed and the text is displayed below. This takes a lot of space, and means people have to scroll to find the "try again" button. I don't think we need an icon in the error case. As before, just showing the text would be great. Maybe a method could be added to the nordpol widget to display error text?
device-2016-02-01-155010
* I got a stacktrace, which seems unrelated: https://github.com/open-keychain/open-keychain/issues/1709
* Can we add a png displaying a hand that holds a yubikey besides the hand with the card?
* Could you add the original svgs to the nordpol git repo?

dschuermann:
* The last screen with the mark icon is shown for a long time. I know that I added a delay there so that people actually see this screen, but currently it's really long. Did the widget or nordpol add an additional delay?

Franzaine:
Nordpol PR with a USB NFC token: https://github.com/fidesmo/nordpol/pull/53

Franzaine:
@dschuermann what do you think about this?
Nordpol now has a hand holding a USB + NFC token
The SVG graphical resources are now in Nordpol
The error-step of NfcGuideView is no longer used
I looked around a bit and I don't think that the delay has changed after successful operation. Nordpol also has the TagArbiter that could remove the need for ever removing the token by the way. Just passing around the token in the app as needed. There'll soon be some more noise about this from us :)

dschuermann:
@Franzaine Can you please rebase on top of current master?

Franzaine:
@dschuermann done! :)

dschuermann:
Awesome work, thanks for this contribution. Looking much better now!

Franzaine:
@dschuermann Sure! My pleasure. The NfcGuideView isn't the brightest star yet but I think we together can iterate it to perfection :)

Valodim:
I'm not sure if this was mentioned elsewhere, but currently there is a very long delay (2+ seconds) after taking away the security token before the app proceeds. this should definitely be fixed before a release.

dschuermann:
I created a new issue here: https://github.com/open-keychain/open-keychain/issues/1734 to track this problem. 

@Valodim I also cleaned up the milestone: https://github.com/open-keychain/open-keychain/milestones/OpenKeychain%203.9
Do you have more blockers for a release? I would propose we track them in the milestone to have a better roadmap until releasing.

dschuermann:
@Franzaine the youtube demo for the TagArbiter looks awesome!

Issue #1701, opened 2016-01-25
crash in KeyserverSyncAdapterService


Valodim:
only happened a single time, so not really high prio

```
java.lang.RuntimeException: Unable to start service org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService@434523f0 with Intent { act=cancel flg=0x10 cmp=org.sufficientlysecure.keychain/.service.KeyserverSyncAdapterService }: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
	at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:2877)
	at android.app.ActivityThread.access$2200(ActivityThread.java:161)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1362)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:157)
	at android.app.ActivityThread.main(ActivityThread.java:5356)
	at java.lang.reflect.Method.invokeNative(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:515)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1265)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1081)
	at de.robv.android.xposed.XposedBridge.main(XposedBridge.java:132)
	at dalvik.system.NativeStart.main(Native Method)
Caused by: java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
	at org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService.isSyncEnabled(KeyserverSyncAdapterService.java:528)
	at org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService.onStartCommand(KeyserverSyncAdapterService.java:85)
```

PR #1706, opened 2016-01-27
Prevent Rare KeyserverSyncAdapterService Crash

adithyaphilip:
Fixes https://github.com/open-keychain/open-keychain/issues/1701. 

Caused by no account being created by the time `isSyncEnabled` was called. The intent was using `ACTION_CANCEL` (used in the `onResume` of `BaseActivity`), and since no Activity could have started before `onCreate` in `KeychainApplication` (<a href="http://developer.android.com/reference/android/app/Application.html#onCreate()">documentation</a>) was called, it looks like `onSyncCancelled` in `KeyserverSyncAdapter` was triggered somehow before `onCreate` completed. Is this possible? 

Either way, this PR should prevent a crash in case an account is not created as expected.


dschuermann:
Sounds like problems with specific Android distributions which that prevents us from creating accounts...

adithyaphilip:
Ah, that makes sense.

PR #1708, opened 2016-02-01
performance improvements (WIP)

Valodim:
I did some profiling yesterday, and was horrified to find that on my device, decrypting a small amount of data from k9 takes about 500ms:
* 100ms setup time from the binder framework
* **300ms spent on s2k and key extraction**
* 30 on the RSA operation
* 70 on the rest (aes decrypt, another shoter rsa for verification, sending)

I looked around a bit and found that we use the immensely expensive s2k operation all over the place, sometimes within 20 lines of code without caching. I spent some time caching the session keys in addition to the passphrases, and got the same operation down to 250ms total. I cache per "compatible s2k-configuration", i.e. encryption hash algo + method + iteration count + iv.

I also got rid of *all* uncached calls to getSecretKeyType, which were strewn all over the crypto operations. This is actually the change where I expect more stuff to break because of some underlying changes to ProviderHelper.

This is a WIP pull request, it does a lot of changes which are likely to break stuff so it requires extensive testing! This is still a large deal in my opinion, and we should try to get it into a release ASAP.


Valodim:
@kbsriram happy to report the first instance of a bug prevented by your tests in this PR :+1:

kbsriram:
Nice :smile: and happy to know they are helping.

dschuermann:
Can you add license headers and a little bit of documentation how these classes are used to SessionKeySecretKeyDecryptorBuilder and ComparableS2K?

Iambecomeroot:
I am still getting very poor performance on my LG G4. I moved the key from my moto g 2013 to my G4. Decryption takes about a second on the moto g and about a minute on the G4. I didn't want to create an issue because similar problems were talked about before. Any suggestions?

dschuermann:
@Iambecomeroot please create an issue. this is a closed PR.

Issue #1676, opened 2016-01-12
Switch from Spongy Castle to Bouncy Castle?


dschuermann:
According to the Book Android Security Internals, the bouncycastle packages have been renamed to "com.android.org.bouncycastle" starting with Android 3.0. If I understand this correctly, we could switch back to bouncy instead of spongy.

See also https://android.googlesource.com/platform/libcore/+/0ac85ead96f1ba7d35f3acadd154de4ef0a8fd87^!/

I will test this in a branch when I get time.

dschuermann:
@rtyley do you know something about this?

rtyley:
> @rtyley do you know something about this?

Not much beyond what you've already seen - it's a good change in Android, it's what they should have done in the first place! I think it's possible that _some_ post-3.0 devices by ill-behaved manufacturers may still bundle `org.bouncycastle.*` in their system class library, making it a little unpredictable as to whether a user will hit a problem or not - a Nexus device is probably fine, but for a non-vanilla Android there's no guarantees :crying_cat_face: I'd be interested to know the results of your tests.

For the time being at least, I'll continue to repackage Bouncy Castle as Spongy Castle, on the grounds that there are still plenty of pre-Android-3.0 devices out there that need decent cryptography.

dschuermann:
@rtyley Thanks for your answer. Will try to find some Samsung devices and test if they still include ``org.bouncycastle`` or if it's safe to migrate.

rtyley:
Cool, let me know how you go!

dschuermann:
@rtyley I did some tests and it looks ok, so we now switched to bouncycastle on git master.

I tested on the following devices with Amazon Device Farm:

Top 5 amazon devices:
Amazon Kindle Fire HDX 7 (WiFi)
4.4.3
LG G Pad 7.0 (AT&T)
4.4.2
Samsung Galaxy S5 (T-Mobile)
4.4.2
Samsung Galaxy S6 (Verizon)
5.0.2
Samsung Galaxy Tab 4 10.1 Nook (WiFi)
4.4.2

And I selected some older devices/Android versions:
ASUS Nexus 7 - 1st Gen (WiFi)
4.2.1
Samsung Galaxy Note (AT&T)
4.1.2
Samsung Galaxy S4 (AT&T)
4.4.4
Samsung Galaxy S4 (AT&T)
4.2.2

dschuermann:
[...]

PR #1715, opened 2016-02-08
Switch to Bouncy Castle

dschuermann:
As detailed in https://github.com/open-keychain/open-keychain/issues/1676 I would like to switch to Bouncy Castle.

This has the advantages of making pull requests to the original bouncycastle repo easier and it is one small part of a bigger step of making our OpenPGP code available as a library for desktop systems.

Currently only tested on Sony z3 Cyanogenmod 12.1 nightly. Will test some more devices tomorrow.


Valodim:
I'm a little worried that this will clash with shit vendors do. Considering samsung can't even keep collisions with the support library straight...

PR #1722, opened 2016-02-15
Better seperation of remote activities

dschuermann:
Awaiting feedback from @Valodim

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -844,6 +844,12 @@
             android:name=".remote.ui.AccountSettingsActivity"
             android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
             android:exported="false" />
+        <activity
+            android:name=".remote.ui.RemoteSecurityTokenOperationActivity"
+            android:allowTaskReparenting="true"
+            android:launchMode="singleTop"
+            android:taskAffinity=":Nfc"
+            android:theme="@style/Theme.Keychain.Light.Dialog" />
 
         <!-- DEPRECATED service,
              using this service may lead to truncated data being returned to the caller -->
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -845,11 +845,18 @@
             android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
             android:exported="false" />
         <activity
+            android:name=".remote.ui.RemotePassphraseDialogActivity"
+            android:theme="@style/Theme.Keychain.Transparent" />
+        <activity
             android:name=".remote.ui.RemoteSecurityTokenOperationActivity"
             android:allowTaskReparenting="true"
             android:launchMode="singleTop"
             android:taskAffinity=":Nfc"
             android:theme="@style/Theme.Keychain.Light.Dialog" />
+        <activity
+            android:name=".remote.ui.RemoteImportKeysActivity"
+            android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
+            android:label="@string/title_import_keys" />
 
         <!-- DEPRECATED service,
              using this service may lead to truncated data being returned to the caller -->
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -847,6 +847,8 @@
         <activity
             android:name=".remote.ui.RemotePassphraseDialogActivity"
             android:theme="@style/Theme.Keychain.Transparent" />
+        <!-- see SecurityTokenOperationActivity for reasons why
+             allowTaskReparenting, singleTop, and taskAffinity is used  -->
         <activity
             android:name=".remote.ui.RemoteSecurityTokenOperationActivity"
             android:allowTaskReparenting="true"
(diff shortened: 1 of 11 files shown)


OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -844,6 +844,19 @@
             android:name=".remote.ui.AccountSettingsActivity"
             android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
             android:exported="false" />
+        <activity
+            android:name=".remote.ui.RemotePassphraseDialogActivity"
+            android:theme="@style/Theme.Keychain.Transparent" />
+        <activity
+            android:name=".remote.ui.RemoteSecurityTokenOperationActivity"
+            android:allowTaskReparenting="true"
```
Valodim: a line of documentation for this one (and maybe for singleTop) would be nice

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPendingIntentFactory.java
```diff
@@ -0,0 +1,182 @@
...
+            case PASSPHRASE: {
+                return passphrase(requiredInput, cryptoInput);
+            }
+
+            default:
+                throw new AssertionError("Unhandled required input type!");
+        }
+    }
+
+    private PendingIntent nfc(RequiredInputParcel requiredInput, CryptoInputParcel cryptoInput) {
```
Valodim: three letter method name?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPendingIntentFactory.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    PendingIntent error(String errorMessage) {
+        Intent intent = new Intent(mContext, RemoteErrorActivity.class);
+        intent.putExtra(RemoteErrorActivity.EXTRA_ERROR_MESSAGE, errorMessage);
+
+        return build(intent);
+    }
+
+    private PendingIntent build(Intent intent) {
```
Valodim: "build" method name should only be used where we use the builder-pattern

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPendingIntentFactory.java
```diff
@@ -0,0 +1,182 @@
...
+import org.sufficientlysecure.keychain.service.input.CryptoInputParcel;
+import org.sufficientlysecure.keychain.service.input.RequiredInputParcel;
+import org.sufficientlysecure.keychain.ui.ViewKeyActivity;
+
+import java.util.ArrayList;
+
+public class ApiPendingIntentFactory {
+
+    Context mContext;
+    private Intent mPendingIntentData;
```
Valodim: why introduce this state here instead of a parameter?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPermissionHelper.java
```diff
@@ -187,20 +172,15 @@ protected AccountSettings getAccSettings(String accountName) {
     }
 
     /**
-     * Deprecated API
+     * @deprecated
```
Valodim: should be annotation of method, not part of documentation

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SecurityTokenOperationActivity.java
```diff
@@ -274,17 +268,7 @@ protected void doNfcInBackground() throws IOException {
 
     @Override
     protected void onNfcPostExecute() {
```
Valodim: make this one final maybe? there should never be a reason to override this

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SecurityTokenOperationActivity.java
```diff
@@ -315,6 +299,13 @@ protected void onPostExecute(Void result) {
         }.execute();
     }
 
+    protected void handleResult(CryptoInputParcel inputParcel) {
```
Valodim: document please

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPendingIntentFactory.java
dschuermann: Because ``mPendingIntentData`` stays the same, e.g., in ``OpenPgpService.decryptAndVerifyImpl()`` different methods of ApiPendingIntentFactory are executed always with the same ``mPendingIntentData``

Valodim:
please wait for a "review finished" comment before merging in the future.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPendingIntentFactory.java
Valodim: I have never seen a use of the factory pattern where the factory contained actual state which is made part of the objects it creates (i.e., the Context one is ok). Feels very confusing, I would advise against this.

dschuermann: done in https://github.com/open-keychain/open-keychain/commit/73ce6fc7bb9cd5a024bfe510e862dd9d6af3d2de

PR #1724, opened 2016-02-15
WIP: Backup API

dschuermann:
WIP: needs testing...

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/BackupOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/BackupOperation.java
@@ -20,6 +20,7 @@ package org.sufficientlysecure.keychain.operations;
 
 
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.DataOutputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -97,6 +98,12 @@ public class BackupOperation extends BaseOperation<BackupKeyringParcel> {
 
     @NonNull
     public ExportResult execute(@NonNull BackupKeyringParcel backupInput, @Nullable CryptoInputParcel cryptoInput) {
+        return execute(backupInput, cryptoInput, null);
+    }
+
+    @NonNull
+    public ExportResult execute(@NonNull BackupKeyringParcel backupInput, @Nullable CryptoInputParcel cryptoInput,
+                                OutputStream outputStream) {
 
         OperationLog log = new OperationLog();
         if (backupInput.mMasterKeyIds != null) {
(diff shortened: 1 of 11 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/BackupCodeFragment.java
```diff
@@ -436,10 +457,22 @@ private void startBackup() {
...
                 : Constants.FILE_EXTENSION_ENCRYPTED_BACKUP_PUBLIC);
 
+        Passphrase passphrase = new Passphrase(mBackupCode);
+        if (Constants.DEBUG && mDebugModeAcceptAnyCode) {
+            passphrase = new Passphrase("AAAA-AAAA-AAAA-AAAA-AAAA-AAAA");
+        }
+
+        // special case for remote API, see RemoteBackupActivity
+        if (mReturnBackup) {
+            ((RemoteBackupActivity) getActivity()).handleResult(new CryptoInputParcel(passphrase));
```
dschuermann: @Valodim Do you have an idea how to handle this separation better?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupKeyringParcel.java
```diff
@@ -28,15 +28,12 @@
 
 public class BackupKeyringParcel implements Parcelable {
     public Uri mCanonicalizedPublicKeyringUri;
-    public Passphrase mSymmetricPassphrase;
```
dschuermann: @Valodim I will now use CryptoInputParcel for the passphrase. Is this okay or are there reasons to leave this inside the operation parcel?

dschuermann: For explanation: the CryptoInputParcels are properly cached when used with the API without further work.

dschuermann:
TODOs for me:
* I wasn't able to finish the backup process when entering the backup code, looks like a problem with the new edittext
* I wasn't able to properly get back to the last character, the special whitecases prevent this
* I need to implement EXTRA_REQUEST_ASCII_ARMOR (currently always true)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/BackupOperation.java
```diff
@@ -97,6 +98,12 @@ public BackupOperation(Context context, ProviderHelper providerHelper,
 
     @NonNull
     public ExportResult execute(@NonNull BackupKeyringParcel backupInput, @Nullable CryptoInputParcel cryptoInput) {
+        return execute(backupInput, cryptoInput, null);
```
Valodim: doesn't this just crash?

PR #1727, opened 2016-02-22
try to guess if data is text or not

### LINKED ISSUE #1687: Unknown filename
When trying to open any encrypted data from either clipboard or a file the app gives me an Uknown filename output. When I'm trying to open it it gives me 'show with Nova launcher or decrypt with openkeychain' the latter results in getting to the same screen I came from. 
I'm using Gpg4win to encrypt stuff

Valodim:
We very often get data which is marked as binary even though it's really text. This happens from gpg4win, and also when people use `gpg --encrypt <<< blah` without using the --text option.

I added a mechanism to attempt an educated guess on whether the data we receive is binary, by trying to decode the data as the specified charset, so basically we now show the "View in OpenKeychain" intent whenever there is no reason *not* to believe that the data we receive is text, e.g. when we have a known mimeType which cannot be text, or the data contains characters which are invalid for the specified charset.

This should hopefully fix #1687

Feel free to test & review, will merge in a couple of days otherwise.


Valodim:
Oh almost forgot: This introduces an isLikelyText flag to the OpenPgpMetadata class, indicates this status (i.e. it's false if we have a reason to believe it's binary). There is more information we could expose, but this is the one which was important for DecryptListFragment so it's the only one I included for now.

dschuermann:
Is there a sufficiently high probability that something marked with ``isLooksLikeText`` is in fact not text but binary? If the probability is negligible, do we really need to introduce ``isLooksLikeText``? Couldn't we simply set the mime type to ``text/plain``? Even in the case that it's binary there is still the option to save the file or "open with" from long click.

Valodim:
I have no good numbers on that, but depending on the charset the probability that a binary file is recognized as text seems quite realistic. I also really dislike setting metadata based on guesswork without at least setting a flag about it, dealing with false positives if other apps do this kind of thing is a huge pain (stuff like #1223 comes to mind).

"application/octet-stream, but all printable in this charset" seems like the right information to give if that is all we have.

Valodim:
Did a fact check: It's actually very improbable that any non-tiny number of random bytes would come out as valid utf-8. Also, considering that the mimetype we set is based entirely on guesswork to begin with, setting it to text/plain might be okay for the API.

*However*, I would still like to provide that info as output of our internal operations, for which we also use OpenPgpMetadata. The logic which decides what we do with a file by type (icon, filename text, immediate display as text, …) is in DecryptListFragment, and it behaves differently in regards to looksLikeText: A text file which is known to be text is immediately displayed as such, a text file which is only guessed isn't.

I'm not entirely sure if that behavior is really what we want there. I wonder what's more consistent for the user...

dschuermann:
To be honest, I think the user will not care if it has been guessed as text or if it has been properly encrypted as text with ``--text`` option. I also think we should display it immediately in both cases.

The mime type is almost always guesswork. Only the OpenPGP packets with "u"/"t" are properly mapped to "text/plain", all other are mapped based on file endings, but yeah you know that :grimacing: . Back then, the idea was to provide developer usability by providing a mime type instead of this OpenPGP mess, because Android operates on mime types instead of file endings or magic bytes.
Not sure if we need more information internally than what others need in regards to file types... I would argue that the guessed mime type is enough :stuck_out_tongue_closed_eyes:

Valodim:
Hm. Yeah I guess.

I changed the behavior now: If we don't know the type or it's application/octet-stream and there are no errors decoding it as utf-8, we change it to `text/plain`, which also means we it is displayed as text immediately. If the type is `text/*` we don't show the data immediately, but we do add "View with OpenKeychain" intent to the chooser. For other scenarios like "charset was specified but is faulty", we have that information but don't expose it at the moment, while all other behavior is unchanged.

dschuermann:
yep looks and sounds good!

PR #1737, opened 2016-03-01
Fix backup code comparison

runnerway:
Backup functionality was broken because of a wrong comparison of the backup code. 
Save the string as it's used more than one time.


dschuermann:
Great thanks

Valodim:
The previous comparison was broken? I fail to see a semantic change in your commit, what exacty was broken here?

runnerway:
Here: https://github.com/open-keychain/open-keychain/blob/77150418cdf530c37440915a56b59685fc24d5b8/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/BackupCodeFragment.java#L374

"getText()" missing.

Valodim:
Ah, you're right! Nice catch.

Issue #1733, opened 2016-03-01
import private key crashes due to lack of permission


spaetz:
This is cyanogenmod 13 (android 6) on my moto g 2014. Trying to import my private key, I open the file manager, select my key file and BOOM OpenKeychain crashes. Stacktrace below indicates a lack of file permission, and the privacy settings do not show an file acces permission indeed.
This is OpenKeychain 3.8.2 from fdroid, in case this matters.

```
E/Keychain(27421): FileNotFoundException!
E/Keychain(27421): java.io.FileNotFoundException: /storage/emulated/0/spaetz/gpg_2016_sspaeth_65B22153_sec.asc: open failed: EACCES (Permission denied)
E/Keychain(27421): 	at libcore.io.IoBridge.open(IoBridge.java:452)
E/Keychain(27421): 	at java.io.FileInputStream.<init>(FileInputStream.java:76)
E/Keychain(27421): 	at java.io.FileInputStream.<init>(FileInputStream.java:103)
E/Keychain(27421): 	at android.content.ContentResolver.openInputStream(ContentResolver.java:662)
E/Keychain(27421): 	at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListLoader.getInputData(ImportKeysListLoader.java:143)
E/Keychain(27421): 	at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListLoader.loadInBackground(ImportKeysListLoader.java:76)
E/Keychain(27421): 	at org.sufficientlysecure.keychain.ui.adapter.ImportKeysListLoader.loadInBackground(ImportKeysListLoader.java:45)
E/Keychain(27421): 	at android.support.v4.content.AsyncTaskLoader.onLoadInBackground(AsyncTaskLoader.java:296)
E/Keychain(27421): 	at android.support.v4.content.AsyncTaskLoader$LoadTask.doInBackground(AsyncTaskLoader.java:54)
E/Keychain(27421): 	at android.support.v4.content.AsyncTaskLoader$LoadTask.doInBackground(AsyncTaskLoader.java:42)
E/Keychain(27421): 	at android.support.v4.content.ModernAsyncTask$2.call(ModernAsyncTask.java:128)
E/Keychain(27421): 	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
E/Keychain(27421): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
[...]

PR #1738, opened 2016-03-02
Always check read permission

runnerway:
This solve a wrong "File not found" when the user try to import a key when the app was not granted read permission previously. (Using Android 6.0)
Allow the user to grant read permission even in this case.

This is a quick fix.
Probably the code here need a bit of refactoring. 
If you agree i will try to restructure it a bit in the future.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysListFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysListFragment.java
@@ -330,8 +330,15 @@ public class ImportKeysListFragment extends ListFragment implements
     }
 
     public void loadNew(LoaderState loaderState) {
-
         mLoaderState = loaderState;
+
+        if (mLoaderState instanceof BytesLoaderState) {
+            BytesLoaderState ls = (BytesLoaderState) mLoaderState;
+
+            if ( ! checkAndRequestReadPermission(ls.mDataUri)) {
+                return;
+            }
+        }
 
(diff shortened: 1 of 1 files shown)


runnerway:
Probably this fix https://github.com/open-keychain/open-keychain/issues/1733 too.

runnerway:
Added a refactoring of permissions requesting to remove a lot of duplicate code.
This will ensure coherence throughout the app.

No functional changes apart from a notification in case of Camera permission denied as with Storage permission.

I would like to know your opinion.

dschuermann:
I am not totally sold to moving the methods into a helper class :P
Besides, I would propose we don't always show a toast if a permission has been denied. Previously, when the qr code scanning activity has been started and the camera permission was denied, we would simply finish the Activity. It is pretty clear to the user why, I don't think we need to bother showing a Toast here.

@Valodim what do you think?

runnerway:
Maybe adding a "boolean showToast" to "checkRequestedPermission" can be a solution?

Do you have better ideas to avoid this amount of code duplication?
I'm thinking when more permissions will be needed in more places than now and things probably will become messy if we don't keep things in one place.

Or you prefer to leave things as they are now?

dschuermann:
IMHO, there is not much code duplication currently. Moving the methods in its own class is an indirection making it more difficult to directly see the program flow for new devs.

Can you make one pull request to only fix the bug and one where you implement the helper class?

runnerway:
For sure. Removed the helper class. Probably this PR can be merged.
For the helper class, i will think a bit about it.

Issue #1720, opened 2016-02-14
Contact synchronization notification


ghost:
On Android 6.0 Openkeychain keeps asking me to allow contact synchronization. It seems it'll keep bothering me until I allow permission to my contacts. It should only appear once if at all (if contact synchronization is disabled there's no point to ask for contact access)
To be clear, it's not an Android permission dialogue, it's notification from the app itself that links to the settings page for synchronization.

Besides, I don't see the point in syncing openkeychain to my contacts since it will only create duplicated contacts.

clawoflight:
It will not create duplicate contacts. It will link the openkeychain contact with any existing system contact, like you may be used to from whatsapp or similar.

Valodim:
Yeah that way you have a link from contact details to their keys.

Either way, it is certainly a bug that this dialog keeps coming up. If the user said no once (or maybe twice), it should not come up again.

akoul:
Is it possible to test this bug on emulator? I do not have android 6.0

dschuermann:
@akoul sure, why shouldn't it be reproduced in an emulator?

advaita13:
I have a fix in mind. Before that I'm trying to reproduce the bug so I can later understand if the fix is working as expected. Should be fairly straightforward.

tdjogi010:
@dschuermann I am able to reproduce the bug. The notification comes from the ContactSyncAdapterService class, do you want to stop the service after the user has ignored it several times (number of times ignored can be stored in sharedpreferences) OR let the service be running and just don't publish the notification? I think the former one suits the best.

advaita13:
@dschuermann I too was considering using SharedPreferences to store how many times the notification has been generated. Let's see if that works.

dschuermann:
This notification should not come up because the automatic sync is turned of when before the notification is shown, so it should only come up once. BUT: I think requestContactsSync() ignores if automatic sync is turned on or off, maybe this method needs to check before syncing if automtic sync is turned on. But this is just a guess. This needs to be tested thoughtfully.

advaita13:
True. The issues seems to stem from the requestContactsSync() method. I've implemented the fix using SharedPreferences and it seems to work. I tracked the usages of the ContactSyncAdapterService class and used this information to test the app against every situation that might cause the notification to resurface. It hasn't reappeared so far :)

dschuermann:
[...]

PR #1743, opened 2016-03-02
Contact Synchronisation Notification

advaita13:
Issue: https://github.com/open-keychain/open-keychain/issues/1720

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/ContactSyncAdapterService.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/ContactSyncAdapterService.java
@@ -143,6 +143,13 @@ public class ContactSyncAdapterService extends Service {
     }
 
     public static void requestContactsSync() {
+        // if user has disabled automatic sync, do nothing
+        if (!ContentResolver.getSyncAutomatically(
+                new Account(Constants.ACCOUNT_NAME, Constants.ACCOUNT_TYPE),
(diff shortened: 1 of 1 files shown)


dschuermann:
Great work, thanks!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/ContactSyncAdapterService.java
```diff
@@ -143,6 +143,13 @@ public IBinder onBind(Intent intent) {
     }
 
     public static void requestContactsSync() {
+        // if user has disabled automatic sync, do nothing
```
Valodim: for the record, a nicer way to do this is:
```
       boolean isSyncEnabled = ContentResolver.getSyncAutomatically(
                 new Account(Constants.ACCOUNT_NAME, Constants.ACCOUNT_TYPE), ContactsContract.AUTHORITY); 
       if (!isSyncEnabled) {
             return;
         }
```
it's the same number of lines, but instead of writing a comment you give the condition "isSyncEnabled" an explicit name. that way, someone who reads this can grasp everything that's happening by reading only java, without the need to parse english somewhere inbetween

advaita13: Good point, I'll keep this in mind.. I'll make this change in my next pull request.

Issue #1570, opened 2015-10-26
Remove org.apache.http dependencies in LinkedTokenResource etc.


dschuermann:
org.apache.http has been deprecated in SDK 23. We currently use a legacy dependency for this support, see https://github.com/open-keychain/open-keychain/commit/67a51cdfe3f6fb7982b27eb2b6d4b1b5d49f69aa

Valodim:
what's the recommended alternative?

dschuermann:
UrlConnection

Valodim:
ah, yes. I already did that in the relatively new code in `LinkedIdCreateGithubFragment`, the code where I used apache stuff is older. will update at some point.

dschuermann:
You could also use OkHttp btw.

dschuermann:
done in https://github.com/open-keychain/open-keychain/pull/1748

PR #1744, opened 2016-03-02
Replaced org.apache.http dependencies in LinkedTokenResource etc. #1570

kepuss:
I replaced the org.apache.http to UrlConnection. 
Checked manually on device on github linked identity
(Sorry for any mistakes, it's my first request),


dschuermann:
Because we now use OkHttp which also supports Proxies and Tor, I would propose to replace UrlConnection with OkHttp APIs. @kepuss can you do that?

kepuss:
Sure

PR #1747, opened 2016-03-04
Fix duplicate file insertion while encrypting

runnerway:
Uri should be removed from pending even when it was already added.
Otherwise we are stuck and can not add new files anymore.


dschuermann:
Thanks, nice catch! :+1:

Issue #1659, opened 2016-01-04
Notification to remind user of backup


dschuermann:
I thought of integrating a backup screen into the create key wizard to force the user of creating a backup of the newly generated private key. This way users can revoke their keys later if it got compromised. We could also generate a revocation certificate, but I think this is less usable for the user what this is exactly and how it can be used later.

On a second thought, it could make the wizard too long and annoying, so what about reminding the user via a notification after lets say 14 days to create a backup of their newly created key?

Valodim:
wizard would indeed get too long and annoying, especially if we enforce this.

Tcheutchoua-Steve:
I think a reminandal notification will be better. Should we go for implementing a notification after 14 days  @dschuermann   @Valodim ?

Valodim:
five days or so sounds better, but changing that at a later point is not a problem. this should obviously be tied to individual created keys, and removed when the user creates a backup themself

Valodim:
other than that, this should be straightforward. anyone wanna give it a go? :)

Tcheutchoua-Steve:
I want to give it a go @Valodim. So can I proceed ?

Valodim:
Sure! There's not really a need to ask either, although it's of course a good idea to mention you are working on it so nobody else works on the same thing simultaneously.

Feel free to make a work-in-progress pull request once you have something, that makes it easy for us to give feedback.

rohands:
@Tcheutchoua-Steve 
Are you working on this issue?

Valodim:
This was more than a month ago, I would assume he isn't (unless noted otherwise)

007durgesh219:
Hi Developers, I am a new contributor  and want to work on this issue. Kindly let me know if anyone else is working on it. Thanks.

007durgesh219:
Please let me know what should be target branch of this fix??

dschuermann:
@007durgesh219 nobody is working on it. Target is always master.

007durgesh219:
@dschuermann I was thinking of using shared preferences for it.  Whenever user creates a new key or edit than we will store a flag in the shared preferences that user haven't created backup for this particular key and set AlarmManager to remind after 5 days and notify user.  In the meanwhile if user manually creates a backup we will ignore reminder of alarm manager.  Am I going in the right direction if not kindly suggest any other approach you have in mind

PR #1749, opened 2016-03-05
Backup reminder notification implemented (#1659)

007durgesh219:
Currently, it will trigger a notification after 30 seconds (For testing).

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/BackupNotificationAlarmReceiver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/BackupNotificationAlarmReceiver.java
@@ -1,0 +1,99 @@
+package org.sufficientlysecure.keychain.receiver;
+
+import android.app.AlarmManager;
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.app.NotificationCompat;
+import android.util.Log;
+
+import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.R;
+import org.sufficientlysecure.keychain.provider.KeychainContract;
+import org.sufficientlysecure.keychain.ui.ViewKeyActivity;
+import org.sufficientlysecure.keychain.util.Preferences;
+
+
+public class BackupNotificationAlarmReceiver extends BroadcastReceiver {
+    public static final String MASTER_KEY_ID = "masterKeyId";
+    public static final String NAME = "name";
+    public static final String EMAIL = "email";
+
+    public static final String PERFORM_BACKUP = "performBackup";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.i(Constants.TAG, "Inside backup notification alarm receiver");
+
+        NotificationManager notificationManager = (NotificationManager)
+                context.getSystemService(Context.NOTIFICATION_SERVICE);
+
+        long masterKeyId = intent.getLongExtra(MASTER_KEY_ID, 0);
+        String name = intent.getStringExtra(NAME);
+        String email = intent.getStringExtra(EMAIL);
+
+        // check if backup is already created
+        String masterKeyIds = Preferences.getPreferences(context).getBackupReminders();
+        if (masterKeyIds == null || !masterKeyIds.contains(Long.toString(masterKeyId)) || masterKeyId == 0) {
+            // cancel the repeating alarm by same PendingIntent
+            Intent intentAlarm = new Intent(context, BackupNotificationAlarmReceiver.class);
+
+            intentAlarm.putExtra(MASTER_KEY_ID, masterKeyId);
+            intentAlarm.putExtra(NAME, name);
+            intentAlarm.putExtra(EMAIL, email);
+
+            AlarmManager alarmManager = (AlarmManager)context.getSystemService(Context.ALARM_SERVICE);
+            PendingIntent pendingIntent = PendingIntent.getBroadcast(context, 1, intentAlarm,
+                                                            PendingIntent.FLAG_UPDATE_CURRENT);
+            alarmManager.cancel(pendingIntent);
+
+            return;
+        }
+
+        notificationManager.notify(Long.valueOf(masterKeyId).hashCode(), getNotification(context,
+                masterKeyId, name, email));
+    }
+
+    private Notification getNotification(Context context, long masterKeyId, String name, String email) {
+        NotificationCompat.Builder builder = new NotificationCompat.Builder(context);
+        builder.setSmallIcon(R.drawable.ic_stat_notify_24dp)
+                .setColor(context.getResources().getColor(R.color.primary))
+                .setContentTitle(context.getString(R.string.backup_reminder_notif_title))
+                .setContentText(context.getString(R.string.backup_reminder_notif_touch_to_backup));
+
+        NotificationCompat.InboxStyle inboxStyle = new NotificationCompat.InboxStyle();
+
+        inboxStyle.setBigContentTitle(context.getString(R.string.backup_reminder_notif_title));
+
+        inboxStyle.addLine(name+ " <" +email+ ">");
+
+        // Moves the big view style object into the notification object.
+        builder.setStyle(inboxStyle);
+
+        Intent resultIntent = new Intent(context, ViewKeyActivity.class);
+        resultIntent.setData(KeychainContract.KeyRings.buildGenericKeyRingUri(masterKeyId));
+        resultIntent.setAction(PERFORM_BACKUP);
+
+        PendingIntent createBackupPi = PendingIntent.getActivity(
+                context,
+                0,
+                resultIntent,
+                PendingIntent.FLAG_UPDATE_CURRENT
+        );
+
+        // Add backup PI to normal touch
+        builder.setContentIntent(createBackupPi);
+
+        // Add backup PI action below text
+        builder.addAction(
+                R.drawable.ic_view_list_grey_24dp,
+                context.getString(R.string.backup_reminder_notif_backup_now),
+                createBackupPi
+        );
+
+        return builder.build();
+    }
+}
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/BackupCodeFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/BackupCodeFragment.java
@@ -53,7 +53,6 @@ import android.view.View.OnClickListener;
 import android.view.ViewGroup;
 import android.view.animation.AccelerateInterpolator;
 import android.view.inputmethod.EditorInfo;
-import android.widget.EditText;
 import android.widget.TextView;
 
 import com.github.pinball83.maskededittext.MaskedEditText;
@@ -63,6 +62,7 @@ import org.sufficientlysecure.keychain.R;
 import org.sufficientlysecure.keychain.operations.results.ExportResult;
 import org.sufficientlysecure.keychain.provider.TemporaryFileProvider;
 import org.sufficientlysecure.keychain.service.BackupKeyringParcel;
+import org.sufficientlysecure.keychain.util.BackupNotificationUtil;
 import org.sufficientlysecure.keychain.ui.base.CryptoOperationFragment;
 import org.sufficientlysecure.keychain.ui.util.Notify;
 import org.sufficientlysecure.keychain.ui.util.Notify.ActionListener;
@@ -211,6 +211,12 @@ public class BackupCodeFragment extends CryptoOperationFragment<BackupKeyringPar
                 }
 
                 popBackStackNoAction();
+
+                if (mMasterKeyIds == null) {
+                    BackupNotificationUtil.removeAllBackupReminders(getContext());
+                } else {
+                    BackupNotificationUtil.removeBackupReminder(getContext(), mMasterKeyIds[0]);
+                }
 
                 break;
             }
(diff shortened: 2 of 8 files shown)


007durgesh219:
@dschuermann Any updates on this?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
```diff
@@ -362,6 +365,18 @@ public boolean onCryptoSetProgress(String msg, int progress, int max) {
         mCreateOpHelper.cryptoOperation();
     }
 
+    private void setBackupReminderNotification(EditKeyResult result) {
```
Valodim: I'd suggest moving this method to BackupNotificationService, passing a context and masterKeyId.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/BackupCodeFragment.java
```diff
@@ -425,6 +428,21 @@ public void onBackStackChanged() {
         }
     }
 
+    private void removeBackupReminders() {
```
Valodim: move to BackupNotificationService (like below)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupNotificationService.java
```diff
@@ -0,0 +1,88 @@
...
+import org.sufficientlysecure.keychain.receiver.BackupNotificationAlarmReceiver;
+import org.sufficientlysecure.keychain.util.Preferences;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * This class is used to setup a backup reminder for every new key created by user
+ */
+public class BackupNotificationService extends Service {
```
Valodim: Maybe I'm missing something here, but what is the point of having these methods in a service? Could they not simply be run directly?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupNotificationService.java
```diff
@@ -0,0 +1,88 @@
...
+
+        intentAlarm.putExtra(BackupNotificationAlarmReceiver.MASTER_KEY_ID, mMasterKeyId);
+        intentAlarm.putExtra(BackupNotificationAlarmReceiver.NAME, name);
+        intentAlarm.putExtra(BackupNotificationAlarmReceiver.EMAIL, email);
+
+        PendingIntent pendingIntent = PendingIntent.getBroadcast(this,1,intentAlarm,
+                PendingIntent.FLAG_UPDATE_CURRENT);
+
+        AlarmManager alarmManager = (AlarmManager)this.getSystemService(Context.ALARM_SERVICE);
+        alarmManager.setInexactRepeating(AlarmManager.RTC_WAKEUP, System.currentTimeMillis(),
```
Valodim: This is not really a repeating event, is there a reason not to use `set` for this and simply have it trigger once?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/BackupNotificationAlarmReceiver.java
```diff
@@ -0,0 +1,97 @@
...
+
+        NotificationManager notificationManager = (NotificationManager)
+                context.getSystemService(Context.NOTIFICATION_SERVICE);
+
+        long masterKeyId = intent.getLongExtra(MASTER_KEY_ID,0);
+        String name = intent.getStringExtra(NAME);
+        String email = intent.getStringExtra(EMAIL);
+
+        // check if backup is already created
+        String masterKeyIds = Preferences.getPreferences(context).getBackupReminders();
```
Valodim: storing this in preferences *and* the intents seems redundant. why not simply use the data from the intent? can you elaborate on this decision?

007durgesh219: As we are showing a reminder notification after 5 days. But what if the user has already created a backup (let say on 2nd day) then it will show notification unnecessarily after 5 days when the alarm will trigger. To handle this situation I am using preferences.  It will save masterKeyIds in preferences as a key is created (In CreateKeyFinalFragment) and remove them from preferences as the backup is created (in BackupCodeFragment) .  In this way at any point of the time masterKeyIds of all those keys whose backup is not created will be there in preferences. And I will check when the alarm is fired whether masterKeyId of the corresponding key is in the preference or not. If not that means backup is created in the meanwhile so, notification will not be shown.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupNotificationService.java
007durgesh219: I thought we will show notification after every 5 days till user does not create the backup. That's why it's repeating. But I see your point if he doesn't create the backup in the first time then we should not show notification again. He might not interested in creating backup at all. I will change it as you suggested. Thanks.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
007durgesh219: OK, I will do that. Thanks.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupNotificationService.java
Valodim: No this might be fine, actually.

007durgesh219: Sure. But this might irritate the user. As we will show notification again and again.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
Valodim: see the note on BackupNotificationService below first

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/BackupNotificationService.java
007durgesh219: Yup, you are right. Initially, I thought it will be better to not run this on the UI thread (using AsyncTask or service) and created a service.  Instead, I should simply rename this class to BackupNotificationUtil and put in util.

Valodim: I'm not so sure about that, firing off an alarmmanager intent seems like a harmless thing for the ui thread, my intuition would be to keep it on there for simplicity. Otherwise, an AsyncTask is a good lightweight option. If you want to go the extra mile here you can of course profile it to get the right answer™.

007durgesh219: Sure you are right.  I might have overthought on this. Will change it to simply util class.

007durgesh219:
I've updated PR Kindly review it again.

007durgesh219:
Any review on this ??

Valodim:
oh, this one slipped through the cracks in my list. will review later today.

007durgesh219:
Rebase to remove conflicts with master

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -444,6 +446,40 @@ public boolean getExperimentalEnableKeybase() {
...
+        SharedPreferences.Editor editor = mSharedPreferences.edit();
+        editor.putString(Pref.BACKUP_REMINDERS, reminders);
+        editor.apply();
+    }
+
+    public void removeBackupReminder(long masterKeyId) {
+        String reminders = mSharedPreferences.getString(Pref.BACKUP_REMINDERS,null);
+        if (reminders != null) {
+            List<String> masterKeyIds = new ArrayList<>(Arrays.asList(reminders.split(",")));
+            masterKeyIds.remove(masterKeyId+"");
```
Valodim: please use Long.toString, this isn't javascript ;)

Valodim: this functionality is very specific and should be in a (static) method

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -444,6 +446,40 @@ public boolean getExperimentalEnableKeybase() {
...
+            reminders = TextUtils.join(",", masterKeyIds);
+            SharedPreferences.Editor editor = mSharedPreferences.edit();
+            if (reminders == "")
```
Valodim: string comparison with ==

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -444,6 +446,40 @@ public boolean getExperimentalEnableKeybase() {
...
+                editor.remove(Pref.BACKUP_REMINDERS);
+            else
+                editor.putString(Pref.BACKUP_REMINDERS, reminders);
+            editor.apply();
+        }
+    }
+
+    public String getBackupReminders() {
+        return mSharedPreferences.getString(Pref.BACKUP_REMINDERS,null);
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/BackupNotificationAlarmReceiver.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class BackupNotificationAlarmReceiver extends BroadcastReceiver {
+    public static final String MASTER_KEY_ID = "masterKeyId";
+    public static final String NAME = "name";
+    public static final String EMAIL = "email";
+
+    public static final String PERFORM_BACKUP = "performBackup";
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.i(BackupNotificationAlarmReceiver.class.getName(),"Inside alarm receiver");
```
Valodim: `Constants.LOG_TAG`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
```diff
@@ -328,6 +329,9 @@ public SaveKeyringParcel createOperationInput() {
             @Override
             public void onCryptoOperationSuccess(EditKeyResult result) {
 
+                BackupNotificationUtil.setBackupReminder(getContext(),result.mMasterKeyId,
```
Valodim: `result.mMasterKeyId` might be `null` at this point. this is probably better kept in `finishWithResult`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/BackupNotificationUtil.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * This class is used to setup a backup reminder for every new key created by user
+ */
+public class BackupNotificationUtil {
+
+    public static void setBackupReminder(Context context, long mMasterKeyId, String name, String email) {
```
Valodim: misleading parameter name mMasterKeyId

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/BackupNotificationUtil.java
```diff
@@ -0,0 +1,68 @@
...
+        if (context == null) {
+            return;
+        }
+
+        Preferences preferences = Preferences.getPreferences(context);
+        List<String> masterKeyIds = Arrays.asList(preferences.getBackupReminders().split(","));
+        NotificationManager notificationManager = (NotificationManager)
+                context.getSystemService(Context.NOTIFICATION_SERVICE);
+        for (String masterKeyId : masterKeyIds) {
+            notificationManager.cancel(((Long) (Long.parseLong(masterKeyId))).hashCode());
```
Valodim: is this cast to Long needed?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/BackupNotificationUtil.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * This class is used to setup a backup reminder for every new key created by user
+ */
+public class BackupNotificationUtil {
+
+    public static void setBackupReminder(Context context, long mMasterKeyId, String name, String email) {
+        if (context == null) {
```
Valodim: context should be annotated `@NonNull` and the check performed outside

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
007durgesh219: I need mSharedPreferences in this function which is non-static. So, I cannot make this static method. Is that what you mean or am I missing something??

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
007durgesh219: ```finishWithResult``` take ```OperationResult``` as an input argument. But I need masterKey. How do I get that??

Valodim: You can also do it before `finishWithResult`, the important thing is that it's after the `null` check.

007durgesh219:
It's ready for another review. I've updated PR.

dschuermann:
This pull request is currently on hold. We are thinking about improvements to the UX. Currently the user is sent to the Backup screen directly. This could be confusing, but we haven't decided right now what the best user journey would be.

vishwesh3:
Backup reminder notification implemented or not?

dschuermann:
Read my comment

Valodim:
closing stale PR

PR #1751, opened 2016-03-05
Fix wrong behavior of third-party apps while opening files after decryption

runnerway:
Some behave strangely when opening files after decryption because they have no name. 
(Ex: Adobe Reader crashes)

Fix it by setting the correct name after decryption.


dschuermann:
Great, thanks! much appreciated!

Issue #1670, opened 2016-01-11
sync only when unmetered internet is available?


Valodim:
This can probably be answered by @adithyaphilip: what are our current considerations for internet availability for key sync? If a user has a lot of keys, since we have to download the entire key to find if new content is available, it's quite possible that we sync relevant amounts of data (i.e. 1MB+). Do we currently restrict the key sync operation to a time when unmetered internet is available? Or would it also be run on mobile data?

I might have missed a conversation on this topic, but @dschuermann didn't know either. Has this been considered?

adithyaphilip:
Nope, we hadn't considered that at the time. Right now we're only run if the user has an active internetconnection, but that's pretty much it. So unless the user has restricted background data we run on mobile data too.

We could check if the user is connected to an unmetered WiFi network, failing which we could postpone the sync by say, three hours? There is however the issue of if both mobile and WiFi are connected, we may not be able to predict which connection Android decides to use. What should we do in that case?

Valodim:
Hmmm, yeah that seems like the way to go. Checking wifi state requires a new permission though :(

dschuermann:
This is really difficult to get right. There is an article on http://developer.android.com/training/basics/network-ops/managing.html#manage-usage that implements settings similar to those available in Google Play's settings screen for auto-updating apps. However, even if we would implement something like this, what should we do if a user has "only on wifi" and then never connects to one? The keys will get old...

@Valodim ACCESS_NETWORK_STATE is automatically granted on Android 6 and I think it is also one of the "other" permissions no longer shown in Google Play.

Valodim:
You're right, it's a [normal](http://developer.android.com/reference/android/Manifest.permission.html#ACCESS_NETWORK_STATE) permission. "Only on Wifi" but user is never on wifi seems like an edge case, I wouldn't consider it for now.

advaita13:
Ignoring the edge case it should be straightforward to implement a "Sync Only Over Wifi" preference. If option is checked and there is no wifi connection sync can be postponed by a given time interval.

Much of this is already implemented in the KeyserverSyncAdapterService class for Automatic Key Updates. I can give it a try if this solution is acceptable.

adithyaphilip:
[...]

PR #1755, opened 2016-03-05
sync only when unmetered internet is available? #1670

advaita13:
Created a Sync Only on Wi-Fi switch in Settings. When On, a receiver is
created when the sync is pending and Wi-Fi is disconnected. This
listens for any Wi-Fi state change broadcasts. Once connected the
receiver is disabled.

Reference : #1670

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -5,11 +5,6 @@
     android:installLocation="auto">
 
     <!--
-        General remarks
-        ===============
-        - Last APG 1 version was 10900 (1.0.9 beta 00)
-        - Keychain starting with versionCode 20000!
-
         Association of file types to Keychain
         =====================================
         General remarks about file ending conventions:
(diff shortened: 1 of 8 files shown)


advaita13:
Reference : https://github.com/open-keychain/open-keychain/issues/1670

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,11 @@
     android:installLocation="auto">
 
     <!--
+        General remarks
+        ===============
+        - Last APG 1 version was 10900 (1.0.9 beta 00)
+        - Keychain starting with versionCode 20000!
+
```
dschuermann: These have been removed on git master, you somehow rebased your commits in the wrong way.

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -1733,7 +1734,6 @@
     <string name="help_donation_paypal_item">OpenKeychain Donation</string>
     <string-array name="help_donation_google_catalog_values">
         <item>1 EUR</item>
-        <item>2 EUR</item>
```
dschuermann: Same as my other comment. This has just been added on git master

advaita13:
On account of deletion of a few lines and rearrangement of import statements the commits are showing unnecessary deletions from the Git master. I'll close this request and open up another one with proper changes. Apologies.

Issue #1758, opened 2016-03-06
CertifyKeyActivity back button on action bar not working


007durgesh219:
To Reproduce 
Try to confirm any key with fingerprint or phrases and go back without confirming using back button on the action bar.

PR #1759, opened 2016-03-06
Fix CertifyKeyActivity back button on action bar not working #1758

007durgesh219:


007durgesh219:
@dschuermann I think it will be better to handle the home menu item click in BaseActivity. To avoid those small mistakes in future. Should I do that??

Valodim:
Sounds good to me.

007durgesh219:
I've updated PR. Now it will handle home click in BaseActivity

opg7371:
@Valodim is this issue still open...??

Valodim:
needed to rebase for testing, and pushed from there. thanks :+1:

Issue #1599, opened 2015-11-17
Saving of multiple files


JasperWeiss:
Saving of files not supported. This is a limitation on current android.
I'm sure this should be possible, perhaps with some workaround. 3th party file managers can copy/move multiple files. So why wouldn't Openkeychain be able to store multiple at once?

Wouldn't it be possible to queue the files and store them one by one?

dschuermann:
We don't like to rely on libraries or external file managers for this. However, we could implement this feature for Android 5 with http://developer.android.com/reference/android/content/Intent.html#ACTION_OPEN_DOCUMENT_TREE . At one point this was already planned, that's why the message is shown currently.

Valodim:
Hm. Well we *could* just show five save dialogs one after another...

dschuermann:
@Valodim You mean as a fallback for Android < 5 ?

JasperWeiss:
Ah, I get what the problem is now. It's not about saving multiple files it's just that you can't save multiple files with standard the save dialog. 
Either way you still have to name the individual files unless you keep the original files or you could save them as 1.pgp, 2.pgp etc and then we just need a custom dialog for the directory. Would that be difficult to make?
E.g you select a bunch of files, it asks for a directory and saves them as 1.pgp, 2.pgp, 3.pgp etc. The same way you could select all the .pgp files, decrypt them, and store them again with their original file names. (After selecting a directory)

dschuermann:
@JasperWeiss basically yes, but as I said, on Android 5 there is already an Intent to select a directory (ACTION_OPEN_DOCUMENT_TREE). We don't want to implement a dialog ourselves for this, this is more complicated and error prone than it sounds.

JasperWeiss:
I see. we could ofcourse implement that just for android 5 and show multiple dialogs as a fallback for older versions (or keep the current message)

dschuermann:
I would propose implementing it for Android 5 only and hide the menu item on lower Android versions. No need to put time in legacy Android versions in this case.

JasperWeiss:
That would probably the best thing to do. I'm currently learning Android development myself so I might be able to help in the near future.

dschuermann:
Valodim and I talked about this a little more.
[...]

PR #1760, opened 2016-03-06
WIP: Improve EncryptFilesFragment

runnerway:
We have some limitations in EncryptFilesFragment. Mainly:
- Saving of files not supported <= KitKat
- Saving of multiple files not supported
- No control on output filenames (useful when sharing with email..)

So, i spent some time to improve the situation by solving these problems.
This is a screenshot after this PR at this time (remember it's WIP)

screenshot_20160306-225101

Here we have the list of files we want to Encrypt.
If you tap on a file you get:

screenshot_20160306-225117

Here you can set the filename of the output file. (Probably we can add other file specific options..)
This filename will be used for saving and share actions.

If you tap the save icon in the action bar you get a folder picker to set the path where you want to place the output files. Now i'm using a library for the folder picker because i wanted to share a more or less completed idea in the shortest time to get feedback. But the plan is to write an own folder picker, specific for our use case.

This PR is a WIP because code needs a bit of clean up, some checks are needed, UI need to be improved, bugs need to be fixed, granting of storage permission, i need to write a proper folder picker, i will add gestures etc..
But i wanted to start sharing this work to hear your opinions.
Do you like this idea? Do you have any suggestions before i go deep on this way?

I think all these things are pretty useful.

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -3,6 +3,11 @@ apply plugin: 'witness'
 apply plugin: 'jacoco'
 apply plugin: 'com.github.kt3k.coveralls'
 
+repositories {
+    mavenCentral()
+    maven { url 'http://guardian.github.com/maven/repo-releases' }
+}
+
 dependencies {
     // NOTE: Always use fixed version codes not dynamic ones, e.g. 0.7.3 instead of 0.7.+, see README for more information
     // NOTE: libraries are pinned to a specific build, see below
@@ -61,6 +66,8 @@ dependencies {
     compile 'org.apache.james:apache-mime4j-dom:0.7.2'
     compile 'org.thoughtcrime.ssl.pinning:AndroidPinning:1.0.0'
     compile 'com.cocosw:bottomsheet:1.3.0@aar'
+    compile 'net.rdrei.android.dirchooser:library:3.2@aar'
+    compile 'com.gu:option:1.3'
 
     // Material Drawer
     compile 'com.mikepenz:materialdrawer:4.6.3@aar'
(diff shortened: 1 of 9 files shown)


dschuermann:
Hey, thanks for your work on this and your other awesome pull requests.
Unfortunately, I don't like to introduce an implementation of a save dialog again. Not by us or by an external dependency. We had some security issues due to these in the past (see https://github.com/open-keychain/open-keychain/wiki/cure53-Security-Audit-2015), I suspect that we easily make new vulnerabilities by implementing this again. IMHO it's okay to leave this feature out for users below Kitkat.

What I'd like to see is a possibility of saving multiple encrypted files using http://developer.android.com/reference/android/content/Intent.html#ACTION_OPEN_DOCUMENT_TREE . See https://github.com/open-keychain/open-keychain/issues/1599 for a previous discussion. Currently I would propose saving all encrypted files separately and not in a dedicated zip archive (as discussed in the issue).

runnerway:
Oh sorry. I've missed that discussion. 

Tommorrow I will edit this PR introducing ACTION_OPEN_DOCUMENT_TREE and remove the external dependency. 
Then in next days I will clean up code and setup a good UI until we are satisfied.  

Yes. In this PR encrypted files are already saved separately.

runnerway:
I tried using ACTION_OPEN_DOCUMENT_TREE and it works very well. 
Doing so will also solves other problems.

I'm waiting for my other PR https://github.com/open-keychain/open-keychain/pull/1771 to be finished and merged before pushing the latest version of this PR. 

After that, i will return on this PR to finish this work too.
In this way i will to be able to produce good and robust code.

dschuermann:
Hey @runnerway 
If you are still interested in contributing to OpenKeychain, it would be great to see new progress on this pull request. If not please tell me.

Valodim:
closing stale PR

Issue #1745, opened 2016-03-03
Save split user ids in database


Valodim:
This is a follow-up of #1536. We want to be able to disambiguate user ids in the key list even if they have comments, so we can't simply compare user id strings for this (cf. [here](https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java#L311)) without completely destroying search index performance.

As a first step, the split user ids should be saved in the database in addition to the full user id string. This is redundant, but user ids are short so it's no big deal.

Valodim:
I have a concrete need for this, if someone gave this a go I would appreciate it. (I'll probably do it myself sometime next week otherwise)

fjodorver:
First step is done in pull request #1745

fjodorver:
pull request is updated

fjodorver:
Updated HAS_DUPLICATE_USER_ID in KeychainProvider

dschuermann:
done in https://github.com/open-keychain/open-keychain/pull/1763

PR #1763, opened 2016-03-07
WIP: Save split user ids in database #1745

fjodorver:
Save split user ids in database


Valodim:
Looks good. However, this is missing a migration path from an old database format, and should also increase the db version.

fjodorver:
pull request is updated?

vershinin:
@Valodim, could you have a look?

Valodim:
Merged this PR, if you want to do the other thing simply open a new one. Thanks! :+1:

PR #1764, opened 2016-03-07
Display Key in UploadKeyActivity Issue #506

### LINKED ISSUE #506: Display key in UploadKeyActivity
Display key in UploadKeyActivity like it is done in CertifyKeyActivity.

To prevent duplicated code put this view and the loader in a fragment

007durgesh219:
screenshot_2016-03-07-17-36-11
Fix #506

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyActivity.java
@@ -28,6 +28,7 @@ import org.sufficientlysecure.keychain.ui.base.BaseActivity;
 public class CertifyKeyActivity extends BaseActivity {
 
     public static final String EXTRA_RESULT = "operation_result";
+    // For sending masterKeyIds to MultiUserIdsFragment to display list of keys
     public static final String EXTRA_KEY_IDS = MultiUserIdsFragment.EXTRA_KEY_IDS ;
     public static final String EXTRA_CERTIFY_KEY_ID = "certify_key_id";
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyActivity.java
@@ -28,7 +28,7 @@ import org.sufficientlysecure.keychain.ui.base.BaseActivity;
 public class CertifyKeyActivity extends BaseActivity {
 
     public static final String EXTRA_RESULT = "operation_result";
-    public static final String EXTRA_KEY_IDS = "extra_key_ids";
+    public static final String EXTRA_KEY_IDS = MultiUserIdsFragment.EXTRA_KEY_IDS ;
     public static final String EXTRA_CERTIFY_KEY_ID = "certify_key_id";
 
     @Override
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyFragment.java
@@ -62,52 +62,19 @@ import java.util.ArrayList;
 import java.util.Date;
 
 public class CertifyKeyFragment
-        extends CachingCryptoOperationFragment<CertifyActionsParcel, CertifyResult>
-        implements LoaderManager.LoaderCallbacks<Cursor> {
-
-    public static final String ARG_CHECK_STATES = "check_states";
+        extends CachingCryptoOperationFragment<CertifyActionsParcel, CertifyResult> {
 
     private CheckBox mUploadKeyCheckbox;
-    ListView mUserIds;
 
     private CertifyKeySpinner mCertifyKeySpinner;
 
-    private long[] mPubMasterKeyIds;
-
-    public static final String[] USER_IDS_PROJECTION = new String[]{
-            UserPackets._ID,
-            UserPackets.MASTER_KEY_ID,
-            UserPackets.USER_ID,
-            UserPackets.IS_PRIMARY,
-            UserPackets.IS_REVOKED
-    };
-    private static final int INDEX_MASTER_KEY_ID = 1;
-    private static final int INDEX_USER_ID = 2;
-    @SuppressWarnings("unused")
-    private static final int INDEX_IS_PRIMARY = 3;
-    @SuppressWarnings("unused")
-    private static final int INDEX_IS_REVOKED = 4;
-
-    private MultiUserIdsAdapter mUserIdsAdapter;
+    private MultiUserIdsFragment mMultiUserIdsFragment;
 
     @Override
     public void onActivityCreated(Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
 
-        mPubMasterKeyIds = getActivity().getIntent().getLongArrayExtra(CertifyKeyActivity.EXTRA_KEY_IDS);
-        if (mPubMasterKeyIds == null) {
-            Log.e(Constants.TAG, "List of key ids to certify missing!");
-            getActivity().finish();
-            return;
-        }
-
-        ArrayList<Boolean> checkedStates;
-        if (savedInstanceState != null) {
-            checkedStates = (ArrayList<Boolean>) savedInstanceState.getSerializable(ARG_CHECK_STATES);
-            // key spinner and the checkbox keep their own state
-        } else {
-            checkedStates = null;
-
+        if (savedInstanceState == null){
             // preselect certify key id if given
             long certifyKeyId = getActivity().getIntent()
                     .getLongExtra(CertifyKeyActivity.EXTRA_CERTIFY_KEY_ID, Constants.key.none);
@@ -124,12 +91,6 @@ public class CertifyKeyFragment
 
         }
 
-        mUserIdsAdapter = new MultiUserIdsAdapter(getActivity(), null, 0, checkedStates);
-        mUserIds.setAdapter(mUserIdsAdapter);
-        mUserIds.setDividerHeight(0);
-
-        getLoaderManager().initLoader(0, null, this);
-
         OperationResult result = getActivity().getIntent().getParcelableExtra(CertifyKeyActivity.EXTRA_RESULT);
         if (result != null) {
             // display result from import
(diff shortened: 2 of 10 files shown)


Valodim:
Huh. Do the checkmarks actually do anything here?

007durgesh219:
No, It's useless. We can hide it. I thought of hiding. Just wanted to discuss before doing that. I will hide it.

007durgesh219:
I've updated the PR. Please review it.

screenshot_20160313-002038

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyActivity.java
```diff
@@ -28,7 +28,7 @@
 public class CertifyKeyActivity extends BaseActivity {
 
     public static final String EXTRA_RESULT = "operation_result";
-    public static final String EXTRA_KEY_IDS = "extra_key_ids";
+    public static final String EXTRA_KEY_IDS = MultiUserIdsFragment.EXTRA_KEY_IDS ;
```
Valodim: we don't do this anywhere else, please leave the constants individual

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/adapter/MultiUserIdsAdapter.java
```diff
@@ -138,6 +145,11 @@ public void onCheckedChanged(CompoundButton compoundButton, boolean b) {
             }
         });
         vCheckBox.setClickable(false);
+        if (checkboxVisibility) {
```
Valodim: nicer with ?: operator

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/UploadKeyActivity.java
```diff
@@ -89,15 +90,6 @@ public void onClick(View v) {
             return;
         }
 
-        try {
```
Valodim: why was this removed?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/UploadKeyActivity.java
```diff
@@ -53,8 +50,8 @@
 
     // CryptoOperationHelper.Callback vars
     private String mKeyserver;
-    private long mMasterKeyId;
     private CryptoOperationHelper<UploadKeyringParcel, UploadResult> mUploadOpHelper;
+    private MultiUserIdsFragment mMultiUserIdsFragment;
```
Valodim: can be a local var

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CertifyKeyFragment.java
```diff
@@ -62,52 +62,19 @@
...
-        }
-
-        ArrayList<Boolean> checkedStates;
-        if (savedInstanceState != null) {
-            checkedStates = (ArrayList<Boolean>) savedInstanceState.getSerializable(ARG_CHECK_STATES);
-            // key spinner and the checkbox keep their own state
-        } else {
-            checkedStates = null;
-
+        if (savedInstanceState == null){
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
```diff
@@ -0,0 +1,223 @@
...
+import org.sufficientlysecure.keychain.provider.KeychainContract;
+import org.sufficientlysecure.keychain.provider.KeychainDatabase;
+import org.sufficientlysecure.keychain.service.CertifyActionsParcel;
+import org.sufficientlysecure.keychain.ui.adapter.MultiUserIdsAdapter;
+import org.sufficientlysecure.keychain.util.Log;
+
+import java.util.ArrayList;
+
+/**
+ * Created by durgeshchoudhary on 07/03/16.
```
Valodim: pointless comment, please remove

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
```diff
@@ -0,0 +1,223 @@
...
+    public void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+
+        ArrayList<Boolean> states = mUserIdsAdapter.getCheckStates();
+        // no proper parceling method available :(
+        outState.putSerializable(ARG_CHECK_STATES, states);
+    }
+
+    public ArrayList<CertifyActionsParcel.CertifyAction> getSelectedCertifyActions() {
+        return mUserIdsAdapter.getSelectedCertifyActions();
```
Valodim: throw AssertionError if checkboxVisibility is false

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
```diff
@@ -0,0 +1,223 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Created by durgeshchoudhary on 07/03/16.
+ */
+public class MultiUserIdsFragment extends Fragment implements LoaderManager.LoaderCallbacks<Cursor>{
+    public static final String ARG_CHECK_STATES = "check_states";
+    public static final String EXTRA_KEY_IDS = "extra_key_ids";
+    // Indicating whether checkbox should be visible or not (Default visible)
```
Valodim: pointless comment: either the comment is unneeded, or the variable name isn't descriptive enough

Valodim:
nicely done with the fragment extraction :+1:

(sorry for the late review)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
007durgesh219: Those are automatically inserted by AndroidStudio  Anyway I will remove them. Thanks for pointing out

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/UploadKeyActivity.java
007durgesh219: Because now there is no need of getting masterKeyId from DataUri. As I am sending masterKeyId in Intent itself while calling this activity itself.  So, this try-cache is moved there.  Please see ViewKeyAdvShareFragment.java where I am calculating masterKeyId from dataUri and sending to this activity with Intent.

007durgesh219:
I've updated PR with suggested changes. Please review it again

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
```diff
@@ -455,8 +455,17 @@ protected void onPostExecute(Bitmap qrCode) {
     }
 
     private void uploadToKeyserver() {
+        long keyId = 0;
```
Valodim: this error seems like it's not handled correctly

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
```diff
@@ -0,0 +1,221 @@
...
+    public void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+
+        ArrayList<Boolean> states = mUserIdsAdapter.getCheckStates();
+        // no proper parceling method available :(
+        outState.putSerializable(ARG_CHECK_STATES, states);
+    }
+
+    public ArrayList<CertifyActionsParcel.CertifyAction> getSelectedCertifyActions() {
+        assert(checkboxVisibility);
```
Valodim: the assert statement is optimized away in release builds, but we do want a crash report if this happens for some reason. please use `if (...) { throw new AssertionError("description"); }`

Valodim:
the "Identities for" headline is not correctly aligned, there is a little too much space to the left (can also be seen in your screenshot). just tested it and looks good otherwise! :+1:

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
007durgesh219: I think adding return in catch block will do. Is that what you mean??

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MultiUserIdsFragment.java
007durgesh219: Should I also include throws AssertionError to explicitly tell that this method might through an exception So, that it can be handled by the caller.

Valodim: The point of [errors](https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html) is that they are never caught by the caller. This is an illegal state caused by an error in programming that should be fixed, so we want the app to crash if it happens.

007durgesh219:
I've updated the PR. Now its aligned.  Please Review it again. 
screenshot_20160331-165237

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
Valodim: Correct. In this case, the keyId variable doesn't have to be initialized - if the compiler complains about that, it means the variable might be used when there's a 0 in there, which is a sign that something is awry.

Apart from that, we use snackbars instead of toasts

007durgesh219: I've updated it with suggested changes

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/adapter/MultiUserIdsAdapter.java
```diff
@@ -46,6 +47,11 @@ public MultiUserIdsAdapter(Context context, Cursor c, int flags, ArrayList<Boole
         mCheckStates = preselectStates == null ? new ArrayList<Boolean>() : preselectStates;
     }
 
+    public MultiUserIdsAdapter(Context context, Cursor c, int flags, ArrayList<Boolean> preselectStates, boolean checkboxVisibility) {
+        this(context,c,flags,preselectStates);
```
Valodim: code style: spaces

Valodim:
Cool, looks good now! Thanks for your work :+1:

Issue #1228, opened 2015-04-27
Link to search from AutoTextComplete on empty result set


Valodim:

advaita13:
Isn't there a search button just beside the AutoCompleteTextView field?

daquexian:
Hi :) Does this issue still need work? It seems that there is already a search button on empty result set.

Valodim:
I think this issue refers to what was attempted in #1765, i.e. a link to search from the dropdown that appears in the "Encrypt to" field.

Thinking about this now, I'm not so sure that there is a nice way to do this, breaking out into a full dialog from a dropdown seems messy. Thoughts @dschuermann ?

dschuermann:
I am no longer sure if this is an important issue worth pursuing. We should focus on different aspects in OpenKeychain. I will close this for now. If you think otherwise @Valodim please re-open.

PR #1765, opened 2016-03-07
Link to search from AutoTextComplete

advaita13:
Issue : https://github.com/open-keychain/open-keychain/issues/1228

Added a small search button besides `EncryptKeyCompletionView`. Directly performs cloud key search using the text obtained from the view.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
@@ -17,6 +17,8 @@
 
 package org.sufficientlysecure.keychain.ui;
 
+import android.app.Activity;
+import android.content.Intent;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -50,6 +52,7 @@ public class EncryptModeAsymmetricFragment extends EncryptModeFragment {
 
     private KeySpinner mSignKeySpinner;
     private EncryptKeyCompletionView mEncryptKeyView;
+    private String mQuery;
 
     public static final String ARG_SINGATURE_KEY_ID = "signature_key_id";
     public static final String ARG_ENCRYPTION_KEY_IDS = "encryption_key_ids";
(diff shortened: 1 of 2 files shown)


advaita13:
@adithyaphilip Could you pls review and merge this.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
```diff
@@ -50,6 +52,7 @@
 
     private KeySpinner mSignKeySpinner;
     private EncryptKeyCompletionView mEncryptKeyView;
+    private String mQuery;
```
adithyaphilip: Why is this global? Try to get rid of all the warnings shown in Android Studio, that should prevent issues like this ("field can be converted to local").

adithyaphilip:
This does let you search for missing keys, but it might be better to make the option part of the drop down itself (e.g. if there are no keys, the` EncryptKeyCompletionView` shows `Search for keys with  "John" online`). @Valodim your thoughts?

Valodim:
I concur. The search option should be in the dropdown, when no key can be found for the typed text.

The KeyAdapter class already has some logic which is used in KeyListFragment to display the "You have no keys yet!" dummy item. It should be fairly simple to add a way to display an item based on the search query (which it already knows).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
advaita13: I wasn't getting any such warning.
Will make the change. Thanks.

advaita13:
On it. Thanks guys.

advaita13:
KeyListFragment has a different implementation of it altogether.
For this I will need to populate the list with `Search for keys with "John" online` and implement some sort of `OnClickListener` to handle search. Could you guys guide me a bit about how that drop-down menu is populated?

Valodim:
The dropdown menu is populated from the database using a loader and adapter, also using a holder pattern.

advaita13:
If the items in the dropdown menu are returned from the database directly then will it be possible to add a custom menu item?
Or how about using `swapCursor` to add a dummy cursor object on the dropdown?

adithyaphilip:
Sure, go ahead and implement what you have in mind and we can take it from there. `swapCursor` seems to be the way to go. Also look at the `getViewForObject` function.

advaita13:
I've been trying to get around this problem in multiple ways, no breakthrough so far. Here's why...
I created a MatrixCursor and added a row to it containing mostly null values to see if I can get it to show something on the drop down. Doesn't work. The dummy row (filled with mostly null values) isn't the right way to go since many other methods perform certain actions on the values and hence give an exception when those values are null.
As a result there is not way to pass off a dummy cursor without causing dependent functions to throw runtime errors.

The search button besides seems like the way to go for now. The button allows you to search for keys even beyond the query results displayed in the drop down.

dschuermann:
Please open a new pull request if you have new implementation ideas. I will close this one for now.

PR #1767, opened 2016-03-08
Sync only when unmetered internet is available? #1670

advaita13:
@adithyaphilip I have included the changes you suggested on the previous pull request.

However there were 2 issues left unresolved (refer notes). There might be a case where the wifi is off when the periodic sync takes place. The sync should either be postponed or resumed once wifi is back else it will never happen (unless wifi is active when next scheduled sync takes place, even then it can only possibly happen once in 3 days). What are your thoughts on that?

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -76,6 +76,8 @@
 
     <!-- other group (for free) -->
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.NFC" />
     <uses-permission android:name="android.permission.READ_SYNC_SETTINGS" />
     <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS" />
@@ -89,6 +91,18 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/Theme.Keychain.Light">
+        <!-- broadcast receiver for Wi-Fi Connection -->
+        <receiver
+            android:name=".receiver.NetworkReceiver"
+            android:enabled="true"
+            android:exported="true" >
+            <intent-filter>
+                <action android:name="android.net.wifi.WIFI_STATE_CHANGE"/>
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.net.wifi.STATE_CHANGE"/>
+            </intent-filter>
+        </receiver>
         <!-- singleTop for NFC dispatch, see SecurityTokenOperationActivity -->
         <activity
             android:name=".ui.MainActivity"
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -94,7 +94,7 @@
         <!-- broadcast receiver for Wi-Fi Connection -->
         <receiver
             android:name=".receiver.NetworkReceiver"
-            android:enabled="true"
+            android:enabled="false"
             android:exported="true" >
             <intent-filter>
                 <action android:name="android.net.wifi.WIFI_STATE_CHANGE"/>
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -97,10 +97,7 @@
             android:enabled="false"
             android:exported="true" >
             <intent-filter>
-                <action android:name="android.net.wifi.WIFI_STATE_CHANGE"/>
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.net.wifi.STATE_CHANGE"/>
+                <action android:name="android.net.conn.CONNECTIVITY_CHANGE"/>
             </intent-filter>
         </receiver>
         <!-- singleTop for NFC dispatch, see SecurityTokenOperationActivity -->
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
@@ -3,6 +3,7 @@ package org.sufficientlysecure.keychain.receiver;
 import android.content.BroadcastReceiver;
 import android.content.ComponentName;
 import android.content.Context;
+import org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.ConnectivityManager;
@@ -12,6 +13,7 @@ import org.sufficientlysecure.keychain.Constants;
 import org.sufficientlysecure.keychain.util.Log;
 
 public class NetworkReceiver extends BroadcastReceiver {
+    private static final String ACTION_UPDATE_ALL = "update_all";
 
     Context context;
     public NetworkReceiver(){}
@@ -27,7 +29,8 @@ public class NetworkReceiver extends BroadcastReceiver {
 
             // broadcaster receiver disabled
             setWifiReceiverComponent(false, context);
-            //KeyserverSyncAdapterService.enableKeyserverSync(context);
+            Intent serviceIntent = new Intent(context, KeyserverSyncAdapterService.class);
+            serviceIntent.setAction(ACTION_UPDATE_ALL);
         }
     }
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
@@ -13,10 +13,6 @@ import org.sufficientlysecure.keychain.Constants;
 import org.sufficientlysecure.keychain.util.Log;
 
 public class NetworkReceiver extends BroadcastReceiver {
-    private static final String ACTION_UPDATE_ALL = "update_all";
-
-    Context context;
-    public NetworkReceiver(){}
 
     @Override
     public void onReceive(Context context, Intent intent) {
@@ -30,7 +26,7 @@ public class NetworkReceiver extends BroadcastReceiver {
             // broadcaster receiver disabled
             setWifiReceiverComponent(false, context);
             Intent serviceIntent = new Intent(context, KeyserverSyncAdapterService.class);
-            serviceIntent.setAction(ACTION_UPDATE_ALL);
+            serviceIntent.setAction(KeyserverSyncAdapterService.ACTION_UPDATE_ALL);
         }
     }
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
@@ -1,0 +1,51 @@
+package org.sufficientlysecure.keychain.receiver;
+
+import android.content.BroadcastReceiver;
+import android.content.ComponentName;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+
+import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.util.Log;
+
+public class NetworkReceiver extends BroadcastReceiver {
+
+    Context context;
+    public NetworkReceiver(){}
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+
+        ConnectivityManager conn = (ConnectivityManager)
+                context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = conn.getActiveNetworkInfo();
+
+        if (networkInfo != null && networkInfo.getType() == ConnectivityManager.TYPE_WIFI) {
+
+            // broadcaster receiver disabled
+            setWifiReceiverComponent(false, context);
+            //KeyserverSyncAdapterService.enableKeyserverSync(context);
+        }
+    }
+
+    public void setWifiReceiverComponent(Boolean isEnabled, Context context){
+
+        PackageManager pm = context.getPackageManager();
+        ComponentName compName = new ComponentName(context,
+                NetworkReceiver.class);
+
+        if(isEnabled) {
+            pm.setComponentEnabledSetting(compName,
+                    PackageManager.COMPONENT_ENABLED_STATE_ENABLED, PackageManager.DONT_KILL_APP);
+            Log.d(Constants.TAG, "Wifi Receiver is enabled!");
+        }
+        else {
+            pm.setComponentEnabledSetting(compName,
+                    PackageManager.COMPONENT_ENABLED_STATE_DISABLED, PackageManager.DONT_KILL_APP);
+            Log.d(Constants.TAG, "Wifi Receiver is disabled!");
+        }
+    }
+}
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
@@ -3,13 +3,13 @@ package org.sufficientlysecure.keychain.receiver;
 import android.content.BroadcastReceiver;
 import android.content.ComponentName;
 import android.content.Context;
-import org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 
 import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.service.KeyserverSyncAdapterService;
 import org.sufficientlysecure.keychain.util.Log;
 
 public class NetworkReceiver extends BroadcastReceiver {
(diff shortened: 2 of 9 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
```diff
@@ -0,0 +1,51 @@
...
+
+        ConnectivityManager conn = (ConnectivityManager)
+                context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = conn.getActiveNetworkInfo();
+
+        if (networkInfo != null && networkInfo.getType() == ConnectivityManager.TYPE_WIFI) {
+
+            // broadcaster receiver disabled
+            setWifiReceiverComponent(false, context);
+            //KeyserverSyncAdapterService.enableKeyserverSync(context);
```
advaita13: This will force a sync to take place once a wifi connection is established.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
```diff
@@ -176,8 +183,26 @@ public KeyserverSyncAdapter() {
...
+                ConnectivityManager connMgr = (ConnectivityManager)
+                        getSystemService(Context.CONNECTIVITY_SERVICE);
+                @SuppressWarnings("deprecation") // our min is API 15, deprecated only in 23
+                        NetworkInfo networkInfo = connMgr.getNetworkInfo(ConnectivityManager.TYPE_WIFI);
+                boolean isWifiConn = networkInfo.isConnected();
+
+                // if Wi-Fi connection doesn't exist then receiver is enabled
+                if(!isWifiConn){
+                    new NetworkReceiver().setWifiReceiverComponent(true, getContext());
+                    //postponeSync();
```
advaita13: Or, we could postpone sync if it fails to happen due to unavailability of wifi.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
adithyaphilip: `enableKeyserverSync` is meant to enable the periodic 3 day sync. Take a look at `SYNC_NOW` in `KeyserverSyncAdapterService` for a more deterministic way of starting a sync (start the service with that intent).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
adithyaphilip: The reason I believe this is unnecessary is if a sync fails due to unavailability of WiFi, all subsequent postponed syncs will also fail, until WiFi is available. When WiFi becomes available, we are triggering a sync anyway, so we don't need to depend on a postponed sync occurring when WiFi is present.

advaita13: Agreed. It was either of those anyway.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SettingsActivity.java
```diff
@@ -436,6 +436,18 @@ public void onResume() {
                     account,
                     ContactsContract.AUTHORITY
             );
+
+            SwitchPreference pref = (SwitchPreference) findPreference(Constants.Pref.ENABLE_WIFI_SYNC_ONLY);
```
adithyaphilip: The entire `onPreferenceChangeListener` can be avoided now that you have set android:persistent=true.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -424,6 +424,18 @@ public CloudSearchPrefs createFromParcel(Parcel in) {
         };
     }
 
+    // sync preferences
+
+    public void setWifiOnlySync(Boolean isSyncEnabled){
```
adithyaphilip: This function can be skipped too, since `android:persistent=true`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
```diff
@@ -74,6 +77,10 @@
     private static final String ACTION_CANCEL = "cancel";
 
     private AtomicBoolean mCancelled = new AtomicBoolean(false);
+    Context context;
+
+    //zero argument constructor
+    public KeyserverSyncAdapterService(){}
```
adithyaphilip: Doesn't Java supply a zero argument constructor by default if you haven't specified any other constructor? Shouldn't be required now that you removed the other constructor that was here earlier.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
```diff
@@ -0,0 +1,54 @@
...
+import android.net.NetworkInfo;
+
+import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.util.Log;
+
+public class NetworkReceiver extends BroadcastReceiver {
+    private static final String ACTION_UPDATE_ALL = "update_all";
+
+    Context context;
+    public NetworkReceiver(){}
```
adithyaphilip: AFAIK, this shouldn't be necessary now since there are no other constructors, could you remove this and test again?

advaita13: Sure. I'll incorporate all the changes and do a final test. Thanks.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
```diff
@@ -0,0 +1,54 @@
...
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+
+import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.util.Log;
+
+public class NetworkReceiver extends BroadcastReceiver {
+    private static final String ACTION_UPDATE_ALL = "update_all";
```
adithyaphilip: Constants are used to prevent copy-paste errors and also so that changing that value of a constant doesn't require making the same change in other places it's used. Please make the constant in the service public and access it instead of creating another one here.

advaita13: The purpose of NetworkReceiver is to initiate sync as soon as wifi is restored. UPDATE_ALL will ensure that happens while SYNC_NOW will delay the sync till the screen is turned off. Since the sync has already been delayed (due to absence of wifi) I did not want to delay it any further. Also, the user might turn off the wifi before turning the screen off which would cause further delay in sync.

dschuermann:
@advaita13 For future commits: Please leave out the "Update xyz.java" from your commit messages. It is unnecessary. It is also allowed to change more than one file in one commit :smiling_imp: , please try to make changes that belong together in one commit.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
```diff
@@ -509,7 +530,7 @@ private Notification getOrbotNoification(Context context) {
         return builder.build();
     }
 
-    public static void enableKeyserverSync(Context context) {
+    public static void enableKeyserverSync(Context context) {]
```
adithyaphilip: A little curious as to how compile errors are sneaking in to your commits.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/KeyserverSyncAdapterService.java
```diff
@@ -67,13 +70,14 @@
...
     private static final String ACTION_IGNORE_TOR = "ignore_tor";
-    private static final String ACTION_UPDATE_ALL = "update_all";
+    public static final String ACTION_UPDATE_ALL = "update_all";
     private static final String ACTION_SYNC_NOW = "sync_now";
     private static final String ACTION_DISMISS_NOTIFICATION = "cancel_sync";
     private static final String ACTION_START_ORBOT = "start_orbot";
     private static final String ACTION_CANCEL = "cancel";
 
     private AtomicBoolean mCancelled = new AtomicBoolean(false);
+    Context context;
```
adithyaphilip: Unused variable? Please make sure you correct/explain all the warnings shown by Studio.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/receiver/NetworkReceiver.java
```diff
@@ -0,0 +1,50 @@
...
+        ConnectivityManager conn = (ConnectivityManager)
+                context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo networkInfo = conn.getActiveNetworkInfo();
+
+        if (networkInfo != null && networkInfo.getType() == ConnectivityManager.TYPE_WIFI) {
+
+            // broadcaster receiver disabled
+            setWifiReceiverComponent(false, context);
+            Intent serviceIntent = new Intent(context, KeyserverSyncAdapterService.class);
+            serviceIntent.setAction(KeyserverSyncAdapterService.ACTION_UPDATE_ALL);
```
adithyaphilip: You have to send the intent to the service using `startService`, please test your code thoroughly first.

advaita13: Yes, I hadn't placed the log messages everywhere so previously the logcat was showing the messages I was hoping for. Been more thorough this time.

adithyaphilip:
Hmm, it doesn't seem to really be working for me. @advaita13 is it working for you? What was the procedure you used to test it?

advaita13:
I placed a log message after `context.startService(serviceIntent);` in NetworkReceiver and another one inside `asyncKeyUpdate` just to verify if the update is really taking place.

Also, I changed the SYNC_INTERVAL and other constants to 1 second for testing

adithyaphilip:
I played around with this for a bit, and it looks like on my OnePlus One the broadcast receiver is called only when I disable WiFi, not sure why. Could you check if this is the same on your device? This <a href="http://stackoverflow.com/questions/5888502/how-to-detect-when-wifi-connection-has-been-established-in-android">SO link</a> might have some information on how to get this working.

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -89,6 +91,18 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/Theme.Keychain.Light">
+        <!-- broadcast receiver for Wi-Fi Connection -->
+        <receiver
+            android:name=".receiver.NetworkReceiver"
+            android:enabled="true"
```
adithyaphilip: Shouldn't this be `false` initially?

advaita13:
Yes, I checked and rechecked it on my Nexus 5. It works. Sometimes it takes a while (not more than a minute) but it definitely works.

adithyaphilip:
Hmm, if you're looking at the update logs only, are you sure it's not our periodic sync firing again (you had mentioned that you had set it to one minute)? Could you please log the following in `onReceive` (before you check for if it's WiFi):
```
        if (networkInfo != null) {
            Log.d(Constants.TAG, "calling wifi " + networkInfo.getTypeName());
        }
```
and see if it occurs when you enable or when you disable your WiFi? It's definitely not only on disabling with the OnePlus, will test with more phones tomorrow.

advaita13:
No, it's not due to the periodic sync. I have put log messages in the `setWifiReceiverComponent` method. Those tell me when the receiver has been enabled and disabled.

Also, the reason you don't see it on enabling because the `onReceive` method doesn't get called then (the receiver is disabled initially). Once enabled `onReceive` will get called whenever Wifi status will change.

adithyaphilip:
It was tested after I removed the bit that disables the receiver, and adding the log statement I had mentioned in my previous comment. I can see `onReceive` being called every time I disable WiFI, for a change in WiFi state, but not when I enable it.

advaita13:
Tested it the same way you just mentioned. Works alright in both cases for me...

> 03-12 02:11:38.445 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:11:38.467 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:12:02.663 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI
> 03-12 02:12:02.702 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI

adithyaphilip:
Great. Might be device related then. Will check with a few more devices tomorrow and get back to you. Was the log you posted the result of you turning your WiFi on and off, or just on?

advaita13:
It was both. Tried it once again, same results.

adithyaphilip:
The reason I ask is because the timestamps on the Wifi changes were so close together (~40ms) that they both probably correspond to one of the two - enabling or disabling. Did you prevent the BroadcastReceiver from being disabled for this test?
```
03-12 02:12:02.663 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI
03-12 02:12:02.702 27715-27715/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI
```

advaita13:
Yes I removed the bit that disables the receiver.

Those ~40ms changes are the changes in the wifi state (enabled, connecting, connected, etc). That happens naturally when you enable wifi. The receiver not only listens for network state changes but also wifi state changes.

adithyaphilip:
Sure, that's alright. What I was saying is that if you disabled and then enabled WiFi, and those logs were meant to show both your enabling and disabling, then what we see (the multiple logs) should correspond to only one of those events - either the WiFi being disabled, or enabled. I guess logging an `isConnected()` should remove all ambiguity, unless of course the logs show you disabling it some point earlier and you just hadn't posted that bit here.

advaita13:
Could be, but those log events corresponded to when I disabled wifi and then enabled it again.

Here are more logs...

> 03-12 02:56:16.261 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.281 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.310 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.328 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.345 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.361 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.377 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.396 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:16.414 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi MOBILE
> 03-12 02:56:56.784 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI
> 03-12 02:56:56.826 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI
> 03-12 02:56:56.850 12743-12743/org.sufficientlysecure.keychain.debug D/Keychain D: calling wifi WIFI

I disabled wifi immediately after which all the "calling wifi mobile" logs got generated. Then waited about 10 seconds and enabled wifi again, after around 2-3 seconds the last 3 log messages were generated.

adithyaphilip:
Nice. Seems to be working then. Will just confirm with a few more devices tomorrow before we go ahead.

advaita13:
Sure. Thanks a lot :)

dschuermann:
I have a feature request :) Can we disable the enableWifiSyncOnly SwitchPreference when the syncKeyserver SwitchPreference is not checked?

advaita13:
Absolutely. Makes no sense otherwise.
Thanks :)

adithyaphilip:
Okay, tested it out on a Moto G 2nd Gen too, and the behaviour was erratic. Worked once or twice out of the several tests I ran. Seemed to work once in a while on the One Plus One too. Not really sure what's wrong, since by looking at the docs it *should* work.

Anyway, replacing the intent filters with just `android.net.conn.CONNECTIVITY_CHANGE` seemed to do the trick on both devices. Could you tell me if it works for you too? If it does we can go ahead with that.

advaita13:
Yes, works perfectly. Thanks :)

adithyaphilip:
Great. Could you please squash/fixup everything into a single commit and rebase on master, so we can do a final review before we merge?

advaita13:
Will it be alright if I open up a new PR? Something went wrong with rebase (my commits aren't reflected on it) so it'll be much cleaner to to open up a new PR.

Really sorry for this. I was planning to remove and clone the entire project once again once this PR was merged.

adithyaphilip:
Sure, go ahead. Though learning to get the rebase process right should come in handy - we tend to have to do that a lot.

advaita13:
Certainly :)

It's broken beyond repair now. A fresh start would ensure things go smoothly hence forth.

PR #1771, opened 2016-03-08
PgpSignEncryptOperation refactoring

runnerway:
This refactoring is needed for two reasons:

1) Because https://github.com/open-keychain/open-keychain/blob/8b8b91e24722f254ad84916d3e1b6c6210906924/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignEncryptOperation.java#L108
2) Because it's needed to produce good code for my other PR #1760

It's a WIP because it needs a bit more testing, a bit of code clean up and probably tests should be adapted too. This doesn't introduce functional changes. I will continue to work on this during next days. Advices and suggestions are welcome :)

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignEncryptOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignEncryptOperation.java
@@ -20,6 +20,7 @@
 package org.sufficientlysecure.keychain.pgp;
 
 import android.content.Context;
+import android.net.Uri;
 import android.os.Parcelable;
 import android.support.annotation.NonNull;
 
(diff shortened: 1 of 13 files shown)


dschuermann:
Just want to say: Awesome to see progress in this area of the backend :)

runnerway:
It's good to see it's appreciated :) I was a bit busy during this week. 
I will improve this PR, so it can be merged when finished.
After this, i can continue the work on the EncryptFilesFragment.

Valodim:
Phew, this PR looks like it will be a handful to review. Do you have any more changes you wanted to get in here in mind? Just drop a note and I will review when you are finished.

runnerway:
This should be ready for a review, i need only to check if some tests are broken because of this.(Probably i will check this today. If all is OK i will drop a note here)

runnerway:
- Fixed a typo
- Adapted test

This should be ready for review. :)

Valodim:
It's on my list. Sorry for the delay :)

runnerway:
The delay was mine. 
There were some problems that needed to be addressed.

Now it should be OK.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignEncryptOperation.java
```diff
@@ -471,7 +527,7 @@ public PgpSignEncryptResult execute(PgpSignEncryptInputParcel input, CryptoInput
 
                 if (enableCompression) {
                     // Use preferred compression algo
-                    int algo = input.getCompressionAlgorithm();
+                    int algo = data.getCompressionAlgorithm();
```
Valodim: for the future, please avoid refactorings like this, or put them in a separate commit. they add a lot of lines to the review which contain no semantic changes

runnerway:
Fixed a remaining little bug of this PR. Now it should be ok.

@Valodim 
I agree that this commit is bigger than what it should be.
(The final size is the result of incremental changes closely connected, but now i see how it could be split at least into 2 different commits)
Next time i will split a change in smaller commits.

But i think there is a semantic change in the line you pointed out. 
This is a quick explanation on this entire PR:

Previously "SignEncryptParcel" was a subclass of "PgpSignEncryptInputParcel". 
But this doesn't seem logically correct. Said in a simple way (leaving out some details):
- SignEncryptParcel is an input to an operation containing one or more inputs to be encrypted using the same options.
- PgpSignEncryptInputParcel is an input to an operation containing one input to be encrypted.
So, i broke this relationship and introduced the class "PgpSignEncryptData" (Maybe we should use another name?). This class contains the encryption options for all the inputs inside a SignEncryptParcel.

The flow after this PR will be: 
1) Take a SignEncryptParcel
2) For every input in this parcel create a PgpSignEncryptInputParcel taking the PgpSignEncryptData from SignEncryptParcel as options for the current PgpSignEncryptInputParcel.

This is the first result achieved with this PR.
Doing so we are able to refactor "PgpSignEncryptOperation" to fix its implementation of the "BaseOperation" interface. And this is the second result achieved with this PR.

All these changes will help me in finishing the work on https://github.com/open-keychain/open-keychain/pull/1760 for saving multiple output files.

runnerway:
@Valodim Is it still on your list or there is something wrong with this? :)

dschuermann:
I removed "WIP" from the title, as this is definitely to be reviewed next.

dschuermann:
@runnerway Can you rebase on top of current master and then ping me in this comment section? I will then merge the pr.

runnerway:
For sure :) I will do it very soon.

dschuermann:
@runnerway ping

runnerway:
@dschuermann Pong :)

I'm sorry for the delay but this week was very busy.
Rebased. Tests run without problems. 
Tomorrow i will try on a real device, but all should be good.

dschuermann:
Great work! I am excited your upcoming work :)

Issue #1673, opened 2016-01-12
handle encrypted files in "import key" file selection


Valodim:
When the user selects an encrypted file for import, they currently get an error. Since we export encrypted keys, importing it back via the import from file dialog seems like a workflow users might want to use, so we should support that.

Valodim:
I noticed this is actually quite important, since right now there is no good way to restore from a backup from our first time dialog.

Juaanma:
Hi, I would like to contribute to this issue. After @runnerway's solution, it's still needed to incorporate his idea to the first time dialog, right?

Valodim:
Not really, the first time dialog simply links to the ImportKeysFragment, so that should already work :)

rhari991:
@dschuermann Seems this issue has been solved ?

dschuermann:
@rhari991 you are right. this has already been done.

PR #1772, opened 2016-03-09
Handle encrypted files in "import key" file selection

runnerway:
This is a solution for:
https://github.com/open-keychain/open-keychain/issues/1673

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -416,11 +416,18 @@ public class ImportKeysActivity extends BaseActivity
             intent.putExtra(ImportKeyResult.EXTRA_RESULT, result);
             setResult(RESULT_OK, intent);
             finish();
-            return;
-        }
+        } else if (result.isOkNew() || result.isOkUpdated()) {
+            // User has successfully imported a key, hide first time dialog
+            Preferences.getPreferences(this).setFirstTime(false);
 
-        result.createNotify(ImportKeysActivity.this)
-                .show((ViewGroup) findViewById(R.id.import_snackbar));
+            // Close activities opened for importing keys and go to the list of keys
+            Intent intent = new Intent(this, MainActivity.class);
+            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+            startActivity(intent);
+        } else {
+            result.createNotify(ImportKeysActivity.this)
+                    .show((ViewGroup) findViewById(R.id.import_snackbar));
+        }
     }
 
     // methods from CryptoOperationHelper.Callback
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
@@ -17,29 +17,46 @@
 
 package org.sufficientlysecure.keychain.ui;
 
+import android.Manifest;
 import android.app.Activity;
+import android.content.ContentResolver;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
+import android.support.v4.content.ContextCompat;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.Toast;
 
 import org.sufficientlysecure.keychain.Constants;
 import org.sufficientlysecure.keychain.R;
 import org.sufficientlysecure.keychain.compatibility.ClipboardReflection;
 import org.sufficientlysecure.keychain.pgp.PgpHelper;
+import org.sufficientlysecure.keychain.ui.ImportKeysListFragment.BytesLoaderState;
 import org.sufficientlysecure.keychain.ui.util.Notify;
 import org.sufficientlysecure.keychain.ui.util.Notify.Style;
 import org.sufficientlysecure.keychain.util.FileHelper;
+import org.sufficientlysecure.keychain.util.Log;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 
 public class ImportKeysFileFragment extends Fragment {
     private ImportKeysActivity mImportActivity;
     private View mBrowse;
     private View mClipboardButton;
 
-    public static final int REQUEST_CODE_FILE = 0x00007003;
+    private Uri mCurrentUri;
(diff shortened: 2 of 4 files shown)


runnerway:
The only thing left is that some activity need to close themselves after the operation is completed.
I will address this probably tomorrow. Then it will be ready.

runnerway:
Now it works pretty well.
The user experience is good and all seems in place.

Valodim:
Awesome lookin' PRs. They also aren't trivial to review, please excuse the delay. I'll try to get to them tomorrow :+1:

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
```diff
@@ -104,20 +124,133 @@ public void onAttach(Activity activity) {
...
+            if (br.read(buffer, 0, length) == length) {
+                isEncrypted = new String(buffer).equals(header);
+            }
+        } catch (IOException e) {
+            throw e;
+        } finally {
+            try {
+                br.close();
+            } catch (IOException e) {
+                e.printStackTrace();
```
Valodim: Log.e

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
```diff
@@ -104,20 +124,133 @@ public void onAttach(Activity activity) {
...
+            br = new BufferedReader(new InputStreamReader(is));
+
+            String header = "-----BEGIN PGP MESSAGE-----";
+            int length = header.length();
+            char[] buffer = new char[length];
+            if (br.read(buffer, 0, length) == length) {
+                isEncrypted = new String(buffer).equals(header);
+            }
+        } catch (IOException e) {
+            throw e;
```
Valodim: isn't this a nop? missing a Log.e mayhaps?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
```diff
@@ -104,20 +124,133 @@ public void onAttach(Activity activity) {
...
+            isEncrypted = isEncryptedFile(mCurrentUri);
+        } catch (IOException e) {
+            e.printStackTrace();
+            Notify.create(mImportActivity, R.string.error_bad_data, Style.ERROR).show();
+            return;
+        }
+
+        if (isEncrypted) {
+            Intent intent = new Intent(getActivity(), DecryptActivity.class);
+            intent.setAction(DecryptActivity.ACTION_DECRYPT_FILE_AND_RETURN);
```
Valodim: I think this should be handled directly as part of the decryption operation, no need to roundtrip. That way we get the same behavior with encrypted keys regardless of what activity the input comes from and the activity stack. Just `startActivity`, `finish` and proceed from there.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
runnerway: Yes. Will fix it.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
runnerway: Yes. Will fix it.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
runnerway: Yes. That was my first try, but i spent a lot of time on it with not very good results regarding to user experience.
I thought the roundtrip was better for some reasons.

But i've just had some new ideas, i will try them and report back.
Thank you for the review.

runnerway:
I've fixed the little problems you pointed out and i tried again to do without roundtrip.
Updated PR to new version. (I still have the older PR on my pc if we want to go back to that)

We have mainly 2 problems:
1) We can't "finish" the activity because otherwise we lose the permission to access the Uri
(http://developer.android.com/guide/topics/providers/content-provider-basics.html#Intents)
2) We need to close all the activities opened during the process to import encrypted keys, as soon as the operation is finished.

So i propose to do as in the latest version of this PR.
After we finish importing new keys, go back to MainActivity. (If not specified otherwise)
This would solve problem n.2 and it makes sense to me. 
Probably the user will go back to MainActivity to see the just imported keys anyway. 

I'm a bit tired now. I hope i didn't forget anything important.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -419,6 +419,16 @@ protected void handleResult(ImportKeyResult result) {
             return;
         }
 
+        if (result.isOkNew() || result.isOkUpdated()) {
+            Preferences prefs = Preferences.getPreferences(this);
```
Valodim: extract into method please

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -419,6 +419,16 @@ protected void handleResult(ImportKeyResult result) {
...
+            prefs.setFirstTime(false);
+
+            Intent intent = new Intent(this, MainActivity.class);
+            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+            startActivity(intent);
```
Valodim: should this `finish()`?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysFileFragment.java
```diff
@@ -121,4 +139,102 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+            Intent intent = new Intent(mImportActivity, DecryptActivity.class);
+            intent.setAction(Intent.ACTION_VIEW);
+            intent.setData(mCurrentUri);
+            startActivity(intent);
+        } else {
+            mImportActivity.loadCallback(new BytesLoaderState(null, mCurrentUri));
+        }
+    }
+
+    private boolean isEncryptedFile(Uri uri) throws IOException {
```
Valodim: make static and move to `FileHelper` please

Valodim:
Good point about the uri permissions. Just tested it, look and feel seems good to me :+1:

runnerway:
I will submit an updated PR with your notes fixed later today.
Thank you for the review.

runnerway:
This should be ready for review. :)

dschuermann:
Looks good. @Valodim : merge?

Valodim:
Yup. I tested this before and code looks good now.

PR #1776, opened 2016-03-10
Update Keyserver Sync if Sync Interval Changed

adithyaphilip:
Until now the only way to update sync interval was uninstalling/re-installing the app. Shouldn't have affected any of our release builds since we never changed it.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
@@ -100,6 +100,11 @@ public class KeychainApplication extends Application {
         // Add OpenKeychain account to Android to link contacts with keys and keyserver sync
         createAccountIfNecessary(this);
 
+        if (Preferences.getKeyserverSyncEnabled(this)) {
+            // will update a keyserver sync if the interval has changed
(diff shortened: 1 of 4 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
```diff
@@ -100,12 +100,14 @@ public void onCreate() {
...
 
-        // if first time, enable keyserver and contact sync
+        // if first time, enable contact sync
         if (Preferences.getPreferences(this).isFirstTime()) {
-            KeyserverSyncAdapterService.enableKeyserverSync(this);
             ContactSyncAdapterService.enableContactsSync(this);
         }
 
+        // will set a keyserver sync if one does not exist or the interval has changed
+        KeyserverSyncAdapterService.updateKeyserverSync(this);
```
dschuermann: when I disable the sync, will this change not enable it every time I start the app?

adithyaphilip: Yes, it would, sorry :disappointed:. Fixed in latest push.

dschuermann:
Thanks for the fix @adithyaphilip . Please rebase, I am okay with a merge!

adithyaphilip:
Thank you, rebased and merged.

PR #1777, opened 2016-03-10
Updated HAS_DUPLICATE_USER_ID in KeychainProvider #1745

fjodorver:


Valodim:
Merged. I also added indexes to the columns to speed up access of the correlated subquery, and changed the matching to be case insensitive.

Issue #927, opened 2014-10-02
Certify flag must be set for first added subkey during key creation


Valodim:
See title. Also, deleting the first subkey makes the second key the new master key, not regarding its flags.

AlexFJW:
I'd be working on this if it's currently not taken. :)

AlexFJW:
I'd like to work on this if it's currently not taken :)

AlexFJW:
The approach I will be taking is to disallow the master key from being deleted, placing an 'edit' button in place of the 'delete' button for the master key. The dialog which appears on click will resemble that of new key addition, with restrictions on the options available to suit the master key. The new key will replace the old key after the user decides on the new settings.

PR #1778, opened 2016-03-11
key creation - master key removal fix

AlexFJW:
A fix for [#927](https://github.com/open-keychain/open-keychain/issues/927), with approach outlined on the page.


Valodim:
does this still work in combination with #1796?

Issue #1779, opened 2016-03-12
Text selection bug on backup code insertion screen


AlexFJW:
Selecting and deleting text in the edit-text field will break the current code and result in an out-of-bounds exception. This seems to be a bug with the Masked-Edittext dependency used.

![resized000](https://cloud.githubusercontent.com/assets/11351531/13720151/673737d2-e83c-11e5-82b5-7224a3777441.png) ![resized002](https://cloud.githubusercontent.com/assets/11351531/13720152/673b018c-e83c-11e5-9d48-4abe1aa66582.png)

AlexFJW:
I've highlighted the issue on Masked-Edittext's github page as well.

PR #1780, opened 2016-03-12
Provisional backup-code text selection fix

AlexFJW:
For #[1779](https://github.com/open-keychain/open-keychain/issues/1779),
A provisional fix is made to make selection more difficult. Preventing selection altogether is not possible due to the internal implementation in the Masked-Edittext dependency.

To be updated when Masked-Edittext fixes this bug.


dschuermann:
great workaround, thanks! I will watch the masked edittext library for a fix.

PR #1782, opened 2016-03-14
Set MIME application/pgp-keys when sending a key

fiaxh:


dschuermann:
Not sure if this is a good idea, I think it prevents sharing ascii armored keys over messaging apps that only support a restricted set of mime types.
But maybe that's actually okay...

fiaxh:
It would.
On the other hand, as the ascii armored key is shared via EXTRA_STREAM, it would only be used by chat apps somehow supporting file transfer. Although there exist chat apps which only support sending files which you can actually display within the app (pictures, ...), I do not see a reason why a chat app would explicitly allow text/* MIME types instead of simply allowing \*/\*.

dschuermann:
yup you are right

Issue #1692, opened 2016-01-17
Change password when key is stripped


dschuermann:
Implement functionality to change the password even when the key is stripped.

Valodim:
It is notable that to implement this feature, the partially implemented notation data for pin and other passphrase types should be removed. these parts are obsolete at this point due to #1541

gabm:
ok guys, i have made some changes here:

https://github.com/gabm/open-keychain/tree/PasswordChangeBugfix

I had to disable some tests because they were not compiling, so please dont get confused by that.
Further there is now a new changeKeyUnlock() method in PGPKeyOperation which by now is called by the "old" modifySecretKeyRing. I dont like this because its like a "hack", so if you know how to call it directly from the menu-button press - please tell me...

I tried to use as much of the old code as possible with changing as least as possible because there are some operations and if's whose meaning I dont understand - please read carefully through the proposal and comment..
By now, i did not remove the notation data handling, but that is a very small thing to do.

Please give me some hints on how to proceed...

dschuermann:
First, please make a pull request with "WIP" in it's title to mark it as "work in progress". We can then better discuss your changes and related issues.
Regarding the disabled tests: Which SDK + version are you using?

Valodim:
Better use a new operation class for this, i.e., define a new operation in KeychainService complete with new parcels for input and result.

dschuermann:
@gabm Any updates?

AlexFJW:
I'll give this a try.

dschuermann:
done in https://github.com/open-keychain/open-keychain/pull/1784

Issue #1085, opened 2015-03-03
Edit passphrase of stripped keys


dschuermann:
Allow passphrase edit of keyrings where the master key has been stripped.
Reference: https://github.com/open-keychain/open-keychain/issues/975#issuecomment-76859607

Valodim:
changing a key's passphrase should be an independent operation from other edits, ui-wise.

Diapolo:
Is this an allowed operation and just currently not supported :)? I was also missing the functionality to change a passphrase without having the master key secret available.

dschuermann:
Just not supported currently. We need to implement this.

Tcheutchoua-Steve:
I will like to fix this ticket but I don't seem to really get what the problem is . I have the app running on my phone and have successfully changed my Passphrase. I don't know if that is the bug in question or it is another one.

montimaj:
You can't change the passphrase if the master key is stripped. That's the issue.

Tcheutchoua-Steve:
OK. Thanks. I get it now

Tcheutchoua-Steve:
This is my approach to solve the ticket
when you select the edit icon , it pops out a list asking you to edit the paraphrase or edit other key items
If you select the " edit passphrase" , you are redirected to another view
The new view requests for the old paraphrase before allowing you to change it.
If you select "edit identities or subkeys " , it asks for the current passphrase and if correctly entered , you are brought to the currently existing view but without the passphrase section (that is change passphrase will not be seen  anymore).
	
Please i need to know if my approach is good so as to solve this ticket. 
Thanks

dschuermann:
@Tcheutchoua-Steve For now, make a "Change passphrase" menu item in the overflow menu that opens a dialog to change the passphrase.

Sharad001:
Hi, sir can i know where can i access the stripped keys functionality. so that i can start working on it !!

dschuermann:
1. Create a new "my key"
2. Edit this key
3. click on your master key (the bold first subkey in the list)
4. select strip key

Sharad001:
So , you want "change passphrase" to be removed from the page and appear on the overflow menu right.

Tcheutchoua-Steve:
I just made a commit of what I did and will like to know if that is what you were talking about. It is on https://github.com/Tcheutchoua-Steve/open-keychain/commit/d2b5969945cd0e0e5cf0bfa490f543666f193c8e

Valodim:
[...]

PR #1784, opened 2016-03-15
WIP: Change password when key is stripped #1692

AlexFJW:
Approach:
Find the first unstripped secret key and use it for passphrase verification
All unstripped keys will have their passphrase changed to new passphrase, if possible.
This quite resembles #1085 (which is closed) so I'm a little unsure if I'm going in the right direction, would like some confirmation before I proceed.

Current Progress:
Changing the passphrase of keys seems to work fine. Pin & related data is not yet removed.
Refactoring to combine "modifySecretKeyring" and newly added method, "modifyKeyRingPassword"
may be possible if given the go-ahead.

Extra notes:
Might be a good idea to clear the remember passphrase option /
wipe all remembered passphrases when a passphrase is changed as well.


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/PassphraseChangeOperation.java
```diff
@@ -0,0 +1,102 @@
...
+import org.sufficientlysecure.keychain.pgp.UncachedKeyRing;
+import org.sufficientlysecure.keychain.provider.ProviderHelper;
+import org.sufficientlysecure.keychain.service.PassphraseChangeParcel;
+import org.sufficientlysecure.keychain.service.input.CryptoInputParcel;
+import org.sufficientlysecure.keychain.ui.util.KeyFormattingUtils;
+import org.sufficientlysecure.keychain.util.ProgressScaler;
+
+
+/**
+ * Created by alex on 3/14/16.
```
Valodim: unnecessary comment, please remove

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/PassphraseChangeOperation.java
```diff
@@ -0,0 +1,102 @@
...
+
+
+    public PassphraseChangeOperation(Context context, ProviderHelper providerHelper, Progressable progressable) {
+        super(context, providerHelper, progressable);
+    }
+
+    /**
+     * Finds the first unstripped key & uses that for passphrase verification.
+     *
+     * @param passphraseParcel  primary input to the operation
```
Valodim: unnecessary comments

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1335,7 +1418,12 @@ private static PGPSecretKeyRing applyNewPassphrase(
             }
 
             sKR = PGPSecretKeyRing.insertSecretKey(sKR, sKey);
+            keysModified++;
+        }
 
+        if(keysModified == 0) {
```
Valodim: code style

AlexFJW:
I'm gonna put this on hold for now, a little short of time as of now.

dschuermann:
@AlexFJW Thanks for your work till now. What are the TODOs that are left for this to be merged?

AlexFJW:
I just realized there's a case that I haven't looked into yet:
-Allowing the change of a Passphrase without a master key may result in some trouble when the user tries to import an unstripped version of his key back into open-keychain. Haven't looked at the code regarding backups so I'm not exactly sure what happens.

Remaining TODO:
Refactoring of "modifySecretKeyring".

I'll see if I can find time to get these done this weekend.

AlexFJW:
Importing a secret keyring would wipe the pre-existing one, which leads to the changed password disappearing. Wiping the pre-existing secret keyring does feel like right thing to do though, since we won't be able to tell which the user wants to keep - the import might be a newer backup from some other source.

Maybe a more explicit warning can be issued for cases where keyrings with stripped master keys are replaced upon importing of keys, just so there's less confusion in such cases.

AlexFJW:
Summary of changes:
Allowed changing passphrase even when masterkey is stripped.

Approach: 
Search for the next unstripped key and use that key's passphrase for validation.
Done by making ChangeUnlockParcel standalone from SaveKeyringParcel, adding a new operation to support changing of passphrases and adding a new method into PgpKeyOperation. SaveKeyringParcel still makes use of a ChangeUnlockParcel when making new keyrings.

Extras:
Added a simple warning dialog which shows when importing keys, to explicitly notify the user that imported keys will replace their existing keys if those keys already exist in the database.

dschuermann:
@AlexFJW What's the status of this pull request? Do you still work on it or should we review it?

AlexFJW:
Ah my bad, should have mentioned that it's ready for a review.

dschuermann:
@AlexFJW great, can you rebase on top of current master?

AlexFJW:
sure. i'll work on that

AlexFJW:
@dschuermann Should be ready for a review now.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
```diff
@@ -300,7 +300,9 @@ private static SaveKeyringParcel createDefaultSaveKeyringParcel(CreateKeyActivit
 
             if(createKeyActivity.mPassphrase != null) {
                 saveKeyringParcel.setNewUnlock(new ChangeUnlockParcel(createKeyActivity.mPassphrase));
-            }
+            } else {
+				saveKeyringParcel.setNewUnlock(null);
+			}
```
dschuermann: please use 4 whitespaces instead of tabs in all changes :)

dschuermann:
please use 4 whitespaces instead of tabs in all changes :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateKeyFinalFragment.java
AlexFJW: Sorry about that. The tab snuck in when I was using meld for conflict resolution. I'll be more careful in the future.

dschuermann:
Great work, thanks!
I merged it but revert the warning dialog. It makes only sense for this specific scenario of importing existing secret keys, but was always shown before import. Maybe we can redesign this when working on the GSoC task of redesigning the import activity.

Issue #1661, opened 2016-01-05
Don't expose CertifyActivity


dschuermann:
* remove exposing of CertifyActivity by removing intent filter on openpgp4fpr.
* this should only be executed using OpenKeychain

Valodim:
Do we do this? I think openpgp4fpr is linked to ImportActivity

see https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/AndroidManifest.xml#L515

Valodim:
yeah ImportKeysProxyActivity goes directly to certification. I'll change it.

Valodim:
Hm. I changed this to ImportKeysActivity now, but I'm not sure if that's really better behavior. Is showing the certify dialog really a problem?

dschuermann:
The issue I have with this that a Barcode Scanner is a non-trusted third-party app, which could change the fingerprint on it's way from the paper with the QR Code to OpenKeychain. That's why I would propose not encouraging direct certification based on this.

Valodim:
Hm. Maybe we could check if the barcode scanner app is a well-known one? Not accepting scanned qr codes from barcode scanner apps with millions of installs feels like treading on paranoid terrain, and this *will* break the workflow of people who are used to, well, scanning codes using their barcode scanner app.

dschuermann:
If it's a targeted attack, it will show the expected information, but is a different key.

Checking if it's a well known scanner app means we must check the apk certificate, not sure if we want to introduce this complexity. Most QR Code scanner apps on Google Play have the same intent filter as "Barcode Scanner" btw.

In general, I think it's a bad security design to expose certification UI without requiring the user to go into OpenKeychain first. It would be only okay to expose this registered API apps. Currently any app can start an intent with "openpgp4fpr" to start certification process, this could also be done from the background via FLAG_NEW_TASK to interrupt an on-going certification btw...

Valodim:
Maybe display an extra activity "You scanned a fingerprint with App X, please scan with openkeychain directly to be safe (click here)"?

dschuermann:
Great idea, yep!

Valodim:
Sounds like a straightforward implementation, and with the whole android intent handling experience to boot :+1:

LuboViluda:
Hi,
[...]

PR #1795, opened 2016-03-20
#1661 - Don't expose CertifyActivity

LuboViluda:
*Add new activity - RedirectImportKeysActivity, which prompts user to
use Openkeychain directly
*Add intent filter for the activity
*Change intent filter for CertifyActivity


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,60 @@
+/*
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
Valodim: we are GPLv3, please change the license so we can merge this

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setFullScreenDialogClose(Activity.RESULT_CANCELED, true);
+        final Intent intent = new Intent(this, org.sufficientlysecure.keychain.ui.ImportKeysActivity.class);
+
+        new AlertDialog.Builder(this)
+                .setTitle("Import key attempt")
+                .setMessage("You scanned a fingerprint with another app, please scan with Openkeychain directly to be safe" )
```
Valodim: text should be in string resources so it can be translated

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,60 @@
...
+                .setPositiveButton(android.R.string.yes, new DialogInterface.OnClickListener() {
```
Valodim: buttons should indicate their actual action, rather than refer to the dialog content. in this case, that would be something like, "Cancel" and "Scan Again"

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,60 @@
...
+                    public void onClick(DialogInterface dialog, int which) {
+                        // intent directly to ImportKeyChain activity
+                        startActivity(intent);
```
Valodim: use a method like `startScanActivity` for the logic here

OpenKeychain/src/main/res/values/dimens.xml
```diff
@@ -5,4 +5,7 @@
     <dimen name="big_toolbar">120dp</dimen>
     <dimen name="huge_toolbar">222dp</dimen>
     <dimen name="minus_statusbar_height">0dp</dimen>
+    <!-- Default screen margins, per the Android Design guidelines. -->
+    <dimen name="activity_horizontal_margin">16dp</dimen>
```
Valodim: is this actually used anywhere?

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -1740,4 +1740,6 @@
         <item>50 EUR</item>
         <item>100 EUR</item>
     </string-array>
+    <string name="title_activity_redirect_key">RedirectKeyActivity</string>
```
Valodim: should have a human readable title

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -1740,4 +1740,6 @@
...
+    <string name="action_settings">Settings</string>
```
Valodim: where does this come from?

OpenKeychain/src/main/res/layout/redirect_import_keys_activity.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
Valodim: this entire layout isn't really used, is it? can we just use a transparent activity to show the dialog instead?

Valodim:
the approach looks good. added some nitpicks about the implementation though :+1:

LuboViluda:
I implemented required changes. I try to implemented transparent window in more elagent way (through new style in styles.xml, however there is [bug](https://code.google.com/p/android/issues/detail?id=2863) in Android which makes it harder. Workarounds doesn't work with minimum API 15). I hope, substitute solution is fine as well.

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -711,6 +711,16 @@
...
 
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".ui.MainActivity" />
+        </activity>
+        <activity
+            android:name=".ui.RedirectImportKeysActivity"
+            android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
+            android:label="@string/title_activity_redirect_key"
+            android:theme="@android:style/Theme.Translucent">
```
dschuermann: Why don't you use ``@style/Theme.Keychain.Transparent`` ?

LuboViluda: Yeah, it is better approach definitely, thanks

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,64 @@
...
+
+        startScanActivity();
+    }
+
+    @Override
+    protected void initLayout() {
+
+    }
+
+    private void startScanActivity() {
```
Valodim: this method seems inappropriately named

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,64 @@
...
+        final Intent intent = new Intent(this, org.sufficientlysecure.keychain.ui.ImportKeysActivity.class);
```
Valodim: why the fully qualified class name? also, shouldn't this start the scanner rather than the import dialog?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.sufficientlysecure.keychain.R;
+import org.sufficientlysecure.keychain.ui.base.BaseActivity;
+
+public class RedirectImportKeysActivity extends BaseActivity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.redirect_import_keys_activity);
```
dschuermann: do we need this when the activity is transparent?

OpenKeychain/src/main/res/values-w820dp/dimens.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
```
dschuermann: not used anywhere -> remove

OpenKeychain/src/main/res/values/dimens.xml
```diff
@@ -5,4 +5,5 @@
     <dimen name="big_toolbar">120dp</dimen>
     <dimen name="huge_toolbar">222dp</dimen>
     <dimen name="minus_statusbar_height">0dp</dimen>
+    <dimen name="activity_vertical_margin">16dp</dimen>
```
dschuermann: not used anywhere -> remove

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -1740,4 +1740,9 @@
         <item>50 EUR</item>
         <item>100 EUR</item>
     </string-array>
+    <string name="redirect_import_key_title">"Import key attempt"</string>
+    <string name="redirect_import_key_message">"You scanned a fingerprint with another app, please scan with Openkeychain directly to be safe"</string>
+    <string name="redirect_import_key_yes">Scan again</string>
+    <string name="redirect_import_key_no">Close</string>
+    <string name="title_activity_redirect_key">Key import redirection</string>
```
dschuermann: please check all strings and put ``"`` around the values!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/RedirectImportKeysActivity.java
LuboViluda: thanks for correction, android.view.Windows was unnecesary, other code is still needed.

dschuermann: I mean ``setContentView(R.layout.redirect_import_keys_activity);``. Is this required? The activity is transparent...

dschuermann:
@LuboViluda My last comment still stands. Do we need ``setContentView`` and the layout?

dschuermann:
Thanks for your work!

Issue #1600, opened 2015-11-17
Limit create key options / redesign subkey creation dialog


JasperWeiss:
Openkeychain offers the option to use an 8192 bit RSA key. I figured I might aswell use that. But it seems people reccommend not to use it because it adds extreme ammounts of overhead and is considered overkill.
Is it? 
And if it is, why does Openkeychain offer it as an option?

Valodim:
Some people do love this feature, and create 16k RSA keys and feel really good about their feeling of increased security.

I agree: We should remove it.

JasperWeiss:
I have to admit I barely see any performance hit. Should I revert to 4096? I haven't used/uploaded it yet because I'm still experimenting with it. I'm stil not sure whether I should use RSA, 4k, 8k, Elliptic curve etc. I know RSA 4k is recommended but I was curious if anything higher/different might be recommended a few years in the future.

Valodim:
even rsa 4096 has a fair margin of healthy overkill to it. check out http://keylength.com

dschuermann:
@Valodim Before we remove this feature: After edit key refactorings, this will be pretty hidden anyway.

If we start removing support for RSA > 4096, it's pretty hard to argue why not simply remove every key creation option besides RSA 4096, ECDSA P-256, and ECDH P-256. All other options are questionable.

Valodim:
RSA2048 is used on security tokens, but yes everything else is questionable

dschuermann:
Still RSA 2048 must not be available via UI :D so even that one...
I am not sure if we should go down this road. As always, I like to have a balance between flexibility and usability. Maybe we can simply move that even more into the advanced screens.

End-to-End is always the extreme example for me, where there aren't even advanced screens to add RSA keys (at least in the last version I saw). Maybe that's going too far.

JasperWeiss:
I feel most users would go with the default configuration. People who specifically change the configuration to anything else probably know what they're doing. Isn't free-open-source software about giving users flexibillity and control? Perhaps we could hide the other alhorithms/key sizes by default and show an option in the configuration menu like "show advanced options"

dschuermann:
If no one has an objection, I would close this issue. Key creation is now only available in the advanced screen.

related:
https://github.com/open-keychain/open-keychain/issues/1657

Valodim:
[...]

PR #1796, opened 2016-03-21
Refer to issue #1600 Limit create key options

AlexFJW:
Also changed key type of default keys in key creation wizard to RSA, 3072 bit.

Implementation details:
Using the default spinner results in the dropdown focusing on the default item (RSA, 3072) in the list
when selected. This obscures RSA 2048 on smaller screens (tested on a 800x480), which is not very intuitive in my opinion. As such, a custom spinner is used to ensure that the focus is always set on the first item in the list instead. A hack was used to achieve this. (refer to FocusFirstItemSpinner.java)


AlexFJW:
Screens of the actual implementation:
dropdown main

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/AddSubkeyDialogFragment.java
```diff
@@ -269,37 +211,65 @@ public void onStart() {
...
+                    // set algorithm
+                    switch (keyType) {
+                        case RSA_2048:
+                        case RSA_3072:
+                        case RSA_4096:
+                            algorithm = Algorithm.RSA;
+                            break;
+                        case ECC_P256:
+                        case ECC_P521:
+                            if(mUsageEncrypt.isChecked())
```
dschuermann: always use brackets for if else

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/AddSubkeyDialogFragment.java
```diff
@@ -269,37 +211,65 @@ public void onStart() {
...
-                    if (algorithm == Algorithm.ECDH || algorithm == Algorithm.ECDSA) {
-                        curve = ((Choice<Curve>) mCurveSpinner.getSelectedItem()).getId();
-                        // Otherwise, get a keysize
-                    } else {
-                        keySize = getProperKeyLength(algorithm, getSelectedKeyLength());
+                    Algorithm algorithm = null;
+
+                    // set keysize & curve, for RSA & ECC respectively
+                    switch (keyType) {
+                        case RSA_2048:
```
dschuermann: use brackets for cases

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/AddSubkeyDialogFragment.java
```diff
@@ -332,206 +302,29 @@ public void onClick(View v) {
...
-                break;
-            }
-        }
-        keySizeAdapter.notifyDataSetChanged();
+            TextView text1 = (TextView) convertView.findViewById(android.R.id.text1);
+            TextView text2 = (TextView) convertView.findViewById(android.R.id.text2);
 
-    }
+            text1.setText(c.getName());
+            text2.setText(" - " + c.getDescription());
```
dschuermann: use description only, remove " - "

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -165,6 +165,7 @@
     <string name="label_keyservers">"Select OpenPGP keyservers"</string>
     <string name="label_key_id">"Key ID"</string>
     <string name="label_key_created">"Key created %s"</string>
+    <string name="label_key_type_and_size">"Type &amp; Size"</string>
```
dschuermann: just call it "Type"

dschuermann:
I like the selection of type so far.

I am still not 100% sold to the way we now handle key usage, thus I would like to re-open the discussion about usage: If we re-introduce the possibility of transferring OpenPGP keys to a security token such as YubiKey, it makes sense to allow the "authenticate" flag somehow. If we don't re-introduce this, the description for RSA 2048 makes no sense.
Also, do I see it right that it's currently not possible to add a master key that is also capable of signing?

AlexFJW:
I have to agree that your point about security tokens didn't quite cross my mind. I don't actually use one myself. Perhaps it'll be good to remove the part about security tokens for now?

It is possible to add a master key that is capable of signing actually. Did I miss something? None refers to just adding a certify flag. Here's the screens for the master key creation:
ecc rsa

dschuermann:
Ah great, looks good. I simply haven't really tested the pr just looked at the code and was wondering if certify+sign is possible :)

I would agree to remove the note about YubiKeys for now.

AlexFJW:
Sure :) Thanks for the rapid code review.

Valodim:
The "(!)" looks a little out of place in the descriptions - we aren't limited to plaintext here. Maybe use a bold font for the warning text instead?

I have a couple of nitpicks for key creation:
- preselect "None" when the key to add is the master key
- show an error and don't allow saving when there is no certify key
- for better user guidance, don't allow deletion of the master key when there are still subkeys

It would be awesome if you could get those in as well. If you prefer we can also merge this PR now and leave that for another one :+1:

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/AddSubkeyDialogFragment.java
```diff
@@ -332,206 +305,29 @@ public void onClick(View v) {
...
-                mFlagSign.setEnabled(false);
-                mFlagEncrypt.setChecked(true);
-                mFlagEncrypt.setEnabled(true);
-                mFlagAuthenticate.setChecked(false);
-                mFlagAuthenticate.setEnabled(false);
-                break;
-            }
-        }
-        keySizeAdapter.notifyDataSetChanged();
+            TextView text1 = (TextView) convertView.findViewById(android.R.id.text1);
```
Valodim: more descriptive names here please

Valodim:
there are a bunch of android studio warnings, particularly in AddSubkeyDialogFragment. please get rid of those, or suppress + annotate a reason.

OpenKeychain/src/main/res/layout/two_line_spinner_dropdown_item.xml
```diff
@@ -0,0 +1,32 @@
...
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    android:paddingLeft="10dp"
+    android:paddingTop="6dp"
+    android:paddingRight="10dp"
+    android:paddingBottom="6dp"
+    android:background="#FFFFFFFF">
```
Valodim: is there a reason you didn't use selectableIconBackground here to get the lollipop ripple effect?

Valodim:
The "None" option for master keys makes it sound like the key is useless then. Maybe "None (subkey binding only)" or "None (master key only)" or "None (bare master key)"

AlexFJW:
My bad. I'll fix the points you highlighted in a bit. 

As for:
"    preselect "None" when the key to add is the master key
     show an error and don't allow saving when there is no certify key
     for better user guidance, don't allow deletion of the master key when there are still subkeys   "

I'm not sure if the stuff I wrote at 1778 still works so I'll close that PR and bring in the changes over here instead.

OpenKeychain/src/main/res/layout/two_line_spinner_dropdown_item.xml
AlexFJW: I simply disliked the gray background which appears. Didn't notice that doing so cancels out the ripple effect. I'll see if I can achieve both.

AlexFJW:
Here's my proposed implementation for the bit relating to master key:
Rather than not allowing the deletion of the master key, we could place an change button and allow changing out the master key instead. A missing delete button might confuse users, while one which gives a toast mentioning to delete other keys first might prompt new users to try it out.

The image I have in mind is..

AlexFJW:
Here's my proposed implementation for the bit relating the master key:
Rather than not allowing the deletion of the master key, we could swap in a change button instead and allow changing out the master key. A missing delete button might confuse users, while one which gives a toast mentioning to delete other keys first might really prompt new users to do just that and change the configuration of other keys. 

The icon I have in mind is 
rotate-3d , from materialdesignicons.com.

I'm for 'changing out' rather than 'editing' of the master key as quite a lot of code is required for the latter, with few benefits.

AlexFJW:
Some new screens: 
I decided to use underlines for the warnings as bold looked a little harsh. 
Went for (subkey binding only) to accompany none. It feels like it gives some explanation for the need of the master key, which is nice. 

changemasterkey dropdown mainscreen

Valodim:
lookin' good. the experimental warning with underline might look better on the next line, separated by a `\n` rather than a `,`? other than that we can merge this from my side :+1:

AlexFJW:
Sure. I'll see if \n works better and squash the commits in a bit.

AlexFJW:
Here's how it looks with line breaks on a 800x480. It does look a bit better than using a comma so I'll be going with it. Have squashed the commits, no other changes made. 

newline

Valodim:
Merged! Thanks :+1:

Issue #1792, opened 2016-03-16
Update NetCipher to get rid of "Tor appears not to be running"-message


vanitasvitae:
Hi!
On AOSP>=Android N and also on CopperheadOS the flag hidepid=2 is set, which prevents apps from detecting, whether Tor is running or not. The GuardianProject fixed this in the latest version of NetCipher. Could you please update NetCipher, to make Key Synchronization via tor possible again?

vanitasvitae:
Edit: See this issue on CopperheadOS:
https://github.com/copperhead/bugtracker/issues/197#issuecomment-196945201

dschuermann:
I tagged this as help-wanted. Should be fixed, yes!

Litun:
@dschuermann Could u review my pull request?

vanitasvitae:
I'll try to implement this.

vanitasvitae:
@dschuerman I think I fixed this. Can you review my pull request?

dschuermann:
done in https://github.com/open-keychain/open-keychain/pull/1865

PR #1797, opened 2016-03-22
Merged the latest changes from NetCipher

Litun:
Issue #1792


Valodim:
looks like a PR for @dschuermann

dschuermann:
Hmmm, I don't really see how this fixes #1792. According to the commit https://github.com/guardianproject/NetCipher/commit/26e6bf76fd2aee484864ead03b8b73f15089474b we need to use the STATUS Intent to check if Orbot is running or not.

Also I don't like to have any code that uses "su" inside OpenKeychain, please don't blindly backport everything from Netcipher.

dschuermann:
The issue has not been fixed. Please open a new pull request with the desired changes.

Issue #1707, opened 2016-01-30
Refresh of my own public key fails


jovandeginste:
I'm unsure if there is something wrong with my keypair, or with openkeychain. I search the issues, but couldn't find a solution so far.

When I refresh my own key (or all keys) I get an error (Dutch):

```
[START] Bezig met importeren van sleutel
 [DEBUG] Sleutelserver hkps://hkps.pool.sks-keyservers.net wordt gebruikt
  [INFO] Bezig met ophalen van sleutelserver: 0xe40adc64193e10fd
 [ERROR] Importeren mislukt!
```

However, I can use it for everything else, I can trust people, encrypt data, etc. How can I see more details about the error?

adithyaphilip:
Stacktrace on downloading `0xe40adc64193e10fd`
```
Encountered bad key on import!
org.sufficientlysecure.keychain.pgp.exception.PgpGeneralException: Expected single keyring in stream, found at least two
at org.sufficientlysecure.keychain.pgp.UncachedKeyRing.decodeFromData(UncachedKeyRing.java:153)
at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:209)
at org.sufficientlysecure.keychain.operations.ImportOperation.access$000(ImportOperation.java:84)
at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:490)
at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:478)
at java.util.concurrent.FutureTask.run(FutureTask.java:237)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
at java.util.concurrent.FutureTask.run(FutureTask.java:237)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
at java.lang.Thread.run(Thread.java:818)
```

jovandeginste:
Any context? Is my key the problem, or an issue with the code?

Valodim:
wow uh, this is a really weird case. can you tell us what software you used to create or ever modify this key? gpg, keybase, openkeychain, etc?

jovandeginste:
[...]

PR #1798, opened 2016-03-22
Import keys when same key available twice on keyserver (#1707)

roryflynn:
#1707 Allow for import of keys when the same key is available twice on the keyserver.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
@@ -206,7 +206,7 @@ public class ImportOperation extends BaseOperation<ImportKeyringParcel> {
                                 log.add(LogType.MSG_IMPORT_FETCH_KEYSERVER, 2, entry.mKeyIdHex);
                                 data = keyServer.get(entry.mKeyIdHex).getBytes();
                             }
-                            key = UncachedKeyRing.decodeFromData(data);
+                            key = UncachedKeyRing.decodeFromData(data, entry.mExpectedFingerprint, entry.mKeyIdHex);
                             if (key != null) {
                                 log.add(LogType.MSG_IMPORT_FETCH_KEYSERVER_OK, 3);
                             } else {
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
@@ -176,7 +176,12 @@ public class ImportOperation extends BaseOperation<ImportKeyringParcel> {
 
                 // If there is already byte data, use that
                 if (entry.mBytes != null) {
-                    key = UncachedKeyRing.decodeFromData(entry.mBytes);
+                    if (entry.mExpectedFingerprint == null && entry.mKeyIdHex == null) {
+                        key = UncachedKeyRing.decodeFromData(entry.mBytes);
+                    } else {
+                        key = UncachedKeyRing.decodeFromDataFindExpected(entry.mBytes,
+                                entry.mExpectedFingerprint, entry.mKeyIdHex);
+                    }
                 }
                 // Otherwise, we need to fetch the data from a server first
                 else {
@@ -206,7 +211,8 @@ public class ImportOperation extends BaseOperation<ImportKeyringParcel> {
                                 log.add(LogType.MSG_IMPORT_FETCH_KEYSERVER, 2, entry.mKeyIdHex);
                                 data = keyServer.get(entry.mKeyIdHex).getBytes();
                             }
-                            key = UncachedKeyRing.decodeFromData(data, entry.mExpectedFingerprint, entry.mKeyIdHex);
+                            key = UncachedKeyRing.decodeFromDataFindExpected(data,
+                                    entry.mExpectedFingerprint, entry.mKeyIdHex);
                             if (key != null) {
                                 log.add(LogType.MSG_IMPORT_FETCH_KEYSERVER_OK, 3);
                             } else {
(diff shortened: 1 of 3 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/UncachedKeyRing.java
```diff
@@ -157,6 +159,50 @@ public static UncachedKeyRing decodeFromData(byte[] data)
...
+            Iterator<UncachedPublicKey> publicKeys = ring.getPublicKeys();
+
+            while (publicKeys.hasNext()) {
+                UncachedPublicKey publicKey = publicKeys.next();
+
+                boolean publicKeyMatches = (
+                        KeyFormattingUtils.convertFingerprintToHex(publicKey.getFingerprint())
+                                .equalsIgnoreCase(expectedFingerprint)
+                        || TextUtils.equals(Long.toString(publicKey.getKeyId()), expectedKeyId));
+                if (publicKeyMatches) {
```
Valodim: you can reduce nesting depth in cases like this with something like `if (!publicKeyMatches) { continue; }`, which also makes it obvious that there is no else case

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/UncachedKeyRing.java
```diff
@@ -157,6 +159,50 @@ public static UncachedKeyRing decodeFromData(byte[] data)
 
     }
 
+    public static UncachedKeyRing decodeFromData(
```
Valodim: name could be more descriptive, i.e. `decodeFromDataFindExpected` or something like that

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/UncachedKeyRing.java
```diff
@@ -157,6 +159,50 @@ public static UncachedKeyRing decodeFromData(byte[] data)
...
+            UncachedKeyRing ring = parsed.next();
+            Iterator<UncachedPublicKey> publicKeys = ring.getPublicKeys();
+
+            while (publicKeys.hasNext()) {
+                UncachedPublicKey publicKey = publicKeys.next();
+
+                boolean publicKeyMatches = (
+                        KeyFormattingUtils.convertFingerprintToHex(publicKey.getFingerprint())
+                                .equalsIgnoreCase(expectedFingerprint)
+                        || TextUtils.equals(Long.toString(publicKey.getKeyId()), expectedKeyId));
```
Valodim: seems better to turn expectedKeyId into a Long earlier, then compare if non-null

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/ImportOperation.java
```diff
@@ -206,7 +206,7 @@ private ImportKeyResult serialKeyRingImport(Iterator<ParcelableKeyRing> entries,
                                 log.add(LogType.MSG_IMPORT_FETCH_KEYSERVER, 2, entry.mKeyIdHex);
                                 data = keyServer.get(entry.mKeyIdHex).getBytes();
                             }
-                            key = UncachedKeyRing.decodeFromData(data);
+                            key = UncachedKeyRing.decodeFromData(data, entry.mExpectedFingerprint, entry.mKeyIdHex);
```
Valodim: what about the other calls to decodeFromData?

roryflynn:
@Valodim should be ready for another review!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/UncachedKeyRing.java
```diff
@@ -163,6 +163,54 @@ public static UncachedKeyRing decodeFromData(byte[] data)
...
+                UncachedPublicKey publicKey = publicKeys.next();
+
+                boolean fingerprintMatches = KeyFormattingUtils.convertFingerprintToHex(publicKey.getFingerprint())
+                        .equalsIgnoreCase(expectedFingerprint);
+                boolean keyIdMatches = (expectedKeyIdLong != null
+                        && expectedKeyIdLong.longValue() == publicKey.getKeyId());
+
+                boolean publicKeyMatches = ( fingerprintMatches || keyIdMatches );
+
+                if ( ! publicKeyMatches) { continue; }
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/util/KeyFormattingUtils.java
```diff
@@ -271,6 +271,25 @@ private static String convertKeyIdToHex32bit(long keyId) {
         return hexString;
     }
 
+    /**
+     * Convert key id from hex string to long
+     * @param keyIdHex - 64 bit hex string representation of keyId, including "0x" prefix
+     * @return keyId as long
+     */
+    public static Long convertKeyIdHexToLong(String keyIdHex) {
+        if (keyIdHex == null) { return null;}
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/util/KeyFormattingUtils.java
```diff
@@ -271,6 +271,25 @@ private static String convertKeyIdToHex32bit(long keyId) {
...
+
+        //we can't convert directly from hex
```
Valodim: why not?

roryflynn: It appears you can't parse a long from a two's complement encoded hex string using Long.parseLong or Long.decode http://stackoverflow.com/questions/1410168/how-to-parse-negative-long-in-hex-in-java

I can make the comment a bit more verbose if you want?

Valodim: sounds fine. including the link might also be a good idea

roryflynn:
@Valodim Code style issues + comments should be fixed now

Valodim:
Would you add a positive and a negative unit test for the behavior you introduced? That would be awesome :+1:

dschuermann:
@roryflynn Still working on this PR?

roryflynn:
@dschuermann Apologies, not at the minute due to exams. Others are welcome to finish it off by creating the test cases though. If not, I'll try to finish it in about two weeks time.

dschuermann:
@roryflynn Take your time with the exams. We appreciate if you get back to us afterwards :)

Valodim:
closing as stale

PR #1804, opened 2016-03-29
tint linked contact picture in toolbar #1298

### LINKED ISSUE #1298: tint linked contact picture in toolbar
the linked contact picture as displayed in the toolbar should be tinted green. otherwise, buttons and text on it may be hidden if the picture doesn't by chance offer enough contrast

vhazali:
Fixes #1298

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyActivity.java
@@ -33,6 +33,7 @@ import android.content.Intent;
 import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.Color;
+import android.graphics.PorterDuff;
 import android.net.Uri;
 import android.nfc.NfcAdapter;
 import android.os.AsyncTask;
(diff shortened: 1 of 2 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyActivity.java
```diff
@@ -923,6 +924,7 @@ protected void onPostExecute(Bitmap photo) {
                                     }
 
                                     mPhoto.setImageBitmap(photo);
+                                    mPhoto.setColorFilter(Color.argb(30, 120, 166, 69));
```
Valodim: The color should be in resources, not hardcoded. Also, please explicitly set the blending mode :+1:

vhazali:
I'll need some more time to get this to work. Somehow I'm not able to let the correct colour show when I use the constant values in resources =\

Valodim:
setColorFilter takes a color value, not a resource id. Both are int though, so you have to be careful with the types. You probably need something like `getResources().getColor()` there.

vhazali:
Thanks! That solved it. I didn't know that `R.color.toolbar` will give the resource id instead of the value of the resource

Valodim:
thanks!

PR #1807, opened 2016-03-30
Fix Crash on encrypt using revoked key

007durgesh219:
To Reproduce
1. Revoke a key
2. Select it from the main screen 
3. Try encrypt file using menu icon in the action bar 

Fix:
Hide encrypt icon in the action bar if any selected key is revoked

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
@@ -254,6 +254,11 @@ public class KeyListFragment extends LoaderFragment
                 String keysSelected = getResources().getQuantityString(
                         R.plurals.key_list_selected_keys, count, count);
                 mode.setTitle(keysSelected);
+                if (mAdapter.isAnyRevokedKeySelected()) {
+                    mode.getMenu().getItem(0).setVisible(false);
+                } else {
+                    mode.getMenu().getItem(0).setVisible(true);
+                }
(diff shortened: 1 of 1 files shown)


007durgesh219:
Any feedbacks on this??

Valodim:
Nice catch for this bug. But this fix only solves the symptom, not the underlying problem.

```
 java.lang.IllegalStateException: no encryption methods specified
 at org.bouncycastle.openpgp.PGPEncryptedDataGenerator.open(PGPEncryptedDataGenerator.java:178)
 at org.bouncycastle.openpgp.PGPEncryptedDataGenerator.open(PGPEncryptedDataGenerator.java:333)
 at org.sufficientlysecure.keychain.pgp.PgpSignEncryptOperation.execute(PgpSignEncryptOperation.java:340)
 at org.sufficientlysecure.keychain.operations.SignEncryptOperation.execute(SignEncryptOperation.java:153)
 at org.sufficientlysecure.keychain.operations.SignEncryptOperation.execute(SignEncryptOperation.java:61)
 at org.sufficientlysecure.keychain.service.KeychainService$1.run(KeychainService.java:146)
```

This should be handled in PgpSignEncryptOperation, and display an error message. For the extra mile, it could also highlight the problematic recipients.

007durgesh219:
OK, But if key is revoked I see that there is no other option in entire app to use it(encript any file or text) except by selecting from home screen. From where we can only use it to encrypt files.  Why is that so??. That's why I thought we should not allow to encrypt it from home screen also.

PR #1736, opened 2016-03-01
Warning when stripping master key.

tdjogi010:

PR #1808, opened 2016-03-31
Warn when stripping master key #1554

### LINKED ISSUE #1554: warn when stripping master key
Stripping master key renders tge key chain useless, cant revoke the keys or even modify anything for that keypair on openkeychain. So users should be warned when stripping master key. Warning may give option to revoke the keypair before the action, like the app does when deleting keypairs. creating another key if it make any sense :p

vhazali:
Based off @tdjogi010's [work](https://github.com/open-keychain/open-keychain/pull/1736), improved on the solution as mentioned in pull request 1736.

Fixes #1554

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyDialogFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyDialogFragment.java
@@ -75,22 +75,7 @@ public class EditSubkeyDialogFragment extends DialogFragment {
                         sendMessageToHandler(MESSAGE_REVOKE, null);
                         break;
                     case 2:
-                        CustomAlertDialogBuilder stripAlertDialog = new CustomAlertDialogBuilder(getActivity());
-                        stripAlertDialog.setTitle(getResources().getString(R.string.title_alert_strip)).
-                        setMessage(R.string.alert_strip).setCancelable(true);
-                        stripAlertDialog.setPositiveButton(R.string.strip, new DialogInterface.OnClickListener() {
-                            @Override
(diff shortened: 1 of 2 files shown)


OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -332,6 +334,7 @@
     <string name="public_key_deletetion_confirmation">"Delete key '%s'?"</string>
     <string name="also_export_secret_keys">"Also export secret keys"</string>
     <string name="reinstall_openkeychain">"You encountered a known bug with Android. Please reinstall OpenKeychain if you want to link your contacts with keys."</string>
+    <string name="alert_strip">"Stripping this subkey will render it useless."</string>
```
Valodim: "useless" is not a good word here, maybe "make it unusable on this device!" would be better? adding the "on this device" part, because that's the whole point of it

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/EditSubkeyDialogFragment.java
```diff
@@ -95,6 +99,25 @@ public void onClick(DialogInterface dialog, int id) {
         return builder.show();
     }
 
+    private void showAlertDialog() {
+        CustomAlertDialogBuilder stripAlertDialog = new CustomAlertDialogBuilder(getActivity());
+        stripAlertDialog.setTitle(getResources().getString(R.string.title_alert_strip)).
```
Valodim: doesn't `setTitle` have a version which uses the resource id directly, just like `setMessage`?

Valodim:
looks good, just minor nitpicks!

vhazali:
Got it! I'll make the changes in a bit

Valodim:
Looks good! Thanks for your work :+1:

PR #1812, opened 2016-04-03
Prevent crashing caused by MaskedEditText

AlexFJW:
Refer to #1809
Crashing is caused by mass deletion of text. Two fixes are introduced to prevent this.

1) Prevent mass delete when backspace button is held
SuccessTextWatcher is modified to watch for mass delete, and to emulate a single delete if detected.

2) Prevent mass selection when double tapping
OnTouch method of MaskedEditText is overridden, limiting selection size to 0.

Note that fix 2) is required as allowing mass selection but not allowing mass delete does not make sense in a UX perspective.


Valodim:
see #1815

PR #1813, opened 2016-04-03
Auto-capitalization for backup code insertion

AlexFJW:
InputType.TYPE_TEXT_FLAG_CAP_CHARACTERS does not work for all keyboards*. Adding the InputFilter.AllCaps filter circumvents this problem.

*Fails on Swiftkey


Valodim:
see #1815

Issue #1732, opened 2016-03-01
Use Rfc822Tokenizer instead of Regex?


dschuermann:
I just discovered Rfc822Tokenizer, which could be used to parse User IDs. Maybe this works better than our regex.

This must be tested!

Valodim:
sounds like a simple fix

rohands:
@dschuermann I'll give it a shot

ludwwwig:
@rohands Have you made any progress? I would be interested to help with this!

dschuermann:
@ludwwwig go ahead, rohands comment is old enough.

Valodim:
rfc822tokenizer is [not the most thoroughly tested class](https://android.googlesource.com/platform/cts/+/e86ece2/tests/tests/text/src/android/text/util/cts/Rfc822TokenizerTest.java), but I suppose it'll do.

fiaxh:
Using Rfc822Tokenizer would lead to a few problems. As emails are an obligatory component of the user ID for the Rfc822Tokenizer, some cases of PGP UIDs are not parsed correctly:
*  A user ID of the form "Name (Comment)" would be interpreted as email: Name, comment: Comment.
* "Name" would be parsed as email: Name
* " (Comment)" wouldn't be parsed at all
The first two cases could be solved by pattern-matching the result from Rfc822Tokenizer. But on the other hand, the regex results could be improved by checking some side cases, too. And by doing this "manually", it's more flexibly adaptable to actually occurring PGP UID formats. (See PR)

dschuermann:
@fiaxh thanks for your tests and your pull request.

PR #1816, opened 2016-04-04
Correctly handle keys only having an email address

### LINKED ISSUE #1699: Does not auto-select keys without an empty name field
I have a few keys, which only have an email field, but no name field. This is mostly for anonymity and is e.g. also a [requirement of Posteo](https://posteo.de/en/help/policies-for-public-keys#names) for publishing the public key in their [OpenPGPKey](https://openpgpkey.info/) DNS.

But when sending an encrypted mail to a mail address whose public key has this format OpenKeyChain always asks for the key and does not select it automatically. Instead I have to choose it manually every time.

fiaxh:
When a key has a user id only consisting of an email, that email is wrongly taken as name.
fixes #1699

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
@@ -54,7 +54,7 @@ import java.io.IOException;
  */
 public class KeychainDatabase extends SQLiteOpenHelper {
     private static final String DATABASE_NAME = "openkeychain.db";
-    private static final int DATABASE_VERSION = 16;
+    private static final int DATABASE_VERSION = 17;
     static Boolean apgHack = false;
     private Context mContext;
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
@@ -456,12 +456,12 @@ public class KeychainProvider extends ContentProvider {
                             if (i != 0) {
                                 emailWhere += " OR ";
                             }
-                            emailWhere += "tmp." + UserPackets.USER_ID + " LIKE ";
-                            // match '*<email>', so it has to be at the *end* of the user id
                             if (match == KEY_RINGS_FIND_BY_EMAIL) {
-                                emailWhere += DatabaseUtils.sqlEscapeString("%<" + chunks[i] + ">");
+                                emailWhere += "tmp." + UserPackets.EMAIL + " LIKE "
+                                        + DatabaseUtils.sqlEscapeString(chunks[i]);
                             } else {
-                                emailWhere += DatabaseUtils.sqlEscapeString("%" + chunks[i] + "%");
+                                emailWhere += "tmp." + UserPackets.USER_ID + " LIKE "
+                                        + DatabaseUtils.sqlEscapeString("%" + chunks[i] + "%");
                             }
                             gotCondition = true;
                         }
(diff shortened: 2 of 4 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
```diff
@@ -456,12 +456,12 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
                             if (i != 0) {
                                 emailWhere += " OR ";
                             }
-                            emailWhere += "tmp." + UserPackets.USER_ID + " LIKE ";
-                            // match '*<email>', so it has to be at the *end* of the user id
                             if (match == KEY_RINGS_FIND_BY_EMAIL) {
-                                emailWhere += DatabaseUtils.sqlEscapeString("%<" + chunks[i] + ">");
+                                emailWhere += "tmp." + UserPackets.EMAIL + " = "
+                                        + DatabaseUtils.sqlEscapeString(chunks[i]);
```
Valodim: this is an exact match, not a substring search like before

Valodim:
About the address parsing: See #1732. To move forward with an implementation, we definitely need unit tests to make sure the address formats we expect in there are all correctly handled. I don't think we should change anything about this implementation without that.

As for the matching in KeychainProvider, we should indeed use the email field we now have.

dschuermann:
@fiaxh is this ready for review again?

fiaxh:
@dschuermann Yes it is. The code continues to use the regex for uid parsing (argumentation see #1732) and adds some more unit tests.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
```diff
@@ -54,7 +54,7 @@
  */
 public class KeychainDatabase extends SQLiteOpenHelper {
     private static final String DATABASE_NAME = "openkeychain.db";
-    private static final int DATABASE_VERSION = 16;
+    private static final int DATABASE_VERSION = 17;
```
dschuermann: if the db schema does not change, we don't need to increase the db version

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/pgp/KeyRingTest.java
```diff
@@ -30,28 +30,63 @@
 @Config(constants = WorkaroundBuildConfig.class, sdk = 21, manifest = "src/main/AndroidManifest.xml")
 public class KeyRingTest {
```
Valodim: The UserId-related tests are worth their own class

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/pgp/KeyRingTest.java
```diff
@@ -30,28 +30,63 @@
...
 
+    public static final String name = "Max Mustermann";
```
Valodim: I think it's easier and more readable to inline these vars, rather than piece them together with `String.format`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
fiaxh: The db version is increased to trigger a ConsolidateOperation once. The uid parsing gives different results in some cases now, but the user_packets.[name, email, comment] fields in the db do not reflect that. In most cases this is not problematic, as only the uid is taken and simply parsed again, but e.g. for the change in KeychainProvider introduced in this PR it is.

fiaxh:
Renamed class and inlined strings.

dschuermann:
Great work again!

Issue #1468, opened 2015-07-29
Unusable signing key not properly disabled


dschuermann:
Revoked and expired keys are disabled in the signing key spinner. Unusable keys (missing signing subkey) are not properly disabled, they can still be selected by click.

Valodim:
Hm. At first I thought I probably forgot to implement this, but the check is there and looks valid to me. Weird. Will look into it

007durgesh219:
Please see #1818

Valodim:
thanks!

PR #1818, opened 2016-04-04
Fix Unusable signing key not properly disabled, Issue #1468

007durgesh219:
In CertifyKeySpinner and SignKeySpinner item at position 0 will not be "none" entry because they are inner adapters. Please see KeySpinner.java:156. That's why even if the first key is revoked or didn't have a signing subkey it is selectable because it consider them as "none" entries.


Valodim:
yup! nice catch.

Issue #1817, opened 2016-04-04
Crash when no encryption subkey is available


Valodim:
Follow-up on #1807

```
 java.lang.IllegalStateException: no encryption methods specified
 at org.bouncycastle.openpgp.PGPEncryptedDataGenerator.open(PGPEncryptedDataGenerator.java:178)
 at org.bouncycastle.openpgp.PGPEncryptedDataGenerator.open(PGPEncryptedDataGenerator.java:333)
 at org.sufficientlysecure.keychain.pgp.PgpSignEncryptOperation.execute(PgpSignEncryptOperation.java:340)
 at org.sufficientlysecure.keychain.operations.SignEncryptOperation.execute(SignEncryptOperation.java:153)
 at org.sufficientlysecure.keychain.operations.SignEncryptOperation.execute(SignEncryptOperation.java:61)
 at org.sufficientlysecure.keychain.service.KeychainService$1.run(KeychainService.java:146)
```

This error should be handled with a log entry and error status in PgpSignEncryptOperation, and display an error message. For the extra mile, it could also highlight the problematic recipients in our own encryption activities.

007durgesh219:
If the key is revoked or don't have valid encryption subkey then there is no option to use it in entire app except use it from the home screen. There should be a way to avoid that too.  In my suggestion, we should show a message to the user that one of the selected keys is revoked or don't have a valid encryption subkey when he/she tries to use it from the home screen and should not take to encryption activity.

007durgesh219:
setting mFailOnMissingEncryptionKeyIds flag of PgpSignEncryptInputParcel to true will fix this. Will it create any additional problem?? or it's a correct fix.

PR #1819, opened 2016-04-04
Fix Crash when no encryption subkey is available, Issue #1817

007durgesh219:
screenshot_20160405-040401

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/OperationResult.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/operations/results/OperationResult.java
@@ -728,6 +728,7 @@ public abstract class OperationResult implements Parcelable {
         MSG_PSE_ERROR_PGP (LogLevel.ERROR, R.string.msg_pse_error_pgp),
         MSG_PSE_ERROR_SIG (LogLevel.ERROR, R.string.msg_pse_error_sig),
         MSG_PSE_ERROR_UNLOCK (LogLevel.ERROR, R.string.msg_pse_error_unlock),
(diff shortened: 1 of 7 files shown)


Valodim:
Looks good.

I'm a little confused about our own code right now though. @dschuermann Is there a  case where the sign encrypt operation *wouldn't* fail on missing encryption key ids?

dschuermann:
@Valodim No there is no case. I think we can safely remove mFailOnMissingEncryptionKeyIds and the corresponding setters and getters and always fail on missing encryption keys

dschuermann:
Ah and nice pull request @007durgesh219 !

007durgesh219:
Also, App will crash if we try to encrypt file using expried key. We should handle that too.

dschuermann:
Hmm, now I am also confused: Shouldn't it be already impossible to select a key using the dropdown which is either revoked, expired or has no encryption subkeys?

007durgesh219:
No it's not enabled in dropdown but we can always go from home screen. From home screen select any expired key and press encrypt button on action bar.

dschuermann:
It's not crashing for me with an expired key, but you are right using it is still possible.
We should also check for expired and revoked keys inside the encrypt activity, as it can also be called from outside of OpenKeychain with an Intent.
@007durgesh219 would you like to add this functionality to this pr?

007durgesh219:
Sure, I will add it

dschuermann:
BTW thinking about this more: There should be no case where ``PgpSignEncryptOperation`` really crashes, it should always return with ``PgpSignEncryptResult(PgpSignEncryptResult.RESULT_ERROR, log);`` for wrong keys etc.
So if you encounter any case where it actually crashes, please report or make a pull request :)

007durgesh219:
I've updated PR please review it again

dschuermann:
Great work, thanks!

Issue #925, opened 2014-10-02
Encrypt UI: Add "+" button besides TokenAutoComplete


dschuermann:

007durgesh219:
I'd like to work on this one

007durgesh219:
Instead of using ```SelectPublicKeyFragment``` can't we simply make enoughtToFilter always return true and programmatically show the dropdown with empty search query.  This will load all keys from the database and show it in the dropdown

Valodim:
Not sure about this one. @dschuermann  what did you have in mind here?

dschuermann:
I like https://github.com/open-keychain/open-keychain/pull/1822

dschuermann:
Issue is still valid, #1822 has been reverted (see #1822 why).

PR #1822, opened 2016-04-05
Add + button besides TokenAutoComplete, Issue #925

007durgesh219:
screenshot_20160405-233757

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EncryptModeAsymmetricFragment.java
@@ -21,6 +21,7 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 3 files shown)


dschuermann:
Nice, I actually like this approach, much less intrusive than having a fullscreen activity.

@Valodim what do you think?

Valodim:
Yup looks good. I'm not fond of the more narrow dropdown though, can you change the dropdown anchor view to extend to the right in width?

007durgesh219:
Sure, I will update PR soon

007durgesh219:
I've updated PR. Now it will look like this:
screenshot_20160414-023059

@Valodim We can make it look like this also. No margin on left and right
screenshot_20160414-022442

Which one is more sutaible according to you??

dschuermann:
Nice solution and great work. I adapted it a little bit to only expand the width to include the add button.

dschuermann:
Reverted with https://github.com/open-keychain/open-keychain/commit/f44bd236224e782d85c9764a4b655ae922356b14 due to https://github.com/open-keychain/open-keychain/issues/1857

Issue #1550, opened 2015-10-03
USB CCID Support


roblabla:
Hello, I'm wondering if it'd be possible to use a card-reader USB CCID interface with open-keychain. The reason I ask is that I currently have a Yubikey NEO, but no NFC on my phone, so the only way to use the keys stored there from my phone would be through a USB OTG cable.

Any idea what would be necessary to support such a device ? It looks like it is possible to communicate with USB devices with the [USB Host Android API](http://developer.android.com/guide/topics/connectivity/usb/host.html). Would you be interested in supporting it ?

Valodim:
Hi!

So first off, this is certainly possible technically. However, it is at least as much additional code complexity (probably more) as the NFC support we have right now. I also suspect that the actual use case is very rare: few people even consider the possibility to attach usb devices to their phone, and equally few people have a yubikey but no NFC reader.

So yes, this is a conceivable feature, but my gut feeling is that it's not worth it.

dschuermann:
There is an OpenKeychain fork at https://github.com/nmikhailov/open-keychain/tree/otg that looks like it implements partial USB support.

@nmikhailov How is the progress on this? Do you have any questions?

nmikhailov:
@dschuermann Hello. Yes I am trying to implement otg smartcard support(only for yubikey for now). So far it was pretty straightforward. Since you insist I will try to get back to it this week(I was focusing on other projects recently) and maybe file a PR or something.

dschuermann:
@nmikhailov Sounds great, I am looking forward to see a PR :grin: If you have any technical questions, don't hesitate to ask here.

dschuermann:
@nmikhailov Now that the Yubikey 4 is out that supports 4096 bit RSA but only USB, this feature becomes more interesting :)

nmikhailov:
@dschuermann Yes, I have seen the announcement today and though that I should finally finish this. Sorry that I wasn't been able to finish it as promised. I indeed worked on it that week, but ended up loosing half of the time solving device specific otg issues, it could be very tricky and device-dependent.

Valodim:
I would also encourage a work-in-progress PR for early feedback (I do that all the time) :+1:

lbischof:
What's the status on this? Would this also work with a Nitrokey?

nmikhailov:
1. I am planning to return to this soon.
2. Most likely no, at least in first iteration. But you can help test it.
If everything will be fine I might rollout more generic solution.

[...]

PR #1824, opened 2016-04-05
[WIP] USB OTG Support for Yubikey

nmikhailov:
Hello,
This is WIP PR for USB OTG Yubikey support (and possibly other devices like Nitrokey).

This is in very early stage, basically a quickly hacked proof of concept, but I decided to open anyway and to show some progress, as @Valodim suggested in #1550 

Despite being in early stage, everything should be usable, so you can test it if you want.

Issues:
- [x] User shouldn't need to reinset device on each operation
- [x] Disconnects are not handled properly
- [x] Reinsert required in some ocasions(encrypting after reinserting on prev screen)
- [x] Update Manage my keys -> Security token thing
- [x] Investigate issue with transmitting -1 
- [ ] Investigate bricking on card reset
- [ ] Needs refactoring (error handling, ....)
- [ ] Tests?
- [ ] Nitrokey & others

cc @Valodim @dschuermann


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/javacard/JavacardDevice.java
```diff
@@ -0,0 +1,105 @@
+package org.sufficientlysecure.keychain.javacard;
+
+import org.sufficientlysecure.keychain.pgp.CanonicalizedSecretKey;
+import org.sufficientlysecure.keychain.util.Passphrase;
+
+import java.io.IOException;
+
+public interface JavacardDevice {
```
Valodim: why the abstraction, do you expect to have other implementations here? did you want to split this up some way, apart from the transport-abstraction?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/javacard/BaseJavacardDevice.java
```diff
@@ -0,0 +1,712 @@
...
+                    keyType.toString()));
+        }
+
+        putKey(keyType.getmSlot(), secretKey, passphrase);
+        putData(keyType.getmFingerprintObjectId(), secretKey.getFingerprint());
+        putData(keyType.getTimestampObjectId(), timestampBytes);
+    }
+
+    public boolean containsKey(KeyType keyType) throws IOException {
+        return keyMatchesFingerPrint(keyType, BLANK_FINGERPRINT);
```
Valodim: isn't this inverted, i.e. containsKey should return *false* if the key matches the blank fingerprint?

Valodim:
Gave this a whirl but couldn't get it to work. When I plugged the yubikey in during the "hold key now" dialog, the animation twitched but nothing else happened. I'm not entirely sure what to expect though - what's your test setup?

nmikhailov:
@Valodim 
Hello,
There are few things to mention:
1. Please check if OTG works as intended on your android device [with this app](https://play.google.com/store/apps/details?id=aws.apps.usbDeviceEnumerator) . Device should appear on both android and linux tabs.
2. Check if your token matches https://github.com/open-keychain/open-keychain/pull/1824/files#diff-822269f17a87a5393710c77aa1434c15R102 . I hardcoded my devices values, but i guess your vid/pid might be different.

My setup is Nexus 5X and two yubikey neo tokens. What is yours?

nmikhailov:
It should work exactly as NFC works, you just have to connect it with OTG cable instead of bringing token to the back. Pluggin the token on openkeychain's main activity should work.

There might be some delays, but they shouldn't be longer than say 30 seconds.

Valodim:
Got it working, we also talked about this for a while on irc. Most important result is to switch from the busy waiting method to an invisible activity with intent filter + local broadcast mechanism (see [here](http://stackoverflow.com/questions/12388914/usb-device-access-pop-up-supression/15151075#15151075))

nmikhailov:
@jans23 are you still intrested in [giving](https://github.com/open-keychain/open-keychain/issues/1550#issuecomment-171696098) nitrokeys for testing/development?
This becomes more relevant now, as I have time to work on this issue.

Valodim:
@dschuermann and I received nitrokeys, so we can add support and test this :+1:

nmikhailov:
Darn, seems like I missed a chance for a free nitrokey ;-)

Valodim:
(for reference: https://github.com/Yubico/yubikey-personalization/blob/master/ykcore/ykdef.h)

nmikhailov:
Implemented hidden activity technique, whitelisted yubikey devices(with ccid enabled), renames some stuff.
Still need to refactor more, not ready for detailed code review.

jans23:
@nmikhailov I can send you a free Nitrokey. Please place an order in our 
online shop, select "wire transfer" and let me know so that I confirm 
your order for free.

Am 06.04.2016 15:39, schrieb Nikita Mikhailov:

nmikhailov:
Thank you @jans23 ; I have emailed you at jan at nitrokey dot com with order details and few notes.

dschuermann:
@nmikhailov Great work, Should we test and review this pr again?

nmikhailov:
@dschuermann @Valodim Well, I am not finished, but I guess it is in shape for an early review. Everything is supposed to work as intended at this point.
Also I was thinking if we should ditch the whitelist and allow any class 11 device. It would be nice to test with actual CCID compatible card reader, unfortunately mine do not support CCID.

jans23:
@nmikhailov The Nitrokey Pro which I sent you is CCID compliant.

nmikhailov:
@jans23 yes, that's great. I am sure that nitrokey will be supported and tested even if we will keep the whitelist.

dschuermann:
The main things worked for me with a YubiKey NEO but only in once :D 

* When creating a key, the screen is shown with "Hold Security Token against back of your device" even when the token is already attached via USB. I was able to skip the screen by removing the token and then re-inserting it. Can we skip the screen directly?

* Sometimes I get "Error: USB error - Failed to transmit data-1" at the same screen. Removing and reinserting the token does not help. Only removing OpenKeychain from the task list and then trying again helped.

* Signing data worked very nice!

* And I succeeded in bricking a YubiKey NEO *yeah* Resetting a security token did not work properly (remove the key from the list and try to create a new one, OK will redirect you to the reset screen). I got "Error: Reactivating failed". Then I get "initilaization failed" and later "invalid length field" after trying to reactivate it using ``gpg-connect-agent --hex`` and the [procedures from ykneo applet](https://github.com/Yubico/ykneo-openpgp/blob/master/doc/ResetApplet.txt). gpg says:
```
schuerm@smaug:~$ gpg --card-status 
gpg: detected reader `Yubico Yubikey NEO CCID 00 00'
gpg: pcsc_connect failed: sharing violation (0x8010000b)
gpg: apdu_send_simple(0) failed: locking failed
Please insert the card and hit return or enter 'c' to cancel: c
gpg: selecting openpgp failed: general error
gpg: OpenPGP card not available: general error
schuerm@smaug:~$ gpg-connect-agent --hex "scd apdu 00 f1 00 00" /bye 
ERR 100663427 Conditions of use not satisfied <SCD>
schuerm@smaug:~$ gpg-connect-agent --hex
> scd apdu 00 e6 00 00
ERR 100663427 Conditions of use not satisfied <SCD>
> scd apdu 00 20 00 81 08 40 40 40 40 40 40 40 40
ERR 100663427 Conditions of use not satisfied <SCD>
> scd apdu 00 e6 00 00
ERR 100663351 Invalid value <SCD> 
```

* We should maybe disable the reset procedure for USB. 

* I would prefer if we keep a whitelist of security tokens that we successfully tested! And also maintain a list on our website.

nmikhailov:
Well your experience was much worse than I anticipated :hurtrealbad: 
1. I guess you launched it though Manage my keys -> use security token? Yeah, I kinda skipped this screen.
2. I will have to take a look into it.
3. ok
4. Well that is really bad. I don't know what went wrong, I have done it multiple times during my tests, it worked flawlessly. Sorry for bricking your yk :-(

dschuermann:
no worries, I have many test YubiKeys. I think it was one from the later shipments which are locked down, else I would have just re-flashed the applet.

nmikhailov:
Could you try with unlocked yk then? If its only openpgp applet which is broken then reflashing should work i guess.

nmikhailov:
I have successfully reseted yk with pgp-applet versions 01 00 10 and 01 00 08

nmikhailov:
There is a possibility of bugged yubikey:
http://forum.yubico.com/viewtopic.php?f=26&t=1520
http://forum.yubico.com/viewtopic.php?f=26&t=1692

nmikhailov:
@dschuermann I have updated PR, there should be much less transmission errors and CreateSecurityTokenWaitFragment would be skipped if there is a device connected.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/UsbTransport.java
```diff
@@ -0,0 +1,303 @@
...
+ * Based on USB CCID Specification rev. 1.1
+ * http://www.usb.org/developers/docs/devclass_docs/DWG_Smart-Card_CCID_Rev110.pdf
+ * Implements small subset of these features
+ */
+public class UsbTransport implements Transport {
+    private static final int USB_CLASS_SMARTCARD = 11;
+    private static final int TIMEOUT = 20 * 1000; // 20s
+
+    private final UsbManager mUsbManager;
+    private final android.hardware.usb.UsbDevice mUsbDevice;
```
dschuermann: Why the usage of full qualified class name here ``android.hardware.usb.UsbDevice`` and in other places in this class?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseSecurityTokenNfcActivity.java
```diff
@@ -121,48 +110,65 @@ protected void onNfcPostExecute() {
...
                 super.onPreExecute();
-                onNfcPreExecute();
+                onSecurityTokenPreExecute();
             }
 
             @Override
             protected IOException doInBackground(Void... params) {
                 try {
-                    handleTagDiscovered(tag);
+                    handleSmartcard(transport);
```
dschuermann: still called smartcard here

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseSecurityTokenNfcActivity.java
```diff
@@ -175,11 +181,11 @@ protected void onPostExecute(IOException exception) {
                 super.onPostExecute(exception);
 
                 if (exception != null) {
-                    handleNfcError(exception);
+                    handleSmartcardError(exception);
```
dschuermann: still called smartcard here

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseSecurityTokenNfcActivity.java
```diff
@@ -230,15 +237,15 @@ public void onNewIntent(final Intent intent) {
         mTagDispatcher.interceptIntent(intent);
     }
 
-    private void handleNfcError(IOException e) {
+    private void handleSmartcardError(IOException e) {
```
dschuermann: still called smartcard here

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperationTest.java
```diff
@@ -887,8 +886,8 @@ public void testKeyToSecurityToken() throws Exception {
             PgpKeyOperation op = new PgpKeyOperation(null);
             PgpEditKeyResult result = op.modifySecretKeyRing(secretRing, cryptoInput, parcelSecurityToken);
             Assert.assertTrue("moveKeyToSecurityToken operation should be pending", result.isPending());
-            Assert.assertEquals("required input should be RequiredInputType.NFC_MOVE_KEY_TO_CARD",
-                    result.getRequiredInputParcel().mType, RequiredInputType.NFC_MOVE_KEY_TO_CARD);
+            Assert.assertEquals("required input should be RequiredInputType.SMARTCARD_MOVE_KEY_TO_CARD",
+                    result.getRequiredInputParcel().mType, RequiredInputType.SMARTCARD_MOVE_KEY_TO_CARD);
```
dschuermann: still called smartcard here

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/UsbTransport.java
nmikhailov: Seems like an IDE artifact after accidental renaming and undoing. Will fix

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/base/BaseSecurityTokenNfcActivity.java
nmikhailov: yeah, I have renamed everything two times now. I decided to stick with SecurityToken as it was used already. Will rename whats left.

dschuermann:
Looks good besides the nitpicks.
Will try to find some time tomorrow to test it again with some YubiKeys.

I think we can merge this one even when the Nitrokey has not been added to the list if everything works as expected tomorrow.

For later pull requests I dream of:
* More devices to the whitelist: e.g., Nitrokey
* Other pictures with USB devices for the dialog: maybe we need to add those to the Nordpol lib
* Key generation on security token: a method is already there but not used because key generation over NFC is extremely unreliable and works only 1/3 of the time. I suspect that is because of power issues, which should not happen with USB.

nmikhailov:
1. I can add it after I will receive it
2. Not sure why 'Fidesmo Nordpol - The Android Support Library for NFC' would accept USB related stuff. Also texts should suggest that user can use otg cable.
3. That is interesting.
+
4. It is not totally clear what happend with your yubikey. I assume its due bad firmware, but who knows.  I think there should be a disclaimer and warning before wiping. 
Nitrokey's official [reset sequence](https://www.nitrokey.com/sites/default/files/nitrokey-reset.txt)  is almost like ours except for invalid pin values, seems like our current implementation should work.

dschuermann:
Awesome work. Let's do improvements in new pull requests. I added an issue for possible improvements here: https://github.com/open-keychain/open-keychain/issues/1833

nmikhailov:
@dschuermann I guess now we can close #1550

tristan-k:
When will this be released?

dschuermann:
@tristan-k later ;)

analogbyte:
A minor suggestion which has not been brought up above: A gpg card PIN may be alphanumeric, but the current implementation only offers a dial keyboard. This prevents me from giving the correct PIN as I am unable to switch to a full keyboard (this may be dependent on the keyboard used, though). I don't think having a full keyboard by default makes much sense, but maybe there is a way to make this optional?

With a numeral-only PIN everything works fine, thanks for your effort @nmikhailov. :+1:

nmikhailov:
@analogbyte you are right, its currently not possible to enter non numeric pin; I will add this to my list

dschuermann:
@analogbyte it's already possible. Please look into the settings. Although it doesn't make much sense security wise to use something more difficult than 6 numbers. 10^6 possible combinations with only 3 tries for an attacker is already hard enough. Compare this with 4 numbers of your credit card ;)

analogbyte:
@dschuermann oh, I don't know how I managed to overlook the option in the afternoon, I'm sorry.

I know about the security implications, but I generally tend to use short (pseudo-)words instead of numbers where ever possible.

Thanks again for your work on this!

PR #1828, opened 2016-04-10
Add missing keyholder string param

nmikhailov:
Not sure about other languages though. I assume they should be fixed via transifex, right?


dschuermann:
Thanks, yep the other languages are "automatically" fixed via transifex.

Issue #1825, opened 2016-04-06
Cannot use i2p Email address


dschuermann:
We prevent the usage of test@mail.i2p. We should fix our regex or simply don't hard-fail if a non-conforming Email address is added. I favor the second option.

Valodim:
Second option seems good. We also might want to support email address only user ids, for posteo

007durgesh219:
Can't we allow numbers after . (dot) in regex. That should fix it.

PR #1831, opened 2016-04-14
Handle user input on key creation more generously

fiaxh:
Allows an empty name on key generation and does not enforce email formats by regex matching. Instead, it shows a warning in the CreateKeyFinalFragment if an email does not match the regex. The explanation is only shown when the email field was clicked on, to not hide the sync option.

relevant issue: #1825


dschuermann:
Great solution, works!

PR #1832, opened 2016-04-15
Fix Wrong file encrypted from cached input

007durgesh219:
To reproduce it:

1.  Select a file to encrypt
2.  Click on the save icon
<< Now user might realize that he selected the wrong file to encrypt So, he/she will cancel it (don't save) >>
3.  Remove file
4.  Select correct file
5.  Save it

But it will still encrypt the previous file from cache.


dschuermann:
Great thanks!

PR #1835, opened 2016-04-15
Check security token

nmikhailov:
Currently user can sign and decrypt with wrong security token which should not happen.
This PR adds checks before signing and decryption.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/input/RequiredInputParcel.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/input/RequiredInputParcel.java
@@ -89,7 +89,7 @@ public class RequiredInputParcel implements Parcelable {
         return new RequiredInputParcel(RequiredInputType.ENABLE_ORBOT, null, null, null, 0L, 0L);
     }
 
-    public static RequiredInputParcel createNfcSignOperation(
+    public static RequiredInputParcel createSecurityTokenSignOperation(
             long masterKeyId, long subKeyId,
             byte[] inputHash, int signAlgo, Date signatureTime) {
         return new RequiredInputParcel(RequiredInputType.SECURITY_TOKEN_SIGN,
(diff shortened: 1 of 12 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/service/input/RequiredInputParcel.java
```diff
@@ -188,14 +188,14 @@ public RequiredInputParcel createFromParcel(final Parcel source) {
         }
     };
 
-    public static class NfcSignOperationsBuilder {
+    public static class SecurityTokrnSignOperationsBuilder {
```
dschuermann: spelling mistake :P

nmikhailov:
@dschuermann oops, fixed

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SecurityTokenOperationActivity.java
```diff
@@ -183,6 +185,13 @@ protected void doSecurityTokenInBackground() throws IOException {
 
         switch (mRequiredInput.mType) {
             case SECURITY_TOKEN_DECRYPT: {
+                long tokenKeyId = KeyFormattingUtils.getKeyIdFromFingerprint(
+                        mSecurityTokenHelper.getMasterKeyFingerprint(KeyType.SIGN));
```
dschuermann: is this the right KeyType?

dschuermann: Ah SIGN is the first key which is also the master key correct?
Is the method ``getMasterKeyFingerprint`` named properly? Should it be named ``getKeyFingerprint``? I know it's not your fault :P

nmikhailov: You are right, I think it would be better to check if  subkey ids match and rename the method

dschuermann:
awesome work, thanks!

PR #1858, opened 2016-05-07
SecurityToken: add support for 3072 & 4096bit RSA keys

### LINKED ISSUE #1852: Error with Yubikey 4 Nano during decryption and signing.
Hi all!
When i try to decrypt or sign some text/file, the app alerts me saying that it expects a sign long 128 or 256 bytes, but it was 512.

I encounter this error only when i try to use the smartcard functionality of the Yubikey 4 Nano, if i import the secret subkeys directly to openkeychain, i can decrypt and sign without problems.

The keys are RSA4096.

nmikhailov:
fixes #1852


dschuermann:
Great work as always

LucaZulberti:
Good job! :-)
When it will be available on Google Play?

PR #1860, opened 2016-05-08
in OpenPgpServiceTest, use new service interface instead of deprecated old interface

mgeier63:
just a small change to make sure the test is using the "new" service


mgeier63:
^^^ oops, the JVM crashed when running (unrelated) tests. Seems to be a problem with Travis itself  ?!

dschuermann:
Thanks, jup it's a bug in Travis...

dschuermann:
for reference: https://github.com/open-keychain/open-keychain/issues/1861

PR #1865, opened 2016-05-11
Fixed 'Orbot appears not to be running' (#1792)

vanitasvitae:
Hi!
I tried my best to fix the faulty "Orbot appears not to be running" message on Android 6 by ditching the old OrbotHelper.isOrbotRunning() method. I implemented a BroadcastReceiver that listens for Orbot Status intents and acts as a singleton with getters for the received states.

Im unsure, whether this is secure and proper implemented or not, so please take a close look at my changes :)


dschuermann:
Looks good.
Does this work properly in your tests?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/orbot/OrbotStatusReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+
+import org.sufficientlysecure.keychain.util.Log;
+
+/**
+ * Created by vanitas on 11.05.16.
```
dschuermann: remove the "created by" line, instead add a license header like in other files with your name

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/orbot/OrbotStatusReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+ * BroadcastReceiver that receives Orbots status
+ */
+public class OrbotStatusReceiver extends BroadcastReceiver {
+
+    private static final String TAG = "OrbStatRec";
```
dschuermann: use Constants.TAG instead

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/orbot/OrbotStatusReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+    private int proxy_port_socks;
+
+    private static OrbotStatusReceiver instance;
+
+    public OrbotStatusReceiver() {
+        instance = this;
+    }
+
+    public static OrbotStatusReceiver getInstance() {
+        if(instance == null) instance = new OrbotStatusReceiver();
```
dschuermann: if else always with {}

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/orbot/TorServiceUtils.java
```diff
@@ -1,156 +0,0 @@
-/* This is the license for Orlib, a free software project to
```
dschuermann: great to get rid of this file :) I don't like it

vanitasvitae:
I changed the points you mentioned. Im not sure what you meant with the authorship so I removed it completely.

I tested this manually on my device by deactivating/starting Orbot. Are there any test cases which I can use or should I write my own?

dschuermann:
Great work!

dschuermann:
Regarding tests: I don't think this can be tested in an automated manner.

vanitasvitae:
Is seems like OpenKeyChain is checking Orbots status nearly every second. I'm going to take a look at this since it may drain battery.

PR #1834, opened 2016-04-15
WIP: Add Nitrokey support

dschuermann:
Currently I get
```
AndroidRuntime  E  Process: org.sufficientlysecure.keychain.debug, PID: 1924
  1924         AndroidRuntime  E  java.lang.RuntimeException: An error occured while executing doInBackground()
  1924         AndroidRuntime  E  at android.os.AsyncTask$3.done(AsyncTask.java:304)
  1924         AndroidRuntime  E  at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
  1924         AndroidRuntime  E  at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
  1924         AndroidRuntime  E  at java.util.concurrent.FutureTask.run(FutureTask.java:242)
  1924         AndroidRuntime  E  at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
  1924         AndroidRuntime  E  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
  1924         AndroidRuntime  E  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
  1924         AndroidRuntime  E  at java.lang.Thread.run(Thread.java:818)
  1924         AndroidRuntime  E  Caused by: java.lang.OutOfMemoryError: Failed to allocate a 688448510 byte allocation with 16777216 free bytes and 166MB until OOM
  1924         AndroidRuntime  E  at org.sufficientlysecure.keychain.securitytoken.UsbTransport.receive(UsbTransport.java:255)
  1924         AndroidRuntime  E  at org.sufficientlysecure.keychain.securitytoken.UsbTransport.setIccPower(UsbTransport.java:82)
  1924         AndroidRuntime  E  at org.sufficientlysecure.keychain.securitytoken.UsbTransport.connect(UsbTransport.java:192)
  1924         AndroidRuntime  E  at org.sufficientlysecure.keychain.securitytoken.SecurityTokenHelper.connectToDevice(SecurityTokenHelper.java:150)
  1924         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenNfcActivity.handleSecurityToken(BaseSecurityTokenNfcActivity.java:412)
[...]

PR #1868, opened 2016-05-14
WIP: Security token USB handling improvements

nmikhailov:
Multiple security token improvements, mainly to support Nitrokey Pro.
Substitutes  #1834
Huge changes so I encourage mass regression testing.

TODO: 
- [x] Nitrokey Pro
- [ ] Nitrokey Start/GNUK
- [x] moving key to Nitrokey doesn't work
- [ ] refactor & add some comments
- [ ] fix CR


dschuermann:
Wow, huge number of changes :D

I can check if everything still works with Yubikey NEO because I have one with me but I cannot check with Fidesmo and others cards because those are in my Office. I am back in my office next month in June, not before (currently on research stay in Helsinki).

nmikhailov:
No problem. I would suggest not to test at all if you only have production device(remember how it went last time? :p).

dschuermann:
Funnily, I have a Yubikey NEO dev device with me that is not my production device. Forgot it in my bag before departure. Will test later with this one...

dschuermann:
I tested e5d3d1f, not the last three commits:
With YubiKey NEO v1.0.11 everything still works as expected (key creation, reset, sign, decrypt) I also tested over the API.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPGPCapabilities.java
```diff
@@ -0,0 +1,174 @@
...
+
+package org.sufficientlysecure.keychain.securitytoken;
+
+import org.sufficientlysecure.keychain.util.Iso7816TLV;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OpenPGPCapabilities {
```
dschuermann: Please call the class ``OpenPgpCapabilities``. I know this is not in line with Bouncy Castle, but this is how Android's code style is defined, see https://github.com/open-keychain/open-keychain#code

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -230,16 +209,11 @@ public void modifyPin(int pw, byte[] newPin) throws IOException {
...
-                + "24" // INS
-                + "00" // P1
-                + String.format("%02x", pw) // P2
-                + String.format("%02x", pin.length + newPin.length) // Lc
-                + getHex(pin)
-                + getHex(newPin);
-        String response = communicate(changeReferenceDataApdu); // change PIN
-        if (!response.equals("9000")) {
-            throw new CardException("Failed to change PIN", parseCardStatus(response));
+        CommandAPDU changePin = new CommandAPDU(0x00, 0x24, 0x00, pw, Arrays.concatenate(pin, newPin));
```
dschuermann: Can you try to keep some comments, like "CLA, INS, P1" from before. It makes it easier for non-javacard geeks to look up the commands in the OpenPGP card spec ;)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPGPCapabilities.java
```diff
@@ -0,0 +1,174 @@
...
+    private int mMaxRspLen;
+
+    private Map<KeyType, AlgorithmFormat> mKeyFormats;
+
+    public OpenPGPCapabilities(byte[] data) throws IOException {
+        Iso7816TLV[] tlvs = Iso7816TLV.readList(data, true);
+        mKeyFormats = new HashMap<>();
+
+        for (Iso7816TLV tlv : tlvs) {
+            switch (tlv.mT) {
```
dschuermann: can you put in some comments where the values in this switch come from

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPGPCapabilities.java
```diff
@@ -0,0 +1,174 @@
...
+                    parseDdo((Iso7816TLV.Iso7816CompositeTLV) tlv);
+                    break;
+            }
+        }
+    }
+
+    private void parseDdo(Iso7816TLV.Iso7816CompositeTLV tlvs) {
+        for (Iso7816TLV tlv : tlvs.mSubs) {
+            switch (tlv.mT) {
+                case 0xC0:
```
dschuermann: can you put in some comments where the values in this switch come from

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPGPCapabilities.java
```diff
@@ -0,0 +1,174 @@
...
+                    break;
+            }
+        }
+    }
+
+    private void parseAlgoCaps(KeyType keyType, byte[] data) {
+        mKeyFormats.put(keyType, AlgorithmFormat.from(data[5]));
+    }
+
+    private void parseExtendedCaps(byte[] v) {
```
dschuermann: again some documentation would be awesome

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPGPCapabilities.java
nmikhailov: What about imported ones like CommandAPDU? I don't want to modify them because they are kinda standard(this matters since I consider possibility of moving this to a separate lib).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -150,54 +155,28 @@ public boolean keyMatchesFingerPrint(KeyType keyType, byte[] fingerprint) throws
...
-                        + "A4" // INS
-                        + "04" // P1
-                        + "00" // P2
-                        + "06" // Lc (number of bytes)
-                        + "D27600012401" // Data (6 bytes)
-                        + "00"; // Le
-        String response = communicate(opening);  // activate connection
-        if (!response.endsWith(accepted)) {
-            throw new CardException("Initialization failed!", parseCardStatus(response));
+        CommandAPDU select = new CommandAPDU(0x00, 0xA4, 0x04, 0x00, Hex.decode("D27600012401"));
```
dschuermann: keep comments about the values for reference

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -254,38 +228,20 @@ public void modifyPin(int pw, byte[] newPin) throws IOException {
...
-        // Receive
-        String result = getDataField(response);
-        while (response.endsWith("61")) {
-            response = communicate("00C00000" + status.substring(2));
-            status = response.substring(response.length() - 4);
-            result += getDataField(response);
-        }
-        if (!status.equals("9000")) {
-            throw new CardException("Deciphering with Security token failed on receive", parseCardStatus(response));
+        CommandAPDU command = new CommandAPDU(0x00, 0x2A, 0x80, 0x86, data, MAX_APDU_NE_EXT);
```
dschuermann: again some documentation would be awesome

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -342,16 +295,11 @@ private void putData(int dataObject, byte[] data) throws IOException {
             verifyPin(0x83); // (Verify PW3)
         }
 
-        String putDataApdu = "00" // CLA
-                + "DA" // INS
-                + String.format("%02x", (dataObject & 0xFF00) >> 8) // P1
-                + String.format("%02x", dataObject & 0xFF) // P2
-                + String.format("%02x", data.length) // Lc
-                + getHex(data);
+        CommandAPDU command = new CommandAPDU(0x00, 0xDA, (dataObject & 0xFF00) >> 8, dataObject & 0xFF, data);
```
dschuermann: keep comments about values

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -391,16 +335,70 @@ private void putKey(int slot, CanonicalizedSecretKey secretKey, Passphrase passp
             verifyPin(0x83); // (Verify PW3 with mode 83)
         }
 
+        ByteArrayOutputStream stream = new ByteArrayOutputStream(),
+                template = new ByteArrayOutputStream(),
+                data = new ByteArrayOutputStream();
+
+        // Public modulus, length 3
```
dschuermann: great to see these and the other operations here explained :+1:

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -518,104 +523,128 @@ public String getUserId() throws IOException {
...
-
-        String response = communicate(apdu);
-
-        if (response.length() < 4) {
-            throw new CardException("Bad response", (short) 0);
-        }
-        // split up response into signature and status
-        String status = response.substring(response.length() - 4);
-        String signature = response.substring(0, response.length() - 4);
+        CommandAPDU command = new CommandAPDU(0x00, 0x2A, 0x9E, 0x9A, dsi, MAX_APDU_NE_EXT);
```
dschuermann: keep comments about values

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -664,38 +692,19 @@ public boolean isFidesmoToken() {
...
-        String login =
-                "00" // CLA
-                        + "20" // INS
-                        + "00" // P1
-                        + String.format("%02x", mode) // P2
-                        + String.format("%02x", pin.length) // Lc
-                        + Hex.toHexString(pin);
-
-        return communicate(login);
+        return communicate(new CommandAPDU(0x00, 0x20, 0x00, mode, pin));
```
dschuermann: keep comments about values

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
nmikhailov: I disagree, they are explicitly documented in CommandAPDU constructor. Which is also much more clear. No objection to leaving a link to spec about concrete command though.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -704,35 +713,36 @@ private String tryPin(int mode, byte[] pin) throws IOException {
...
         // reactivate token!
-        String reactivate1 = "00" + "e6" + "00" + "00";
-        String reactivate2 = "00" + "44" + "00" + "00";
-        String response1 = communicate(reactivate1);
-        String response2 = communicate(reactivate2);
-        if (!response1.equals(accepted) || !response2.equals(accepted)) {
-            throw new CardException("Reactivating failed!", parseCardStatus(response1));
-        }
 
+        CommandAPDU reactivate1 = new CommandAPDU(0x00, 0xE6, 0x00, 0x00);
```
dschuermann: while looking at this I remember something (not related to your refactorings):
Can you please add a comment here:
``// NOTE: keep the order here! First execute _both_ reactivate commands. Before checking _both_ response
// If a token is in a bad state and reactivate1 fails, it could still be reactivated with reactivate2``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/smartcardio/CommandAPDU.java
```diff
@@ -0,0 +1,605 @@
...
+ * <p>Instances of this class are immutable. Where data is passed in or out
+ * via byte arrays, defensive cloning is performed.
+ *
+ * @see ResponseAPDU
+ *
+ * @since   1.6
+ * @author  Andreas Sterbenz
+ * @author  JSR 268 Expert Group
+ */
+public final class CommandAPDU implements java.io.Serializable {
```
dschuermann: ``CommandApdu``

dschuermann: and import java.io.Serializable instead of using the fully qualified package name

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/smartcardio/ResponseAPDU.java
```diff
@@ -0,0 +1,184 @@
...
+ * <p>Instances of this class are immutable. Where data is passed in or out
+ * via byte arrays, defensive cloning is performed.
+ *
+ * @see CommandAPDU
+ *
+ * @since   1.6
+ * @author  Andreas Sterbenz
+ * @author  JSR 268 Expert Group
+ */
+public final class ResponseAPDU implements java.io.Serializable {
```
dschuermann: ``ResponseApdu``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/usb/CcidTransceiver.java
```diff
@@ -0,0 +1,169 @@
...
+            try {
+                atr = receiveRaw();
+                break;
+            } catch (Exception e) {
+                // Try more startTime
+                if (System.currentTimeMillis() - startTime > TIMEOUT) {
+                    break;
+                }
+            }
+            SystemClock.sleep(100);
```
dschuermann: why?

OpenKeychain/src/main/res/xml/usb_device_filter.xml
```diff
@@ -24,4 +27,12 @@
     <!-- Yubikey 4 OTP + U2F + CCID-->
     <usb-device class="11" vendor-id="4176" product-id="1031"/>
 
+
+    <!-- Nitrokey Pro-->
+    <usb-device class="11" vendor-id="8352" product-id="16648"/>
+    <!-- Nitrokey Storage-->
+    <!--<usb-device class="11" vendor-id="8352" product-id="16649"/>-->
+    <!-- Nitrokey Start-->
+    <!--<usb-device class="11" vendor-id="8352" product-id="16913"/>-->
```
dschuermann: can you add a comment here why these two are not activated right now for future reference?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
dschuermann: Ah, yes you are right

dschuermann:
Regarding the imported classes from smartcardio:
Could we just leave them unchanged and in their original package `` javax.smartcardio``?
Then it's much more clear that they are coming from somewhere else and have been imported.

Also if there are fixes available in OpenJDK's repository for these classes, it is important to know which version we have. Could you add some headers on top of these two classes referencing the commit where they have been taken and maybe date when they have been taken from OpenJDK?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/smartcardio/ResponseAPDU.java
dschuermann: new considerations about these classes at the bottom...

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/smartcardio/CommandAPDU.java
dschuermann: new considerations about these classes at the bottom...

dschuermann:
Review done :+1: 

I would suggest to keep this in OpenKeychain for this pull request and then consider a standalone library after this one has been merged and tested in a release. In general I think a library could be nice, maybe we can chat later where an appropriate place for it would be: Bouncy Castle, standalone, or as part of a library with other pgp classes from OpenKeychain's backend. I am not sure.

nmikhailov:
Thanks, I will make appropriate adjustments.

dschuermann:
Forget about the idea to place it in Bouncy Castle: It would not work due to conflicting licenses :/

dschuermann:
@nmikhailov You can now also work in a branch on open-keychain's github org. I gave you access rights.

nmikhailov:
Well what I was initially thinking is moving out stuff from `org.sufficientlysecure.keychain.securitytoken.usb` which is mainly usb transport support. 
And maybe uniting it with NFC under `javax.smartcardio` like api. 

This way it is general-purposed, doesn't know anything about pgp or cryptography at all, and should be standalone.

dschuermann:
Sounds like a good idea, so this library should be Android specific I assume?

nmikhailov:
Yes, as other platforms already have 'javax.smartcardio' implementation based on system libraries.
This one will be based on `android.hardware.usb.UsbDevice` & `android.nfc.tech.IsoDep`.

dschuermann:
BTW@all: I just disabled coveralls comments. They are annoying. It's already displayed in Githubs checks box.

nmikhailov:
I agree, they are a bit annoying; especialy in wip PRs.

nicorikken:
I assume this would work for the Nitrokey Start as well, as it is based on USB also? I have one, as even the crypto is open source, based on [Gnuk](http://www.fsij.org/category/gnuk.html) Either way, great to see Nitrokey support coming 👍

nmikhailov:
@nicorikken check previous discussion regarding status of 'Start' support - https://github.com/open-keychain/open-keychain/pull/1834

nicorikken:
@nmikhailov Thanks for the info. I should have done a search myself, sorry. Great to see support coming though.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KeyFormat.java
```diff
@@ -0,0 +1,87 @@
...
+        STANDARD(0, false, false),
+        STANDARD_WITH_MODULUS(1, false, true),
+        CRT(2, true, false),
+        CRT_WITH_MODULUS(3, true, true);
+
+        private int mValue;
+        private boolean mIncludeModulus;
+        private boolean mIncludeCrt;
+
+        AlgorithmFormat(int value, boolean includeCrt, boolean includeModulus) {
```
Valodim: private

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
```diff
@@ -0,0 +1,154 @@
...
+    private boolean mHasKeyImport;
+
+    private byte mSMAlgo;
+    private int mMaxCmdLen;
+    private int mMaxRspLen;
+
+    private Map<KeyType, KeyFormat> mKeyFormats;
+
+    public OpenPgpCapabilities(byte[] data) throws IOException {
+        Iso7816TLV[] tlvs = Iso7816TLV.readList(data, true);
```
Valodim: There is a lot of logic in this constructor, and it throws an IOException which is not very nice. All fields are determined entirely in the constructor if I'm not missing something, so moving the parsing logic into a static method and making all fields final would yield a nice immutable class here.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/usb/CcidTransceiver.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private void sendRaw(final byte[] data) throws UsbTransportException {
+        final int tr1 = mConnection.bulkTransfer(mBulkOut, data, data.length, TIMEOUT);
+        if (tr1 != data.length) {
+            throw new UsbTransportException("USB error - failed to transmit data " + tr1);
+        }
+    }
+
+    private byte getStatus(byte[] bytes) {
```
Valodim: static, and could also use a short comment on what's going on here (like the other methods refer to relevant specs)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
Valodim: This actually doesn't happen anymore - if response1 is bad, it will throw.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -704,35 +602,36 @@ private String tryPin(int mode, byte[] pin) throws IOException {
...
+        CommandAPDU reactivate2 = new CommandAPDU(0x00, 0x44, 0x00, 0x00);
+        ResponseAPDU response1 = communicate(reactivate1);
+        ResponseAPDU response2 = communicate(reactivate2);
+        if (response1.getSW() != APDU_SW_SUCCESS) {
+            throw new CardException("Reactivating failed!", response1.getSW());
+        }
+        if (response2.getSW() != APDU_SW_SUCCESS) {
+            throw new CardException("Reactivating failed!", response2.getSW());
+        }
     }
```
Valodim: if this point is reached, it means the reactivation was successful, right? might be worth a `Log.d`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -518,104 +412,128 @@ public String getUserId() throws IOException {
...
+            }
+        }
+        if (lastResponse == null) {
+            throw new UsbTransportException("Can't transmit command");
+        }
+
+        result.write(lastResponse.getData());
+
+        // Receive
+        while (lastResponse.getSW1() == 0x61) {
```
Valodim: 0x61 magic constant

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -518,104 +412,128 @@ public String getUserId() throws IOException {
...
-        return Hex.decode(signature);
+        return signature;
     }
 
     /**
      * Transceive data via NFC encoded as Hex
      */
-    private String communicate(String apdu) throws IOException {
-        return getHex(mTransport.transceive(Hex.decode(apdu)));
+    private ResponseAPDU communicate(CommandAPDU apdu) throws IOException {
```
Valodim: Considering testability, this method can be neatly split into `sendCommandAndReceiveFirstResponse(CommandAPDU)` and `readFullResponse(ResponseAPDU)`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/CardCapabilities.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.nio.ByteBuffer;
+
+public class CardCapabilities {
+    private static final int MASK_CHAINING = 1 << 7;
+    private static final int MASK_EXTENDED = 1 << 6;
+
+    private byte[] mCapabilityBytes;
+
+    public CardCapabilities(byte[] historicalBytes) throws UsbTransportException {
```
Valodim: same immutable argument as for OpenPgpCapabilities

Valodim:
Added a couple of minor comments. Most of this looks perfect, great job!

mr-teaser:
Hi,
I want to use my nitrokey pro with my android and therefore I imported the version "nmikhailov:nitrokey" to androidstudio and installed it on my mobile.
It compiles flawlessly, but when I start the app and plug in my nitrokey the app stops.
I appreciate any hints how I could use the app. Which version should I use?

Thanks in advanced and also thanks for all the work!

nmikhailov:
@mr-teaser Do you use Nitrokey Pro? Can you paste adb logs somewhere?

41px:
Any idea on when this feature available in GooglePlay ?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KeyFormat.java
nmikhailov: Its redundant, actually Android Studio even highlights this and add a warning.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
nmikhailov: I don't quite see benefit of your approach, I only see a disadvantage that I will need more code for private constructor which takes all 10 arguments. I am not even sure that this constuctor falls into 'don't do real work in constuctor' antipattern and I don't see the difference with static method - it will be same thing. Same about CardCapabilities

nmikhailov:
@dschuermann Ok, so I fixed some of the CRs, please tell me if I missed something important.
As we discussed this is only Nitrokey Pro, without GNUK/NiktrokeyStart(which would come in another PR later).

dschuermann:
Thanks for your work on this! I will shortly release a beta version with this stuff.

dschuermann:
This stuff has been released as 4.2 beta1 on Google Play for beta testers (see https://www.openkeychain.org/help/ for how to get beta updates).

JohanLorenzo:
Thank you guys for making this support and showing an early preview of it 😃 

For the record, you may experience #1936 on Nitrokey Pro, and #1937 on Nitrokey Storage.

PR #1869, opened 2016-05-14
Add new build type to mitigate debugging issues

nmikhailov:
See http://stackoverflow.com/questions/27909613/cannot-see-parameter-value-in-android-studio-when-breakpoint-is-in-first-line-of

Very annoying.

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -190,9 +190,6 @@ android {
             minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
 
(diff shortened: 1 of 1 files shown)


dschuermann:
Hmmm, annoying bug. Can we define that inverse that "debug" works with the parameter value and "debugWithTestCoverage" is executed with Jacoco (also changing the .travis.yml)?

dschuermann:
great

dschuermann:
Looks like coveralls started working again after your fixes :grimacing: 
great!

nmikhailov:
Yes, it was misconfigured before, maybe due to gradle update or/and adding flavours.

PR #1880, opened 2016-05-23
Adds support for ACTION_DECODE_PUBLIC_KEY

adithyaphilip:
Reads a public key in from the InputStream and returns the master key id, and user ids.
Necessary for the contact key retrieval portion in OX.

Was tested with Conversations (the action and result strings were hardcoded).

Current workflow in Conversations:
- Contact key downloaded, key ID and user IDs retrieved via this action.
- If xmpp:user@email is one of the user IDs, ACTION_GET_KEY is used with the key id to check if the key already exists in OK. Else, an error is toasted saying contact key does not contain the required user id.
- If the key does not exist in OK, it should be added (would firing an "Import Key via OK" intent suffice?).
- If the key exists in OK, it is associated with the contact in Conversations.

Depends on https://github.com/open-keychain/openpgp-api/pull/6 being merged before the corresponding submodule can be updated.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
@@ -69,6 +69,8 @@ import org.sufficientlysecure.keychain.pgp.PgpSignEncryptData;
 import org.sufficientlysecure.keychain.pgp.PgpSignEncryptInputParcel;
 import org.sufficientlysecure.keychain.pgp.PgpSignEncryptOperation;
 import org.sufficientlysecure.keychain.pgp.Progressable;
+import org.sufficientlysecure.keychain.pgp.UncachedKeyRing;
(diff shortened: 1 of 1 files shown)


dschuermann:
Hmmm,
maybe we should create a new class named ``OpenPgpKeyInformation`` that implements Parcelable (similar to ``OpenPgpSignatureResult``) which includes:
```
long keyId;
String primaryUserId;
ArrayList<String> userIds;
```
and maybe more information.

Then we could return that instead of single extras. Also we could then use ``OpenPgpKeyInformation`` as an variable in ``OpenPgpSignatureResult`` and deprecate the variables keyId, primaryUserId, userIds inside that class.

@Valodim what do you think?

Valodim:
Sounds good to me.

I also planned to add an API call to import keys, although it would be nicer if we had working origin management (i.e. remember where a key was originally obtained from)  by the time we did that.

Valodim:
closing stale PR. the conflicts don't look too bad though, please reopen if this is still relevant

PR #1884, opened 2016-06-02
WIP: Migration to symmetrically encoded key blocks for secret keys

alexfjw:
Migration away from s2k to symmetric secret key blocks

TODO:
- [x]     Modify consolidate to write secret key related data directly to db, instead of importing secret keys*
- [x]     Refactoring of provider helper (remove remnants from api level 11)
- [x]     Ask for passphrases when importing
- [x]     Modify secret key related operations (sign, decrypt, new id, etc)
- [x]     Modify new key creation
- [x]     Modify secret key export
- [x]     Migrate old secret keys to symmetric key blocks

*Current consolidation requires reimporting secret keys. This would not work with symmetric key blocks as they would have to be decrypted, which means we have to ask for passphrases on each consolidate. 
- [x]       Refactoring
- [ ]       Rebasing on Andrea's work to add in the change in key import
- [ ]       Using the key import interface for migration

Targets for refactoring:
1) PassphraseDialogActivity (is in a somewhat disorderly state, as it has to handle key s2k keys, sym encrypted keyrings, both types of backup codes)
2) ProviderHelper (has grown quite greatly in size) 

Both the single & multi-passphrase workflows have been added, 
along with an applock which works with the master passphrase. 

There is probably a need to refine the UI for both migration and a fresh install of the application.
However, it is definitely usable in its current state, with support for all pre-existing features before this PR.


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/ParcelableEncryptedKeyRing.java
```diff
@@ -0,0 +1,81 @@
...
+import android.os.Parcelable;
+import android.util.Pair;
+
+import java.util.ArrayList;
+
+/**
+ * Parcelable representation of an encrypted keyring block as raw data,
+ * with related fields not obtainable without decrypting the block
+ */
+public class ParcelableEncryptedKeyRing implements Parcelable{
```
adithyaphilip: code style ;)

AlexFJW: Woops I should fix that. Sounds pretty confusing now that I read that comment ><

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/KeyringPassphrases.java
```diff
@@ -0,0 +1,83 @@
...
+import android.os.Parcelable;
+import android.util.Pair;
+import org.sufficientlysecure.keychain.KeychainApplication;
+
+import java.util.ArrayList;
+
+public class KeyringPassphrases implements Parcelable {
+
+    public final long mMasterKeyId;
+    public final ArrayList<Pair<Long, Passphrase>> mSubkeyPassphrases;
```
AlexFJW: Changing this to hashmap

Valodim: we have a ParcelableHashMap somewhere

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
```diff
@@ -869,6 +871,31 @@ private Intent checkRequirements(Intent data) {
...
+        result = databaseIsUpToDateOrReturnIntent(data);
+        if (result != null) {
+            return result;
+        }
+
+        return null;
+    }
+
+
+    private Intent databaseIsUpToDateOrReturnIntent(Intent data) {
```
Valodim: I like this idea! Always wondered how to do this nicely, but returning a pendingintent which somehow leads the user to the consolidation is a good idea

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/KeyringPassphrases.java
AlexFJW: Yep found it. Thanks!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -24,29 +24,35 @@
...
 import org.sufficientlysecure.keychain.service.ImportKeyringParcel;
+import org.sufficientlysecure.keychain.service.input.CryptoInputParcel;
+import org.sufficientlysecure.keychain.service.input.RequiredInputParcel;
 import org.sufficientlysecure.keychain.ui.base.BaseActivity;
 import org.sufficientlysecure.keychain.ui.base.CryptoOperationHelper;
 import org.sufficientlysecure.keychain.ui.util.KeyFormattingUtils;
 import org.sufficientlysecure.keychain.ui.util.Notify;
-import org.sufficientlysecure.keychain.util.Log;
-import org.sufficientlysecure.keychain.util.ParcelableFileCache;
+import org.sufficientlysecure.keychain.util.*;
```
Valodim: no * imports

AlexFJW:
Regarding failing tests:
2 tests relating to keyring retrieval currently fail as I'm still working on it.

Regarding encoding of keyring blocks: 
A 256 bit encryption key is generated using pbkdf2 with HMAC SHA1, and the keyring is encrypted with AES-CBC. I'm not quite used to working with crypto algorithms directly so do pardon my mistakes if any. 

Code for routines can be found at OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/ByteArrayEncryptor.java

The salt and IV are stored together with the encrypted keyring data through serialization, and the entire serialized object is saved into the db as a blob. We could also store the salt and IV in the database directly in their own columns but I thought that serialization might make the code easier to read. 

I do feel that the encryption routines I wrote are kinda flaky, but I'm gonna leave this for later.

adithyaphilip:
In the migration system's current state, if the user has forgotten the password for one of his keys, the app becomes unusable until he clears data. Would it be possible to make the migration optional, or allow it to be skipped for some keys (with the skipped keys being rendered unusable until migrated)?

AlexFJW:
I was planning on using an interface similar to what we have for key
import, where the user is able to select which key he wants. Will be
working on that part on a later date, perhaps when Andrea is done with the
"import all" UI.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -519,9 +500,7 @@ private PgpEditKeyResult internal(PGPSecretKeyRing sKR, PGPSecretKey masterSecre
             log.add(LogType.MSG_MF_UNLOCK, indent);
             {
                 try {
-                    PBESecretKeyDecryptor keyDecryptor = new JcePBESecretKeyDecryptorBuilder().setProvider(
-                            Constants.BOUNCY_CASTLE_PROVIDER_NAME).build(cryptoInput.getPassphrase().getCharArray());
-                    masterPrivateKey = masterSecretKey.extractPrivateKey(keyDecryptor);
+                    masterPrivateKey = masterSecretKey.extractPrivateKey(null);
```
adithyaphilip: Could you put in a comment explaining where the unlocking is happening now?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1192,81 +1139,38 @@ private PgpEditKeyResult internalRestricted(PGPSecretKeyRing sKR, SaveKeyringPar
...
-            return new PgpEditKeyResult(PgpEditKeyResult.RESULT_ERROR, log, null);
-        }
-
         log.add(LogType.MSG_MF, indent,
                 KeyFormattingUtils.convertKeyIdToHex(wsKR.getMasterKeyId()));
         indent += 1;
         progress(R.string.progress_building_key, 0);
 
-        // We work on bouncycastle object level here
         PGPSecretKeyRing sKR = wsKR.getRing();
```
adithyaphilip: `sKR` actually clashes with our naming convention for static variables (starts with `s` - [variable naming](https://source.android.com/source/code-style.html#follow-field-naming-conventions)), but I see it's what's been used everywhere else in this class, so I guess it's okay.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1192,81 +1139,38 @@ private PgpEditKeyResult internalRestricted(PGPSecretKeyRing sKR, SaveKeyringPar
...
-                sKR = applyNewPassphrase(sKR, masterPublicKey, cryptoInput.getPassphrase(),
-                        changeUnlockParcel.mNewPassphrase, log, indent);
-                if (sKR == null) {
-                    // The error has been logged above, just return a bad state
-                    return new PgpEditKeyResult(PgpEditKeyResult.RESULT_ERROR, log, null);
-                }
-            } catch (PGPException e) {
-                throw new UnsupportedOperationException("Failed to build encryptor/decryptor!");
+        try {
+            sKR = applyNewPassphrase(sKR, masterPublicKey, new Passphrase(), newPassphrase, log, indent);
```
adithyaphilip: `sKR` actually clashes with our naming convention for static variables (starts with `s` - [variable naming](https://source.android.com/source/code-style.html#follow-field-naming-conventions)), but I see it's what's been used everywhere else in this class, so I guess it's okay.

AlexFJW: Yep, thought that it'll be better to just stick to the convention already present.

It seems that we use pretty strange naming in some tests as well. 
mXXX for static variables. Left that alone for the time being.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1192,81 +1139,38 @@ private PgpEditKeyResult internalRestricted(PGPSecretKeyRing sKR, SaveKeyringPar
 
     }
 
-
-    public PgpEditKeyResult modifyKeyRingPassphrase(CanonicalizedSecretKeyRing wsKR,
-                                                    CryptoInputParcel cryptoInput,
-                                                    ChangeUnlockParcel changeUnlockParcel) {
+    public PgpEditKeyResult addS2kToSubKeys(CanonicalizedSecretKeyRing wsKR,
```
adithyaphilip: As discussed, might be nice to mention that the function expects a fully decrypted `CanonicalizedSecretKeyRing`.

AlexFJW:
Just a note, I have not amended the 'no password' indicator which shows up when a user tries to examine his subkeys. Was using it as an indication that stripping s2k was working.

Will be hiding that in a future commit.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/ByteArrayEncryptor.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.security.spec.KeySpec;
+
+public final class ByteArrayEncryptor {
```
dschuermann: Is there a reason why you implement your own encryption method for byte arrays? Why not use our PGP methods for this? If there is no specific reason, I would propose switching to PGP encryption.

dschuermann:
And please merge/rebase master. Some changes shown in your PR by @runnerway are now on master and shouldn't be displayed here as changes.

AlexFJW:
I think something went wrong when I was rebasing. Will fix that in a bit.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/ProviderWriter.java
```diff
@@ -0,0 +1,1203 @@
...
+        int result = SaveKeyringResult.SAVED_PUBLIC;
+
+        long masterKeyId = keyRing.getMasterKeyId();
+        UncachedPublicKey masterKey = keyRing.getPublicKey();
+
+        ArrayList<ContentProviderOperation> operations;
+        try {
+
+            mProviderHelper.log(LogType.MSG_IP_PREPARE);
+            result += 1;
```
adithyaphilip: I think this used to be ```mIndent+=1``` before it was copied over. This also highlights an issue with logging in `ProviderWriter` and `ProviderReader` in general - I'm not very sure whether accessing the ```ProviderHelper```'s operation log this way is ideal (getting it to work would involve changing `ProviderHelper.mIndent` too).

adithyaphilip:
Just got a crash while trying to update a private key, which was imported from a file.
```
08-19 03:55:41.365 2600-2848/org.sufficientlysecure.keychain.debug E/Keychain D: A key could not be imported during multi-threaded import
                                                                                 java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: passphrase is null
                                                                                     at java.util.concurrent.FutureTask.report(FutureTask.java:93)
                                                                                     at java.util.concurrent.FutureTask.get(FutureTask.java:163)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation.multiThreadedKeyImport(ImportOperation.java:506)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation.execute(ImportOperation.java:458)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation.execute(ImportOperation.java:86)
                                                                                     at org.sufficientlysecure.keychain.service.KeychainService$1.run(KeychainService.java:155)
                                                                                     at java.lang.Thread.run(Thread.java:818)
                                                                                  Caused by: java.lang.IllegalArgumentException: passphrase is null
                                                                                     at org.sufficientlysecure.keychain.provider.ProviderReader.getCanonicalizedSecretKeyRingHelper(ProviderReader.java:194)
                                                                                     at org.sufficientlysecure.keychain.provider.ProviderReader.getCanonicalizedSecretKeyRing(ProviderReader.java:184)
                                                                                     at org.sufficientlysecure.keychain.provider.ProviderWriter.savePublicKeyRing(ProviderWriter.java:792)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:328)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation.access$000(ImportOperation.java:86)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:497)
                                                                                     at org.sufficientlysecure.keychain.operations.ImportOperation$1.call(ImportOperation.java:485)
                                                                                     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                                                                                     at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:422)
                                                                                     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                                                                                     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
```

AlexFJW:
Sure, I'll look into that.

adithyaphilip:
Looks like line 792 in `ProviderWriter` is missing a `passphrase == null` check (the comment before it described what action to take if there is no passphrase available).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
```diff
@@ -70,6 +75,12 @@
         String API_ALLOWED_KEYS = "api_allowed_keys";
     }
 
+    private static final String CREATE_MASTER_PASSPHRASE =
+            "CREATE TABLE IF NOT EXISTS master_passphrase ("
+                    + MasterPassphraseColumns.ROW_INDEX + " INTEGER PRIMARY KEY, "
```
adithyaphilip: As discussed, this column is redundant.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/ProviderWriter.java
AlexFJW: Any suggestions on how I should go about fixing the logging issue?

dschuermann:
Somehow my comment is gone regarding ByteArrayEncryptor:
Why do you implement a new encryption routine here? Why not use our OpenPGP encryption routine? Are there any special considerations I don't see currently?

AlexFJW:
My bad, should have happened because of edits to that file. Decided to
finish up essential features for my pr first. Am planning on working on it
soon.

(I didn't use the OpenPGP routines as I didn't quite understand them
actually, will give it another go)

adithyaphilip:
Also, opting out of master passphrase in preferences has a few issues:
* Secret keys retain master password as their respective individual passwords. Should we should ask the user for new passwords for each of his keys?
* [Bug] Opting in again still requires the old master password to be remembered and entered.

Considering these issues, it might be better to not have the preference for now.

AlexFJW:
Thanks Adithya, I"ll look into the bugs in a bit.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -131,6 +125,76 @@ public void setFirstTime(boolean value) {
         editor.commit();
     }
 
+    public boolean isUsingEncryptedKeyRings() {
```
adithyaphilip: Might be nice to add a comment explaining what this means.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -131,6 +125,76 @@ public void setFirstTime(boolean value) {
...
+        return mSharedPreferences.getBoolean(Constants.Pref.USING_ENCRYPTED_KEYRINGS, true);
+    }
+
+    public void setUsingEncryptedKeyRings(boolean value) {
+        SharedPreferences.Editor editor = mSharedPreferences.edit();
+        editor.putBoolean(Constants.Pref.USING_ENCRYPTED_KEYRINGS, value);
+        editor.commit();
+    }
+
+    public boolean isAppLockReady() {
```
adithyaphilip: Again, a comment explaining what it does (as opposed to the `useApplock` preference - add a comment there too) would be helpful.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -131,6 +125,76 @@ public void setFirstTime(boolean value) {
...
+        return mSharedPreferences.getBoolean(Constants.Pref.HAS_MASTER_PASSPHRASE, false);
+    }
+
+    public void setHasMasterPassphrase(boolean value) {
+        SharedPreferences.Editor editor = mSharedPreferences.edit();
+        editor.putBoolean(Pref.HAS_MASTER_PASSPHRASE, value);
+        editor.commit();
+    }
+
+    public boolean isPartiallyMigrated() {
```
adithyaphilip: Explaining what states `isPartiallyMigrated` and `isMidwayChangingPassphraseWorkflow` in the code would be nice.

AlexFJW:
About  "[Bug] Opting in again still requires the old master password to be remembered and entered."

We would still require the master passphrase if want to provide the user to have the option of using the applock. That was the reason I coded it as such. I suppose the alternative would be to prevent applock usage if on the multi-passphrase workflow.

adithyaphilip:
When importing a secret key from file after opting into master passphrase, the associated public key is imported and appears in the public key list, but the secret key is not.
Possibly relevant stacktrace:
```
08-26 20:58:06.453 30196-31503/org.sufficientlysecure.keychain.debug E/Keychain D: Fatal bug in code, the master passphrase we tried is incorrect
                                                                                   org.sufficientlysecure.keychain.provider.ByteArrayEncryptor$IncorrectPassphraseException
                                                                                       at org.sufficientlysecure.keychain.provider.ByteArrayEncryptor.decryptByteArray(ByteArrayEncryptor.java:105)
                                                                                       at org.sufficientlysecure.keychain.provider.ProviderReader.getMasterSecretKey(ProviderReader.java:164)
                                                                                       at org.sufficientlysecure.keychain.provider.ProviderWriter.saveCanonicalizedSecretKeyRing(ProviderWriter.java:665)
                                                                                       at org.sufficientlysecure.keychain.provider.ProviderWriter.saveSecretKeyRingHelper(ProviderWriter.java:1030)
                                                                                       at org.sufficientlysecure.keychain.provider.ProviderWriter.saveSecretKeyRing(ProviderWriter.java:969)
                                                                                       at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:324)
                                                                                       at org.sufficientlysecure.keychain.operations.ImportOperation.serialKeyRingImport(ImportOperation.java:117)
                                                                                       at org.sufficientlysecure.keychain.operations.ImportOperation.execute(ImportOperation.java:443)
                                                                                       at org.sufficientlysecure.keychain.operations.ImportOperation.execute(ImportOperation.java:86)
                                                                                       at org.sufficientlysecure.keychain.service.KeychainService$1.run(KeychainService.java:161)
                                                                                       at java.lang.Thread.run(Thread.java:818)
```
Debugging Line 103, `ByteArrayEncryptor.java` will show the cause to be an `Exception creating cipher` resulting from an `EOFException`. Hope that helps.

adithyaphilip:
@AlexFJW Great work fixing the import bug. Works flawlessly now. Just to nitpick, the passphrase dialog asks for the master passphrase while importing keys but does not cache it. Could you fix that?

adithyaphilip:
Upgrading from master also works great. The only potential issue that remains is if a user has a secret key which he has forgotten the password to, he will have to clear data to proceed. Maybe the option to select which keys will be migrated (with the rest deleted), you had mentioned earlier would be a good solution.

AlexFJW:
I'll look into the caching issue this weekend.

As for the secret key selection for migration, was thinking of waiting till
Andrea's changes are merged into master.

Could try and do it now with the current import key ui but having to fix it
up when the changes come in doesn't feel very appealing.

Valodim:
closing stale PR

PR #1885, opened 2016-06-02
Improve ImportKeysListFragment

runnerway:
I've started working on improving various parts of the UI of the app.

This is a first step regarding ImportKeysListFragment. 
I've in mind other improvements but i think it's already better than before.
Other things will come soon. (for both UI and code)

(I'm sorry for the language of the screenshots)


Valodim:
Neat, data binding. Not something we used before but I'm okay with that. If you want to go software patterns, I'm quite the fan of MVP.

On the UI design, is this based on some other app, or material guidelines? The key container widgets look like tiles (not round like cards), but they aren't uniform in size? I'm also not sure if the selection status is very clear, especially if there is only a single result it's not clear that the user must click the key, which subtly changes colors, and then import selected.

Valodim:
As another thought: Users usually want to import a single key, so we might want to move away from the "import all selected" pattern, towards a "click for details -> (download key and) display details -> import" one.

dschuermann:
While we should import only single keys from keyservers, import from files should still allow multi-select IMHO.

runnerway:
@Valodim Yes. I really like data binding. I think that it's a very clean and effective way to implement some things. I plan to use them in many places. (Obviously only where it's better than a standard solution..)

(They are card very similar to ones of Google Now. Maybe equal. I've tried to follow the Material Design's  guidelines for conventions, aspect and so on.. Anyway i agree with you that selection status isn't very clear..but i've never seen checkboxes on cards. But we "have" to use cards because guidelines tell us to use cards for text larger than 3 lines.. IMHO they are also prettier than a list)

@dschuermann I've spent a lot of time thinking about this. Some times i think it's better to keep things unified for consistency, other times i think it's better to separate things to improve usability of certain operations..


A solution can be doing things like Valodim has said (and keeping things unified), but add an option in the menu for "Import all". Probably it's not very common to import a number of keys > 1 but < than "all". If we agree, i will go for such a solution.

This would be also a solution for the selection status..

Valodim:
those corners look pretty square to me, cards have rounded corners. also, the guidelines explicitly state that homogeneous data should not be displayed in cards but in a list, to be quickly scannable. It's debatable whether this applies for us, but if we diverge from the guidelines we should do so consciously 

https://material.google.com/components/cards.html#cards-usage

runnerway:
Yes, in this case guidelines are a bit contradictory:
- "If more than three lines of text need to be shown in list tiles, use cards instead"
https://material.google.com/components/lists.html
- "They’re also well-suited for showcasing elements whose **size** or supported actions vary.."
https://material.google.com/components/cards.html

The second point is what convinced me. Having a list with items with different heights was a bit strange for me. But this is only my opinion. (If you want i can return to using the list..)

Anyway this can be a rough plan:
- [ ] Remove the bottom toolbar
- [ ] Add option to "Import all" in the menu
- [ ] Add ability to see details before importing via an intermediate view
- [ ] Add 2 action to each card: to reach the details, to import directly 
(Details view contains an action to import the key..).
- [ ] Improve the top textview (maybe it should be incorporated to the action bar..)
..
- [ ] Start adding support for DANE.. (But this is another topic)

Are we missing something?

Cards will be something like:


Maybe we can move some informations in a hidden part of the card like this:

(The point here is: what information we can hide?)

Valodim:
I think I actually prefer cards, info on people in a "business card" way should be familiar to everyone.

My current take would be to show on the cards only the most basic information, which the user needs to distinguish keys and choose which one they want to see details on: name, creation date, type of key, possibly revocation status. this can be neatly shown in a fixed height for almost all keys, i.e. all cases where there is only a single name. For cases where there is more than one name, we can add an "a.k.a." line or something like that.

After the user clicks on a card (or an expand button, but I think that can just be a whole-card action), it can download the key, then expand to display more info.

Couple more thoughts and reasoning on the matter:
* info we get on keys from keyservers is unreliable: we can never know which uids are revoked, which are primary, and which are just plain bogus before we actually download and parse the key
* neither do we know the confirmation status
* for the "single result" case, we could expand the card right away?
* I'm not sure if "expanding card" is better than a full intermediate activity (or dialog), so this should best be implemented in a way where we can easily move around at a later point. I would much prefer a MVP implementation over fragments here, because those get really messy if you want a dynamic number of them.

runnerway:
I think it is more than enough to continue the work on this PR.

* Ok for fixed height cards, "business card" is a good analogy
* Ok for single result already expanded, only if we choose "expanding" card  
 (I think an already opened dialog it's not the best thing but we will see..)
* Ok for MVP :)

And thank you for the other thoughts.

From tomorrow i will start following all the points we highlighted so far. I hope the final result will be good. Probably the initial implementation will use "expanding" cards.

runnerway:
This is a view of the current situation:



**Key ID**, **Creation date** and **User IDs** are taken from the downloaded and parsed key. This is useful in many cases. For example, if we want to know, before importing, the user ids contained in a key found on Keybase. Probably this will be useful also for OpenPGP keys via DANE.

Key is downloaded and parsed (but not saved) when we expand the card.
In this way we don't need to rely only on information retrieved from keyserver.

By the way, it's easy to change what information we want to display on this screen. (Probably i will add some bits to check that the key status obtained from keyserver corresponds to the real one).
If you have other advices, please tell me :smile:

I consider this PR nearly finished. Remaining things are:

- [ ] If the user locks his device while it was in this screen, when he unlock the phone, a new query to keyserver is made (this shouldn't be caused by this PR, but i will try to see what can i do to avoid this behavior)
- [ ] I need to check the flow for "Ask permissions in Android 6" -> "Start loading file containing keys..". Probably i need to restore some bits i've lost during refactoring
- [ ] A bit of refactoring (there will be always something to refactor..)
- [ ] Ability to interrupt download of a key (useful when a download seems stuck). I will try to add this feature.

During the development i've tried to produce clean code while improving the existing code base with refactoring and fixing bugs i've discovered on my path. I've made extensive use of data binding. In this way, it will be easy moving things around, if we decide to change in the future.

I will continue improving this PR.

dschuermann:
@runnerway Looks great.
Without looking into the code: What's the status regarding import from file of pub and secret keys?

runnerway:
Currently, import from file is nearly the same as import from keyserver. When the card is expanded, the key is canonicalized and the same information is displayed. (Maybe for 'import from file' this is less useful?). What's you opinion on this? Do you prefer another behavior or another set of information to be displayed? 

I don't have access to a computer until Sunday. But from that day I will resume the work on this and other things :)

dschuermann:
@runnerway Import from file should allow to import all keys at once. Either by selecting keys or other ways, this is especially important when restoring a backup. I think @Valodim already had some UI for this in mind.

runnerway:
I've already included the ability to import all the keys when importing from file. There is an option in the contextual menu which is hidden when importing from keyserver. :)

dschuermann:
When importing from a file, the default should be to import all keys and importing individual keys should be the "advanced" way. Can you show a screenshot here what it currently looks like?

dschuermann:
And a small nitpick: I would propose we use the OpenKeychain green for the text color of the IMPORT button. It makes it more obvious that it's clickable.

runnerway:
Ok. Now i've access to my PC. Sorry for the delay.

Yes. No problems about green for text color. This is the sort of feedback i'm looking for, to finish this PR.
About import from a file. I agree with you. For the screenshot, it's exactly the same thing as importing from keyserver. (But i want to change it like written below..)

The plan that i want to follow in next days (probably 2-3 days) is:
- [ ] Change text color of the button
- [ ] Add a "basic" way to import keys: without import button and "expand feature". Only a list of cards with primary uid and status. Add a button to import all (Probably in a bottom action bar or fab). Keep the current way as "advanced".
- [ ] Set "basic"as default for import from file. And "advanced" from import from keyserver. Switch from "basic" to "advanced" is possible from an entry in the context menu
- [ ] (Optional) Move text field for search, the buttons near it, buttons about import from file etc..inside the action bar for an effect like this: http://i.imgur.com/zo5LTX3.gif

Alright? Advices? :smile:

dschuermann:
@runnerway sounds good to me. Especially the last TODO sounds great to have a better Material-like interface for these activities.
Regarding the placement of an "import all" button: I prefer one in a bottom bar in this case.

dschuermann:
Another Thing I would really like to see: Tests for keyserver, keybase, http, Facebook. Maybe for Starters simply use sone Well known addresses on the Internet for These Tests.

runnerway:
Rebased, applied changes that we discussed before and some fix/refactoring.

This is a view of the current situation:

 

Left:
- Basic way (default) for import from file. User can switch to advanced (and switch back to basic) via menu.

Right:
- Advanced way (default) for import from keyserver. User can't switch to basic because it makes no sense for import from keyserver.

Is it ok? Does it need some changes? I'm not a native english speaker, are there better words to describe "Basic way" and "Advanced way"?
&nbsp;
&nbsp;
&nbsp;
Ok for tests. At this point i can work on three things:

1.  Add ability to interrupt download of a key: useful when import or preview seems stuck because of network problems/delay/etc.. and we don't want to wait the timeout
2.   Improve activities with Material-like interface
3.  Add tests about the things you've asked

Do you have  any particular preference for the order in which these things should be done?

dschuermann:
Advanced is "Select keys individually". I am also not totally sure if the overflow menu is a good place for this to find.

The order I would prefer is 3,2,1 :)

Valodim:
I think I neither like the overflow menu, nor the button at the bottom very much. I'm also not sure the list of keys in "basic" view is even needed?

How about just showing a card with content like "This file contains 23 keys", together with two buttons for "import all" and "list keys", the latter of which then displays the cards for all keys. That way, we avoid the overflow menu and the bottom button at once, while keeping the interface coherent.

runnerway:
@Valodim Good idea. :)
Tomorrow i will provide an implementation of it.

runnerway:
Ok. With latest commits i've changed things for import from file like @Valodim has said.
Additionally, i've moved action buttons inside action bar: we gained space and i think the final result is aesthetically better. If you think now it is good, i will finish the work with code cleanup and re-adding little things like word completion for key search etc..

Screenshots:

- **Import from File**: 

 

- **Import from Cloud**: 



*(Using SearchView like we have in MainActivity)*

dschuermann:
Looks great, how are the tests progressing?

nitpick: I would rename the button "IMPORT KEYS" to "IMPORT ALL KEYS"

dschuermann:
Nice UI iterations! @Valodim and me discussed the design and would like to propose some changes (again):

* To make the design less cluttered and more usable, remove the buttons per card in the list, make card clickable which should expand the card. "Import" button is part of the expanded card. People who want to import a key normally also want to see the details before importing.
* "list keys" -> "view list"
* for files with only one key -> show single key card
* Show verified or unverified icons (as seen in our main key list) on cards with keys which are already in OK's database. Rename "Import" to "Refresh" button, + additional "show key" button for these cards
* Change "3 keys found" to "File contains 3 keys. 2 new keys"
* SearchView should be shown automatically (if it not already does this) and never close search

runnerway:
*(I've edited your comment to add numbers so i can reply to each point)*

1. OK. Good idea.
2. OK
3. OK. (I think that in this case card should be already expanded..)
4. OK. Good idea.
5. OK. I am only against the word "File". By using this word, we will be a bit limited if we want to reuse this view in similar cases. Is "Found 3 keys (2 new)" ok?
6. OK

dschuermann:
Great!

3. Yup, good idea to already expand it in this case
5. Yup, better not mention "file"

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/ImportKeysListEntry.java
```diff
@@ -82,6 +88,10 @@ public String getKeyIdHex() {
         return mKeyIdHex;
     }
 
+    public long getKeyId() {
+        return new BigInteger(mKeyIdHex.substring(2), 16).longValue();
```
dschuermann: Better put this in one of the util classes, I think https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/util/KeyFormattingUtils.java already contains similar methods.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/adapter/ImportKeysAdapter.java
```diff
@@ -70,6 +70,11 @@ public void setData(List<ImportKeysListEntry> data) {
         for (int i = 0; i < mKeyStates.length; i++) {
             mKeyStates[i] = new KeyState();
         }
+
+        // If there is only one key, get it automatically
+        if (mData.size() == 1)
```
dschuermann: use brackets per coding style

dschuermann:
Hey @runnerway 
What are the remaining TODOs left to be done for this PR to be merged? What's the current status?

dschuermann:
@runnerway ping

runnerway:
Sorry for the delay. I'm not disappeared. I'm just very busy these weeks. 
Anyway, i will try to add the latest needed commit during next week. 

The only thing remaining is to set the correct icon for keys already downloaded.

dschuermann:
@runnerway great, looking forward to it.

dschuermann:
I will now take over this PR to get it merged.

dschuermann:
I am working on it here: https://github.com/open-keychain/open-keychain/commits/import-refactor

dschuermann:
This is now on master! Nice to haves are now in https://github.com/open-keychain/open-keychain/issues/1954

saleemrashid:
This broke smart card support :cry: since it requires the activity to implement `ImportKeysListener` which is implemented by `ImportKeysActivity`. Smart card support uses the activity `CreateKeyActivity`.

```
java.lang.ClassCastException: org.sufficientlysecure.keychain.ui.CreateKeyActivity@xxxxxxx must implement ImportKeysListener
```

dschuermann:
@saleemrashid Thanks for your report. I created an issue at https://github.com/open-keychain/open-keychain/issues/1959

PR #1886, opened 2016-06-08
Instrumented unit tests: fix for execution

runnerway:
I've spent quite some time trying to understand the reason for a lot of proguard errors while trying to start instrumented unit tests. This PR contains an edit to gradle that is necessary to fix the problem. (Probably this is needed because of an upgrade of gradle)

Trying to execute instrumented unit tests, i saw that a lot of them are failing. (Probably because of updates in the code of the app).

If you agree, i will stop a bit from my main task, to spend some time trying to fix all the error in these tests. In this way we will have a good testing enviroment for future developments (mine included).

### DIFF
--- a/OpenKeychain/src/androidTest/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceTest.java
+++ b/OpenKeychain/src/androidTest/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceTest.java
@@ -1,10 +1,8 @@
 package org.sufficientlysecure.keychain.remote;
 
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-
 import android.app.PendingIntent;
+import android.content.Context;
 import android.content.Intent;
 import android.os.IBinder;
 import android.support.test.InstrumentationRegistry;
@@ -20,6 +18,9 @@ import org.junit.runner.RunWith;
 import org.openintents.openpgp.IOpenPgpService2;
 import org.openintents.openpgp.util.OpenPgpApi;
 import org.sufficientlysecure.keychain.R;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 
 import static android.support.test.espresso.Espresso.onData;
 import static android.support.test.espresso.Espresso.onView;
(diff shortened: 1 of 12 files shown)


runnerway:
I've spent a big part of the afternoon to fix "OpenPgpServiceTest".
In the end the main problem was that the test hasn't waited the start of the activities.

I've searched possibile solutions but all of them had problems. For now we use "Thread.sleep". (It's ugly)
Probably a better solution is to implement "IdlingResource" for the PendingIntets.. I will see what's the best thing, but if anyone has other advices they are welcome.


Another related question:
- Why this kind of tests are in this state? (A bit broken and old..)
It's because of lack of time/inadvertence.. or because we think they are not needed?
By the way, i've planned to fix/improve them during next days..

OpenKeychain/src/androidTest/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceTest.java
```diff
@@ -118,14 +116,11 @@ public void testStuff() throws Exception {
             PendingIntent pi = result.getParcelableExtra(OpenPgpApi.RESULT_INTENT);
             pi.send();
 
+            Thread.sleep(1 * 1000); //ugly
```
dschuermann: please comment this properly why this is needed here :)

dschuermann:
It's great to see some love for these tests. I think they are neglected due to time constraints and because we haven't found a reliable way to execute them on Travis. It would be great to get this working.

runnerway:
Thank you for the information.
I'm not an expert of Travis, but maybe i will give a look to see if i can do something. 

I will try to finish fixes for tests as soon as possibile. 
Then i would like to increase their coverage, starting from parts of the app on which i'm working now.

Valodim:
I would *love* to see some maintenance on these tests. The use of androidTests died down for two reasons:

* It's not easy to integrate into our CI chain, moving away from travis because using an emulator there just doesn't cut it. I gave GreenhouseCI a shot at some point, it failed but I don't remember why exactly.
* I was unable to fully eliminate non determinism. Espresso watches for an idle UI thread, and there are some cases where that doesn't work correctly. For example, a click is sometimes registered as a long click, and indeterminate progress bars in the background (during passphrase input) block tests indefinitely.

runnerway:
Thank you for the information.

1) After a fast search, it seems there are some news about the emulator on Travis. I will try to investigate later. 
2) This is interesting. Anyway i would like to try to see if i'm able to solve some problems. It would be great to have at least the current set of tests in a good conditions and i would like to use androidTests for the part of the app on which i'm working. (It would be a nice opportunity to learn such things)

dschuermann:
@runnerway can you add a comment regarding the Thread.sleep that we can merge this PR?

Valodim:
Also, any news on the Travis thing?

runnerway:
@dschuermann Done :)

@Valodim I'm going to setup a personal repository with a minimal Android project to try something like this: https://docs.travis-ci.com/user/languages/android#How-to-Create-and-Start-an-Emulator

runnerway:
Please, wait to merge this PR.

After 30 Travis builds i was able to find a working configuration for androidTests.
I need to clean up things. Then i will push the needed edits .

Now, for a Travis build, it takes around 20 minutes but i plan to optimize it. 

30 buils took around 10 hours. What a pain.. 
Now i need to use androidTests extensively to give to all the time spent on this thing a meaning :sweat_smile: ..pff

dschuermann:
@runnerway Is it somehow possible to start two Travis instances? One for unit tests which are fast and one for androidTests? I saw something like this on Gitlab for the F-Droid project. But they don't use Travis :/

runnerway:
I will take a look. Anyway, in my experience, running a build on Travis takes nearly the same time for unit tests vs androidTests because most of the time is spent building code. (But i need to check).

I think we can run unit tests in parallel with andoridTests lowering the total time, achieving something around 10 minutes, which i think is acceptable.

But this will be a further step. Before i want to push the changes needed for running andoridTests.
Then  i will improve them.

dschuermann:
@runnerway I really like that we run two Travis tests, good job! Now we only need to find a reliable way to execute the androidTests ;)

runnerway:
Ok. I've run android tests with latest commit for more than 10 times without problems.
I think they are reliable now.

In the end we obtained:
- around 6 minutes for unit tests
- around 15 minutes for android tests

Probably we will be able to optimize them but i think this situation is acceptable for now.
Yes. Separate jobs is a really great feature of Travis :smile:


(I've encountered some rare errors in unit tests. I'm going to open an issue)

OpenKeychain/src/androidTest/java/org/sufficientlysecure/keychain/ui/AsymmetricFileOperationTests.java
```diff
@@ -79,8 +79,8 @@
 import static org.sufficientlysecure.keychain.matcher.CustomMatchers.withSignatureNone;
 
 
-@RunWith(AndroidJUnit4.class)
-@LargeTest
+//@RunWith(AndroidJUnit4.class)
```
dschuermann: Please include a comment in all broken tests why they are disabled. I we merge this pr, people will be confused why some tests are commented out. At least put a line on top which states that this tests needs to be fixed with a TODO

OpenKeychain/src/androidTest/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceTest.java
```diff
@@ -79,23 +81,23 @@ public void testStuff() throws Exception {
             PendingIntent pi = result.getParcelableExtra(OpenPgpApi.RESULT_INTENT);
             pi.send();
 
+            Thread.sleep(ACTIVITY_WAIT_TIME * 1000); // Wait for activity to start
```
dschuermann: put the 1000 into the constant

dschuermann:
Thanks, great work! Merging this so we have at least the basis for new androidTests in master.

PR #1891, opened 2016-06-15
Signature message digest

Valodim:
This PR changes `PgpSignatureChecker` to obtain the computed message digest as an intermediate result, which is then placed in OpenPgpSignatureResult.

Unfortunately, I had to work around BouncyCastle a bit more than I would have liked. Most importantly, I ended up doing the PKCS1 DER formatting for RSA signatures manually which uh, sort of sucks. It seems to be the intended way though, since there is a comment "For raw RSA, the DigestInfo must be prepared externally" in DigestSignatureSpi.

The point of this is that this digest can be cached, and revisited at a later point. This is helpful internally because we can "refresh" the key status simply by running the verification again. It also means we don't have to decrypt and verify all data again after download if we didn't have a signing key available when we first processed the data.

The same applies for client apps. Decryption delay in K-9 is a serious problem, which is why I would like to cache some of the decryption results in K-9 at some point. However, just caching the decrypted data isn't enough, and I really wouldn't feel comfortable "caching" the signature status, since it can change over time (due to expiry or revocation).

### DIFF
--- a/OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPRawDigestContentVerifierBuilderProvider.java
+++ b/OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPRawDigestContentVerifierBuilderProvider.java
@@ -1,0 +1,153 @@
+package org.bouncycastle.openpgp.operator.jcajce;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.security.InvalidKeyException;
+import java.security.Provider;
+import java.security.Signature;
+import java.security.SignatureException;
+
+import org.bouncycastle.asn1.ASN1Encoding;
+import org.bouncycastle.asn1.ASN1ObjectIdentifier;
+import org.bouncycastle.asn1.DERNull;
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.asn1.x509.DigestInfo;
+import org.bouncycastle.bcpg.PublicKeyAlgorithmTags;
+import org.bouncycastle.jcajce.provider.util.DigestFactory;
+import org.bouncycastle.jcajce.util.DefaultJcaJceHelper;
+import org.bouncycastle.jcajce.util.NamedJcaJceHelper;
+import org.bouncycastle.jcajce.util.ProviderJcaJceHelper;
+import org.bouncycastle.openpgp.PGPException;
+import org.bouncycastle.openpgp.PGPPublicKey;
+import org.bouncycastle.openpgp.PGPRuntimeOperationException;
+import org.bouncycastle.openpgp.operator.PGPContentVerifier;
+import org.bouncycastle.openpgp.operator.PGPContentVerifierBuilder;
+import org.bouncycastle.openpgp.operator.PGPContentVerifierBuilderProvider;
+
+public class JcaPGPRawDigestContentVerifierBuilderProvider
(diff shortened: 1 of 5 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignatureChecker.java
```diff
@@ -255,26 +272,47 @@ boolean verifySignatureOnePass(Object o, OperationLog log, int indent) throws PG
...
+        PGPContentVerifierBuilder pgpContentVerifierBuilder = new JcaPGPRawDigestContentVerifierBuilderProvider()
+                .get(signingKey.getAlgorithm(), signature.getHashAlgorithm());
+        PGPContentVerifier pgpContentVerifier = pgpContentVerifierBuilder.build(signingKey.getPublicKey());
+        byte[] signedMessageDigest = digest.digest();
+
+        try {
+            OutputStream outputStream = pgpContentVerifier.getOutputStream();
+            outputStream.write(signedMessageDigest);
+        } catch (IOException e) {
+            e.printStackTrace();
```
dschuermann: re-throw as PGPException?

OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPRawDigestContentVerifierBuilderProvider.java
```diff
@@ -0,0 +1,153 @@
...
+                    DigestInfo dInfo = new DigestInfo(
+                            new AlgorithmIdentifier(digestIdentifier, DERNull.INSTANCE), bytes);
+
+                    return dInfo.getEncoded(ASN1Encoding.DER);
+                } catch (PGPException e) {
+                    throw new IOException("Unknown hashing algorithm!", e);
+                }
+
+            default:
+                return bytes;
```
dschuermann: the handling of this switch looks confusing to me since there is no break used. Is this style applied in other Bouncy Castle files?

Valodim: I don't know. A result variable with break would be four extra lines, do you think that would be more clear? I think I like this style better personally, but single point of return also has its benefits. No strong opinion either way.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignatureChecker.java
Valodim: :+1:

Valodim:
So um actually, I completely missed that to re-verify the digest, we need the signature data itself as well. It's available for detached signatures, but there are cases where it isn't, so not sure how to handle that elegantly other than to keep the serialized data in the signature result info... hmm.

(I still think it's nice to be able to verify raw hashes, and I never particularly liked the fragmented PGPSignature/PGPOnePassSignature control flow)

OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/JcaPGPRawDigestContentVerifierBuilderProvider.java
dschuermann: maybe add a ``// fall through`` comment at the end of the case. I have seen this in other projects to indicate that this is an exception to the typical ``break``. BTW Go has ``fallthrough`` :)

Valodim:
closing stale PR

Issue #1846, opened 2016-04-29
Use KeyRings.NAME etc instead of KeyRing.splitUserId


dschuermann:
Since we introduced KeyRings.NAME, KeyRings.EMAIL in https://github.com/open-keychain/open-keychain/pull/1763 we should now use this in more places to replace calls to `KeyRing.splitUserId`.

kritkaran94:
@dschuermann I am ready to work on this. Can you please elaborate as I am new to open-keychain.

dschuermann:
@kritkaran94 Please look into the source code of https://github.com/open-keychain/open-keychain/pull/1763/files . Now, name and email are extracted when writing a key to the database, thus we don't need to use ``KeyRing.splitUserId`` inside other UI classes etc.

kritkaran94:
@dschuermann I will get back to you soon on this.

kritkaran94:
@dschuermann I am referring a branch and starting to make changes. Please review it as I progress.

top-gun007:
is this issue fixed @dschuermann

daquexian:
I could work on this.

dschuermann:
Search for occurrences of ``KeyRing.splitUserId`` and carefully replace them by using the new database columns.

daquexian:
As for UncachedKeyRing, which is irrelevant to database, is it necessary to remove KeyRing.splitUserId? ( for example KeyRing.splitUserId in KeyItem(CanonicalizedPublicKeyRing ring) )

PR #1893, opened 2016-06-21
updated calls to KeyRings.Name & KeyRings.Email

kritkaran94:
In Progress of Issue #1846. If this seems headed in the correct direction. I would proceed further.

ccing @dschuermann  (I apologise in advance if it's not correct)


dschuermann:
does not compile, see travis

dschuermann:
Please, before making pull requests compile and test locally.

kritkaran94:
@dschuermann Sure I will do that and come up with a better one.

PR #1914, opened 2016-08-16
[WIP] Support for GNUK and Nitrokey Start

nmikhailov:
There are some difficulties with GNUK:
- can't be rested to default - in fact our procedure will most likely brick it
- keys can't be overriden(there is a tool to remove keys with admin PIN in gnuk sources, but I think that it out of scope for us)
- Some weird admin-less mode which messes our token initialization routine.
- [x] fix sign/decrypt
- [ ] fix(or disable) token initialisation for gnuk
- [ ] disable key move if its gnuk
- [ ] disable token reset if its gnuk

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
@@ -533,7 +533,8 @@ public class SecurityTokenHelper {
         // Receive
         while (lastResponse.getSW1() == APDU_SW1_RESPONSE_AVAILABLE) {
             // GET RESPONSE ISO/IEC 7816-4 par.7.6.1
-            CommandAPDU getResponse = new CommandAPDU(0x00, 0xC0, 0x00, 0x00, lastResponse.getSW2());
+            CommandAPDU getResponse = new CommandAPDU(0x00, 0xC0, 0x00, 0x00,
+                    lastResponse.getSW2() == 0 ? MAX_APDU_NE : lastResponse.getSW2());
             lastResponse = mTransport.transceive(getResponse);
             result.write(lastResponse.getData());
         }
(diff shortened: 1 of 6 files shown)


szszszsz:
Hi @nmikhailov !
How is it going with the PR? Could I help with testing?

Valodim:
Can you say more on why the GNUK can't be reset to defaults? Isn't that standard behavior of the javacard applet?

nmikhailov:
@Valodim GNUK is not javacard, its a firmware for microcontroller written in C.

jans23:
Gnuk 1.2 can be reset to factory defaults. Earlier Nitrokey Start can be updated to Gnuk 1.2 too.

jeremydrake-eAc:
It sounds like [Gnuk 1.2.5](https://anonscm.debian.org/cgit/gnuk/gnuk/gnuk.git/commit/?h=release/1.2.5&id=83414a747ae8f09639ecaf1d9a8c7e3988068c0f) fixed how it handles reset, so your token reset code should work with Gnuk 1.2.5 and later.

I don't really understand the other issues - can you explain more about what's wrong with token initialization, and what you mean by "key move"?

jeremydrake-eAc:
OK, I got the current master branch and merged these changes, and was able to confirm that token reset does work against Gnuk 1.2.5, however setting up a blank token fails.  I added logging to output every APDU transmitted and every response, and here is where it went wrong:
`V/Keychain D( 7100): Sending APDU {00da00c106010800001103}`
`V/Keychain D( 7100): Received APDU {6f00}`

Gnuk expects the algorithm attributes for RSA 2048 to be `06010800002000`, the differences being that the exponent should sent as a 32-bit value (despite still having the value 65537), and that it expects the 'standard' format rather than the 'crt with modulus' format.

dschuermann:
This pull request is superseded by https://github.com/open-keychain/open-keychain/pull/2201

@jeremydrake-eAc Thanks a lot for this detailed debugging information. This helped us a lot to fix GNUK issues ❤️

PR #1915, opened 2016-08-17
Fix travis android-24 dep

nmikhailov:


nmikhailov:
@dschuermann i remember that we specifically used openjdk instead of oracle, why was that the case?

nmikhailov:
Problem is that there is no openjdk8 on travis https://docs.travis-ci.com/user/languages/java/
Java8 is required by android-24

dschuermann:
I think OpenJDK was required to compile Bouncy Castle and get it to working as a JCA provider without signing it with a vendor key.

Valodim:
Indeed. The JCA implementation in Oracle JDK does not allow for unsigned JCA providers, so for the unit tests to work there is no way to use that as long as we patch anything in BouncyCastle

dschuermann:
I commented on one of there issues (https://github.com/travis-ci/travis-ci/issues/2968) but I there are many issues related to OpenJDK open on their issue tracker.

nmikhailov:
It works now.
Used openjdk8 from ppa, seems like its the only option now.

dschuermann:
Awesome!

PR #1920, opened 2016-08-22
auto-linkify text in DisplayTextActivity, if data is signed

Valodim:
Wanted to send a link and noticed that in our text view activity, links weren't auto-linkified.

There is some room for discussion here: At the moment, I only linkify text if it comes with a valid signature, both confirmed and unconfirmed. We could handle this more strictly (only valid+confirmed) or less strictly (all non-error states), there probably isn't a perfect decision here


dschuermann:
Great, I like this! I think it's okay to leave this to signed texts with unverified + verified keys.

PR #1923, opened 2016-08-23
Travis cleanup

nmikhailov:
~~Testing, ignore this PR for now~~
Actually wanted to use trusty so using ppa's won't be needed, but `language: android` doesn't work there :/ Anyway cleaned up a bit as a side effect of testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,6 @@ env:
 
 # force non-container build
 sudo: required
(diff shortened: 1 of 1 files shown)


dschuermann:
great, looks like this still works.

nmikhailov:
Yes, looks like these problems with hostname were fixed in openjdk8

Issue #1928, opened 2016-09-13
Physical keyboard causes crash


Matvei27:
If you try to navigate this app with the arrow keys on a keyboard or on an Android device with physical keys, it will crash. Usually happens on the button for the sidebar drawer.
Please fix this so I can use this app instead of the outdated APG app.

Valodim:
Did you send a play store crash report when you get the "unfortunately..." message? That would be very helpful for debugging

Matvei27:
It doesn't give me that option : (
I can only click OK.


On Sep 13, 2016, at 11:18 PM, Vincent Breitmoser <notifications@github.com> wrote:

Did you send a play store crash report when you get the "unfortunately..." message? That would be very helpful for debugging

Valodim:
Ah. Can you try to obtain the stack trace from logcat, using adb? That would help tremendously :)

Matvei27:
Sure, if you give me the commands I would need to input I can run ADB with the device in USB debug mode. 

On Sep 14, 2016, at 01:42 AM, Vincent Breitmoser <notifications@github.com> wrote:

Ah. Can you try to obtain the stack trace from logcat, using adb? That would help tremendously :)

Valodim:
`adb logcat` should do it

Paramjeet2810:
@Matvei27 : I tried using physical keyboard on various emulators and it works perfectly fine. So, could you please share with me about the device specifications on which the app crashes.

Matvei27:
KDDI SHF31 

On Sep 17, 2016, at 10:29 PM, Paramjeet S Desai <notifications@github.com> wrote:

@Matvei27 : I tried using physical keyboard on various emulators and it works perfectly fine. So, could you please share with me about the device specifications on which the app crashes.

Matvei27:
The app itself works, but navigation with the physical keys causes crashes. 

On Sep 17, 2016, at 10:29 PM, Paramjeet S Desai <notifications@github.com> wrote:

@Matvei27 : I tried using physical keyboard on various emulators and it works perfectly fine. So, could you please share with me about the device specifications on which the app crashes.

Paramjeet2810:
@Matvei27 I have understood your problem and will get back to you as soon as I solve the problem.

Matvei27:
Is the d-pad somehow different from the arrow keys on the keyboard? 
Based on the way they work with apps they APPEAR to behave the same, but maybe not? 

[...]

PR #1934, opened 2016-09-28
Physical keys cause app to crash solved.(Issue number 1928)

Paramjeet2810:
Physical key crash issue solved.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MainActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MainActivity.java
@@ -60,7 +60,7 @@ public class MainActivity extends BaseSecurityTokenActivity implements FabContai
 
     public Drawer mDrawer;
     private Toolbar mToolbar;
-    private int CURSOR;
+    private int cursor;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
@@ -170,8 +170,8 @@ public class MainActivity extends BaseSecurityTokenActivity implements FabContai
     public boolean onKeyUp(int keyCode, KeyEvent event) {
         if(mDrawer.isDrawerOpen()) {
             if (keyCode == 20) {
-                CURSOR = (6 + CURSOR + 1) % 6;
-                switch (CURSOR + 1) {
+                cursor = (6 + cursor + 1) % 6;
+                switch (cursor + 1) {
                     case ID_KEYS:
                         mDrawer.setSelection(ID_KEYS, false);
                         break;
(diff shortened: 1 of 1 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/MainActivity.java
```diff
@@ -59,6 +60,7 @@
 
     public Drawer mDrawer;
     private Toolbar mToolbar;
+    private int CURSOR;
```
nmikhailov: Check out coding style https://github.com/open-keychain/open-keychain#coding-style

dschuermann:
Can you explain why this fixes the issue #1928 . What I see is that you implemented a form of navigation via physical buttons.

Paramjeet2810:
Sir,
When I used the application with an OTG keyboard connected to my android
device, navigation through physical keys was no possible. Whenever I would
press any key the app wouldn't respond.
After adding the navigation via physical button it works perfectly fine.
After that I opened the app in an emulator which had a d-pad, the same code
worked even for the d-pad.

According to my understanding, the user had this problem that whenever he
used to use his physical keys to navigate in the Navigation Drawer as there
was no available functionality his app wouldn't function.
Regards,
Paramjeet Desai

nmikhailov:
Have you managed to reproduce the crash though? Original issue #1928 clearly states that hardware keys crash OKC.

Paramjeet2810:
No, I couldnt reproduce the crash exactly.

Paramjeet2810:
I may not have reproduced the crash but I have solved the problem of app not responding when a keyboard is attached to the mobile. And if given the log-cat I would try to solve the crash problem if it still persists, but I am positive that the problem is solved.

Paramjeet2810:
Actually I am not getting emulators of qwerty(with no touchscreen) devices. So as soon as I find one I will see to it that the app even works for those phones. But I am very positive that it will work as long as they follow the  universally accepted ASCII system.

Valodim:
Hi! Sorry for the delay, I'm a little between things lately :(

Can you explain what the problem was? From what I understand in the code, you handle the onKeyUp method manually now. do you know why this crashed before?

Paramjeet2810:
On key press event wasn't functioning so I had to manually add the code.
Try connecting a USB keyboard to your mobile and use the app. The app
doesn't work when the navigation drawer is open.

Paramjeet2810:
Sir, can we have a video chat at sometime possible by you. I guess it would save alot of your time as well as mine. I want to contribute more in open-keychain and this could motivate me to work more efficiently.
@nmikhailov I am really sorry for taking such a long time for solving a small issue.

dschuermann:
Sry, but we do not consider this an appropriate way to fix the problem.

PR #1940, opened 2016-10-25
Support of OpenPGP card v3

af-anssi:


dschuermann:
Some context: Related project is here: https://github.com/ANSSI-FR/SmartPGP

dschuermann:
@nmikhailov can you help reviewing?

nmikhailov:
@dschuermann I will look at the weekend.

af-anssi:
Could we have some more information on the review status ? Is it still ongoing ? Is there any issue I can help with ?

dschuermann:
@af-anssi Sry, I am extremely busy with my main job currently :disappointed:  I am still trying to free up time for OpenKeychain...

Hopefully @nmikhailov can help.

dschuermann:
@af-anssi @nmikhailov Just to give you an update: I freed up time and I am working on OpenKeychain again. After fixing https://github.com/open-keychain/open-keychain/issues/1960, this pull request is definitely the next thing I will be working on!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -319,6 +320,28 @@ public RSAPrivateCrtKey getCrtSecretKey() throws PgpGeneralException {
         return (RSAPrivateCrtKey)retVal;
     }
 
+    // For use only in card export; returns the secret key.
```
dschuermann: Same comment applies here as to the CanonicalizedPublicKey. Maybe move the smart card related methods to a new file containing the CanonicalizedSecretKey

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
```diff
@@ -1645,20 +1649,44 @@ private static boolean isDivertToCard(PGPSecretKey secretKey) {
...
+        final PGPPublicKey publicKey = key.getPublicKey();
+        ASN1ObjectIdentifier curve;
+
+        switch (publicKey.getAlgorithm()) {
+            case PublicKeyAlgorithmTags.RSA_ENCRYPT:
+            case PublicKeyAlgorithmTags.RSA_SIGN:
+            case PublicKeyAlgorithmTags.RSA_GENERAL:
+                // Key size must be at least 2048
+                if (publicKey.getBitStrength() < 2048) {
+                    log.add(LogType.MSG_MF_ERROR_BAD_SECURITY_TOKEN_SIZE, indent + 1);
```
dschuermann: This is no longer an appropriate log entry. The old if statement limited the allowed RSA keys to 2048 bit only because all available security tokens had this limit.

Can we check what RSA bit sizes and ECC curves are really supported by an attached token in ``SecurityTokenHelper.putKey``? Another possibility: Is an appropriate error returned by the security token which could be displayed to the user in this case?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
```diff
@@ -0,0 +1,714 @@
+package org.sufficientlysecure.keychain.securitytoken;
```
dschuermann: Please add GPL license header

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
```diff
@@ -0,0 +1,714 @@
...
+                                                 final byte[] data) throws IOException {
+        try {
+
+            if (certFactory == null) {
+                certFactory = CertificateFactory.getInstance(CERTIFICATE_FORMAT, PROVIDER);
+            }
+
+            final ECParameterSpec kfParams = getAlgorithmParameterSpec(kf);
+
+            final Certificate _cardCert = certFactory.generateCertificate(new ByteArrayInputStream(data));
```
dschuermann: no underscore in variable names please

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecureMessaging.java
```diff
@@ -0,0 +1,17 @@
+package org.sufficientlysecure.keychain.securitytoken;
```
dschuermann: add license header

af-anssi: done

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
af-anssi: done

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenUtilsTest.java.rej
```diff
@@ -0,0 +1,12 @@
+--- OpenKeychain/src/test/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenUtilsTest.java
```
af-anssi: you are right. i remove it.

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenUtilsTest.java.orig
```diff
@@ -0,0 +1,234 @@
+/*
```
af-anssi: you are right. i remove it.

dschuermann:
Great to see you working on this again! Please also rebase on current git master.

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperationTest.java
```diff
@@ -856,6 +856,7 @@ public void testKeyToSecurityToken() throws Exception {
 
         UncachedKeyRing modified;
 
+        /*
```
dschuermann: remove this test instead of commenting it out.

af-anssi: done

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
af-anssi: done

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -238,7 +326,47 @@ public void modifyPin(int pw, byte[] newPin) throws IOException {
...
+                return response.getData();
+
+            case ECKeyFormatType:
+                data = response.getData();
+
+                final byte[] keyEnc = new byte[encryptedSessionKey[pLen + 2]];
+
+                System.arraycopy(encryptedSessionKey, 2 + pLen + 1, keyEnc, 0, keyEnc.length);
+
+                try {
```
af-anssi: The [3.0 specification](http://www.g10code.com/docs/openpgp-card-3.0.pdf) specifies the PSO:DECIPHER operation (section 7.2.11) in case of ECDH. "In case of ECDH the card supports a partial decrypt only" and "With its own private key and the given public key the card calculates a shared secret in compliance with the Elliptic Curve Key Agreement Scheme from Diffie-Hellman. The shared secret is returned in the response, all other calculation for deciphering are done outside of the card." (page 53).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
af-anssi: This code is testing that the RSA key is 2048 bits at least, not at most. This test can be remove if support of 1024 bits keys is needed. For information, the support of RSA keys smaller than 2048 bits has been removed from version 3.x of OpenPGP card specification.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
af-anssi: I understand your point, and I implemented this method in this class because of an already existing analog method called getCrtSecretKey for RSA. Both methods (getCrtSecretKey and getECSecretKey) require to have access to the private field mPrivateKey of the instance. Implementing these methods out of the class CanonicalizedSecretKey would require to un-private the field mPrivateKey, or to create an accessor for this field. These methods are only used in security token related code, but if we create a "wrapping" class, it will require to alter a lot of code to propagate the use of this new class. Tell me what you think is best and I will implement it.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
dschuermann: Great, please change the comment above to:
``// Key size must be at least 2048 since OpenPGP card specification 3.x``
and change the string ``edit_key_error_bad_security_token_size`` to something similar.

Still: Can we check what RSA bit sizes and ECC curves are really supported by an attached token in ``SecurityTokenHelper.putKey``? Another possibility: Is an appropriate error returned by the security token which could be displayed to the user in this case?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
dschuermann: I took some time to get back into the code and you are right. For now keep the methods where they are but prefix them with ``securityToken``, such as ``getSecurityTokenECSecretKey``. Please also do this in ``CanonicalizedPublicKey``.

Also rename ``getCrtSecretKey`` to ``getSecurityTokenRSASecretKey``.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
dschuermann: OK, please add some comment in the code here for where to find this information.
Couldn't the session key be encrypted with a different algorithm instead of AES?

dschuermann:
@af-anssi I added new comments!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
af-anssi: Done.
For AES the answer comes from rfc6637#section-13 : "This document explicitly discourages the use of algorithms other than AES as a KEK algorithm."

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
af-anssi: done

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpKeyOperation.java
af-anssi: Done.

There is no command in the specification to retrieve such information from a token, and the overall specification (section 4.4.3.8, page 28) is quite unprecise on supported algorithms : support of RSA 2048 or higher is mandatory, EC{DS,H} is optional but if implemented at least one curve listed in the specification shall be supported.

dschuermann:
@af-anssi Great work and thanks for the contribution!

af-anssi:
Do not hesitate to assign bugs/issues related to my modifications !

Issue #1905, opened 2016-08-03
Use secure_delete for SQLite


dschuermann:
A deleted secret key is currently not overwritten by zeros, but this behavior could maybe reached by configuring sqlite correctly.

see recently merged patches in AOSP:
https://android-review.googlesource.com/#/c/210080/
https://android-review.googlesource.com/#/c/209123/

Looks like these options can also be set with a pragma statement (https://www.sqlite.org/pragma.html#pragma_secure_delete) so maybe we could adopt these.

harningt:
Since pragma's are conveniently not sanity-checked, if the user's Android version doesn't support secure_delete, it will just ignore it. :)

dschuermann:
This makes problems: https://github.com/open-keychain/open-keychain/pull/1947 and is not really worth the effort -> closing

PR #1947, opened 2016-11-10
KeychainDatabase: enable secure_delete explicitly to address GH #1905

harningt:
When the database is open, alongside enabling foreign key support, enable secure_delete so that platforms that support secure_delete will use it even if the default wasn't to enable it.


dschuermann:
Thanks for your contribution!

Looks like the instrumentation test fails on Travis with 
```
Starting 0 tests on test(AVD) - 5.0.2
Tests on test(AVD) - 5.0.2 failed: Instrumentation run failed due to 'android.database.sqlite.SQLiteException'
```

Can you locally try to run the android tests and look into it?

PR #1948, opened 2016-11-15
break down result handling in OpenPgpService into methods

Valodim:
Preparations for work on per-client tofu trust model. There *should* be no semantic changes in this PR.

I changed some api level constants into descriptive feature-based names, and broke down the way pgp results are parsed into result intent into nicely separated methods.


dschuermann:
@Valodim Looks good, please rebase.

PR #1950, opened 2016-11-17
OpenPgpSignatureResult v5

### LINKED ISSUE #1949: Date and time API
Hi,
is it possible to create an API that K-9 could retrieve the date and time for a verified signature and when decrypting an encrypted message?

EDIT: The corresponding issue in the K-9 repo is no. 1761.

Valodim:
This PR primarily deprecates v3 of OpenPgpSignatureResult, switching out the int constants of sender result info for an enum. Note that this breaks *forward* compatibility, i.e. if OpenKeychain cannot be newer than the client app it won't work. However, v3 (i.e. sender result) has so far only been used experimentally in the K-9 beta, so if we leave some time between rolling out a new version in K-9 and only then mergereleasing this, it should be fine.

While I was at it, I added a field for the signature timestamp in version 5. No concrete plans to use it yet, but the timestamp definitely belongs next to other signature info. (this closes #1949)


Valodim:
bad idea, new pr in a minute

PR #1973, opened 2016-12-31
fix loading of multiple keys simultaneously

Valodim:
the way `mCurrent` was used simply wasn't capable of handling multiple processes at once. proposed fix


dschuermann:
👍

PR #1974, opened 2017-01-01
fix typos in german translation

kerlerm:


kerlerm:
will use transifex to fix.

Valodim:
Thanks :+1:

PR #1975, opened 2017-01-02
fix wording: remembered -> cached

tessus:
While one can ask an app to remember credentials, the app itself does not remember them, but uses a cache to store them.

The correct term is `cached` instead of `remembered`.


Valodim:
See #1976

PR #1980, opened 2017-01-04
fix loop bug in key selection

Valodim:
I originally wanted to redo this entire activity, but that turned out to be more complex than I had thought. This fixes the current dialog instead - I hope :)

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
@@ -131,14 +131,12 @@ public class OpenPgpService extends Service {
     }
 
     private KeyIdResult returnKeyIdsFromEmails(Intent data, String[] encryptionUserIds, boolean isOpportunistic) {
-        boolean noUserIdsCheck = (encryptionUserIds == null || encryptionUserIds.length == 0);
-        boolean missingUserIdsCheck = false;
-        boolean duplicateUserIdsCheck = false;
+        boolean hasUserIds = (encryptionUserIds != null && encryptionUserIds.length > 0);
 
         HashSet<Long> keyIds = new HashSet<>();
         ArrayList<String> missingEmails = new ArrayList<>();
         ArrayList<String> duplicateEmails = new ArrayList<>();
-        if (!noUserIdsCheck) {
+        if (hasUserIds) {
             for (String rawUserId : encryptionUserIds) {
                 OpenPgpUtils.UserId userId = KeyRing.splitUserId(rawUserId);
                 String email = userId.email != null ? userId.email : rawUserId;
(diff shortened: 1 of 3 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
```diff
@@ -151,19 +149,17 @@ private KeyIdResult returnKeyIdsFromEmails(Intent data, String[] encryptionUserI
...
                     // another entry for this email -> two keys with the same email inside user id
                     if (cursor != null && cursor.moveToNext()) {
-                        duplicateUserIdsCheck = true;
                         duplicateEmails.add(email);
 
                         // also pre-select
-                        long id = cursor.getLong(cursor.getColumnIndex(KeyRings.MASTER_KEY_ID));
-                        keyIds.add(id);
-                        Log.d(Constants.TAG, "more than one user id with the same email");
+                         long id = cursor.getLong(cursor.getColumnIndex(KeyRings.MASTER_KEY_ID));
```
dschuermann: weird whitespaces

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpService.java
```diff
@@ -151,19 +149,17 @@ private KeyIdResult returnKeyIdsFromEmails(Intent data, String[] encryptionUserI
                         long id = cursor.getLong(cursor.getColumnIndex(KeyRings.MASTER_KEY_ID));
                         keyIds.add(id);
                     } else {
-                        missingUserIdsCheck = true;
                         missingEmails.add(email);
                         Log.d(Constants.TAG, "user id missing");
                     }
                     // another entry for this email -> two keys with the same email inside user id
                     if (cursor != null && cursor.moveToNext()) {
```
dschuermann: btw: shouldn't this be a loop? If we have 3 keys with the same email address, all 3 should be pre-selected.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ui/RemoteSelectPubKeyActivity.java
```diff
@@ -103,7 +103,7 @@ protected void onCreate(Bundle savedInstanceState) {
                     public void onClick(View v) {
                         // add key ids to params Bundle for new request
                         Intent resultData = extras.getParcelable(EXTRA_DATA);
-                        resultData.putExtra(OpenPgpApi.EXTRA_KEY_IDS,
+                        resultData.putExtra(OpenPgpApi.EXTRA_KEY_IDS_SELECTED,
```
dschuermann: why do we need an additional extra? What about merging the additional key ids with the existing given ``EXTRA_KEY_IDS`` before returning ``EXTRA_KEY_IDS``?

Valodim: This is the principal change here: By using a different extra, we know that the disambiguation happened (and what its result was), and can use its result instead of calling `getKeyIdsFromEmails`. An alternative would be to just return a boolean and adding the keys to `EXTRA_KEY_IDS`, but that just muddles input we received from the client app and our own state together.

dschuermann: Please correct me, but I thought that ``EXTRA_USER_IDS`` is removed for the second execution. Then only ``EXTRA_KEY_IDS`` is available and thus non-ambiguous.

PR #1993, opened 2017-01-11
Move key extraction logic into its own class

### LINKED ISSUE #1990: Can't select keys from open-keychain to use with Android Password Store
In Android Password Store I can select OpenKeychain as my OpenPGP provider to select which keys to use for encryption. Both my keys are listed but when I try to select receivers and save the screen is just reloaded showing the same keys with none selected.

screenshot_2017-01-11-12-50-11

Valodim:
This PR extracts the key logic ("which encryption keys to use, given a bunch of extras, possibly returning a pending intent for user interaction") from OpenPgpService into its own, testable class - and adds tests for it! Woop woop.

Also fixes #1990


dschuermann:
awesome!

Issue #1935, opened 2016-10-03
User ID matching for key selection is case sensitive


comzeradd:
I report this on [K-9](https://github.com/k9mail/k-9/issues/1645), but probably it's a Openkeychain issue.

I have a contact who, for some weird reason, has a couple of capital letters on his pgp email id. So when I pick him as recipient on K-9, OpenKeyChain fails to detect the correct key.

Valodim:
This is actually a K-9 bug :)

comzeradd:
So, just to be clean if K-9 is asking for an email with capitalized letters will OpenKeyChain match a key with a lowercase equivalent?

Valodim:
This is slightly more tricky than I had thought. So, what brought me to close this issue is that OpenKeychain already does case insensitive matching, and returns all matching keys and their status, together with the user id that matched the query string. What we do *not* return is the query string that matched the user id, which means K-9 would have to also case insensitively match the ids from the result set.

Doing that is not easy though, I'm not sure how I would do it in sql. I'd really like to avoid resorting to a matrix or merge cursor here :\

clawoflight:
> I'm not sure how I would do it in sql.

I vaguely recall that you use SQLite. From https://www.sqlite.org/lang_expr.html:

> The LIKE operator does a pattern matching comparison. The operand to the right of the LIKE operator contains the pattern and the left hand operand contains the string to match against the pattern. A percent symbol ("%") in the LIKE pattern matches any sequence of zero or more characters in the string. An underscore ("_") in the LIKE pattern matches any single character in the string. Any other character matches itself or its lower/upper case equivalent (i.e. **case-insensitive matching**). Important Note: **SQLite only understands upper/lower case for ASCII characters by default**. The LIKE operator is case sensitive by default for unicode characters that are beyond the ASCII range. For example, the expression 'a' LIKE 'A' is TRUE but 'æ' LIKE 'Æ' is FALSE. **The ICU extension to SQLite includes an enhanced version of the LIKE operator that does case folding across all unicode characters.**

Assuming most email addresses fall in the ASCII range, this should mostly work already?

dschuermann:
fixed in https://github.com/open-keychain/open-keychain/pull/2001

PR #2001, opened 2017-01-17
externalprovider: use temp table for query

Valodim:
This commit introduces a temporary table to KeychainExternal provider.

This is used to return a result set that contains the exact query as identifying column, rather than the matching user id. This is helpful because we match case insensitively internally, while the querying implementation might wish to map search results against verbatim search strings.

The available projection columns are now address, status, and user id. If the client queries only address and status, they will be grouped into the best available verification status. Otherwise, it will be one verification status per user id (which is what K-9 does at the moment).

Related to #1935


dschuermann:
LGTM

Issue #1957, opened 2016-11-30
Replace SuperSLiM


dschuermann:
Currently we use https://github.com/TonicArtos/SuperSLiM and https://github.com/FutureMind/recycler-fast-scroll

Looks like SuperSLiM is not maintained anymore. A good alternative good be: https://github.com/davideas/FlexibleAdapter

daquexian:
I could work on this :)

daquexian:
The newest version of FlexibleAdapter is v5.0.0-rc1. Could I use this version?

dschuermann:
yes

daquexian:
And FlexibleAdapter has built-in FastScroller, replace FutureMind/recycler-fast-scroll with it, too?

dschuermann:
yup

PR #2026, opened 2017-02-02
Replace SuperSLim and FastScroll with FlexibleAdapter in KeyListFragment

daquexian:
## Description

I replaced SuperSLim and FastScroll with FlexibleAdapter in KeyListFragment, and also adopt the built-in highlighter. Please tell me if there is something I can improve :)

I will replace these two libraries in other files next days.

## Motivation and Context

issue #1957 

## How Has This Been Tested?

Tested on my phone and ran unit test

## Types of changes


daquexian:
Travis-CI says the quantity of methods exceeds 65536.. I builded successfully when my developping (before I ran `git pull`). I'm sorry that I have no experience about 64K limit.

dschuermann:
Not really your fault. Will look into this later. I made some changes regarding this limit in master.

dschuermann:
@daquexian Please rebase on current master and do git submodule update. Does this help the 64K issue?

daquexian:
I tried it and there is still this issue

dschuermann:
The inclusion of the new library could be the cause that OpenKeychain goes over 64k again. Please also try "gradlew clean" after rebasing to remove old compile objects.

daquexian:
@dschuermann Oh I tried "gradlew clean" and it solve the issue, thanks

dschuermann:
In general: The code looks okay. After fixing the bugs, I will merge this when the dependency on SuperSlim is fully replaced and old code is removed. Currently, KeyItem looks like code that is also still there for the old SuperSlim implementation.
Just keep adding commits to this PR to fix bugs and replace other SuperSlim instances. No need to squash or open a new PR :)

Branch crashes for me with:
```
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  Process: org.sufficientlysecure.keychain.debug, PID: 21496
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.String.substring(int, int)' on a null object reference
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.util.recyclerview.item.KeyItem.getSection(KeyItem.java:121)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.util.recyclerview.item.KeyItem.setHeaders(KeyItem.java:113)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.KeyListFragment.onLoadFinished(KeyListFragment.java:353)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at org.sufficientlysecure.keychain.ui.KeyListFragment.onLoadFinished(KeyListFragment.java:90)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.app.LoaderManagerImpl$LoaderInfo.callOnLoadFinished(LoaderManager.java:476)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.app.LoaderManagerImpl$LoaderInfo.onLoadComplete(LoaderManager.java:444)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.Loader.deliverResult(Loader.java:126)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.CursorLoader.deliverResult(CursorLoader.java:105)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.CursorLoader.deliverResult(CursorLoader.java:37)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.AsyncTaskLoader.dispatchOnLoadComplete(AsyncTaskLoader.java:252)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.AsyncTaskLoader$LoadTask.onPostExecute(AsyncTaskLoader.java:80)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.ModernAsyncTask.finish(ModernAsyncTask.java:485)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.support.v4.content.ModernAsyncTask$InternalHandler.handleMessage(ModernAsyncTask.java:502)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.os.Handler.dispatchMessage(Handler.java:102)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.os.Looper.loop(Looper.java:154)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at android.app.ActivityThread.main(ActivityThread.java:6119)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at java.lang.reflect.Method.invoke(Native Method)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
02-07 20:57:46.891  21496    21496         AndroidRuntime  E  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

dschuermann:
@daquexian BTW: You already did a lot of work! If you like to receive a Fidesmo card as a gift from us (see https://www.openkeychain.org/pr-incentive) email me your address (my email is on my GitHub profile). Also include your Jabber ID if you have one, so we can have a chat :)

dschuermann:
@daquexian Great progress! Should I test again?

daquexian:
@dschuermann Sure :) Please tell me if there is still any bug

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/KeyListFragment.java
```diff
@@ -292,25 +318,73 @@ private void startSearchForQuery() {
...
-        getAdapter().swapCursor(KeySectionedListAdapter.KeyListCursor.wrap(data));
+
+        mKeyItems.clear();
+
+        if (data.moveToFirst()) {
+            while (!data.isAfterLast()) {
+                mKeyItems.add(new KeyItem(null, KeyListCursor.wrap(data), mKeyListener));
+                data.moveToNext();
+            }
+        }
```
dschuermann: This is not the way we want to handle this. This code loads all items into an ArrayList before using the ArrayList from memory to display the items. For a long list of keys this requires that all keys are loaded into the memory, which is not good. Instead, the adapter should be able to work on a Cursor directly (as done before).

Unfortunately, this is a major blocker which needs to be resolved first.

dschuermann: The same is true for the other implementations in this pull request.

daquexian: I noticed that when coding. The reason why I didn't use the original way is that FlexibleAdapter does not provide `onCreateViewHolder` and `onBindViewHolder` in adapter, only provides `bindViewHolder` and `createViewHolder` in items. I will try to find a way to solve this problem.

dschuermann: Arg, I just found https://github.com/davideas/FlexibleAdapter/issues/208

@Valodim What do you think? Do we really need to operate on the Cursor for our lists or is it okay to load the items into memory?

daquexian: @dschuermann @Valodim In my opinion it doesn't matter to fetch all keys into memory, for they won't take up much memory. A compromising way is fetching items only when they are about to showing up, by using [Endless Scroll](https://github.com/davideas/FlexibleAdapter/wiki/5.x-%7C-On-Load-More) feature of FlexibleAdapter.
On the other hand, I was told that the author of superslim has come back on development of superslim. See [this](https://github.com/TonicArtos/SuperSLiM/issues/188) for detail. ( I'm a sensitive boy :) ) So wait for the version5 of superslim is also a choice. If so, please assign the task that migrate to version5 to me :)

davideas:
Hi, I'm the author of FlexibleAdapter, and checking if I can add the support of the Cursor. But...

The `IFlexible` item is a presentation item! For instance, items of a section must have the same header linked and for expandable items a subList assigned to each of them. If they need to be prepared for presentation, how the cursor should return those items already?
By the way, how many items are you talking about? And why they can't be loaded asynchronously with loading more feature by using a higher threshold: more items will be loaded transparently to the user.

dschuermann:
@davideas Thanks for your explanations. We are talking about ~100 entries. I am a power user and have ~400 public keys in my desktop key manager. At what number of items would you propose loading asynchronously?

davideas:
@dschuermann, I would adapt the pageSize by an acceptable loading time using Cursor and I would identify a threshold in a range of 10-30% of the pageSize in order to support fast scrolling: Supposing that loading 100 items with the Cursor and creating the items for presentation it takes a total of ~300ms(?), you can keep a threshold of 20 items before loading more items.

A short briefing on EndlessScroll feature in my library:
It relies on the new bound item. It calculates the difference to reach the end of the list based on the bound position. This solution is lighter than listening from the ScrollListener each scrolled pixel.
In our example with 20 items as threshold, having 100 items loaded and pageSize of 100, in the list, the first loading more is triggered at position 80 when the corresponding item is bound.

For me, it isn't a problem to keep _references_ of 10.000 of items or more in memory if it is requested by the user. I understood that the problem comes when loading them all. In my opinion if they are constantly loaded probably a new mechanism should be implemented, because user won't go through them all anyway in a small screen.

Another solution would be to load those items in delayed series without waiting the user scrolls.

I hope I gave you cleared points to think about.

Valodim:
superseded by #2351

Issue #2022, opened 2017-01-30
Display insecure key warnings


dschuermann:
On import:
* If a subkey is insecure (as in ``PgpSecurityConstants.isSecureKey()``) it should be marked as insecure in our database, similar to revoked. 

In OpenKeychain's key lists/views:
* If all subkeys and the master key in an OpenPGP key are insecure, display them exactly like revoked keys are displayed. Only with a different message such as "This key is no longer secure and thus must not be used anymore."

Selection for encryption:
* Do not allow selection of these keys for encryption, similar to revoked keys handling

rhari991:
I'd like to work on this.

dschuermann:
sure go ahead

rhari991:
@dschuermann When should the key be marked as secure/insecure ? When it is created ?

dschuermann:
@rhari991 When a key is saved into the database. I think an approriate place is in ProviderHelper:423.

PR #2028, opened 2017-02-03
Display insecure key warnings

rhari991:
Added changes related to insecure keys, as described in #2022 

## How Has This Been Tested?
Import insecure keys manually.

## Types of changes


dschuermann:
Great work, I have no required changes to your code.
If you have more time, you could add a test in ProviderHelperSaveTest. For now it is enough to test the CanonicalizedPublicKey.isSecure() method and not the UI code.

dschuermann:
@rhari991 I forgot one thing: That a key is insecure should also be shown when expanding a card on import similar to revocation and expiry.

dschuermann:
https://github.com/open-keychain/open-keychain/issues/2030

PR #2034, opened 2017-02-03
redesign access and allowed secret key api dialogs

Valodim:
This PR redesigns the two api dialogs, from fullscreen into regular material yes/no style dialogs.


Valodim:
screenshot_20170203-181510
screenshot_20170203-182434
screenshot_20170203-193219

scweenshots

dschuermann:
Looks awesome!

I encountered one corner-case bug:

I encrypt a message to two public keys with the same email address "alice@example.com". "Alice" is a secret key, "Alice 2" is only a public key:
device-2017-02-07-204823

I asks for access to "Alice 2" but not for access to "Alice" (where we have the secret part):
device-2017-02-07-204850

Valodim:
nice catch. should be fixed with the last commit

dschuermann:
👍

PR #2041, opened 2017-02-07
fix transparent header

daquexian:
Fix transparent header in ViewKeyAdvCertsFragment
## Description

When working on #1957 , I found that the header in ViewKeyAdvCertsFragment is transparent

Before modification:
photo6111930782709819305

After modification:
photo6111930782709819306

## Types of changes


dschuermann:
thanks!

Issue #1964, opened 2016-12-08
Refactor OpenKeychain into a OpenPGP library for Java and Android, and an Android app


Flowdalic:
As discussed before, it would be great to have a consumable OpenPGP library which runs on Java SE and on Android. This would allow me for example to implement XEP-373 / -374 in Smack.

The work you guys did for OpenPGP is just awesome, I'm sure other projects would love an usable open-source OpenPGP Java library. So please consider splitting the OpenKeychain (OK) codebase into a functional OpenPGP library part and OK.

PR #2052, opened 2017-02-10
Initialize libkeychain

dschuermann:
Create "libkeychain" library folder and move some utils in it.

At one point this will be https://github.com/open-keychain/open-keychain/issues/1964


Valodim:
Hm. To be honest I would have waited with merging this until we had some more changes.

The really difficult part is going to be getting rid of android dependencies in libkeychain, until we have at least a plan for that moving utility classes over is just a half measure.

dschuermann:
I don't think we can get to a library at once. This is a longer process. Maintaining a branch will not work for this.

dschuermann:
Maybe a bit more explanation: As a first step, I don't plan to get rid of Android dependencies. I will simply try to separate backend code into a different folder.

PR #2053, opened 2017-02-11
add EdDSA support

Valodim:
## Description
Support for eddsa \o/ based on [werner's draft](https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-04), adding support for ed25519 to bouncycastle based on [ed25519-java](https://github.com/str4d/ed25519-java) by @str4d (huge thanks!).

The bulk of changes of course happened in the bouncycastle repository:
https://github.com/open-keychain/bouncycastle/compare/openkeychain-master-old...eddsa
I suggest reading those commit by commit, where the first one is purely an import of the ed25519-java classes into the bouncycastle namespace, the second adds them to the BouncyCastleProvider, and the third adds openpgp support on top of that.

## How Has This Been Tested?
I tested this manually with an imported key from gpg, and have some hacky unit tests that use the sample keys from the eddsa draft. I created a key, signed some text with it and verified it with gpg, too. Switching generated keys in unit tests out for eddsa ones also works. This still needs actual unit tests though!

## Types of changes


Valodim:
added a bunch of tests. one checks verification of a gnupg signed message using a gnupg created key. the other generates a key and canonicalizes it, checking key creation, signature creation and signature verification in the process.

str4d:
Is there a reason why you are merging my code into your fork of BouncyCastle instead of using my published library?

`compile 'net.i2p.crypto:eddsa:0.1.0'`

I ask because if there's something the library doesn't provide, I'm happy to add it :smiley: If it's just because you only want to depend on BC then I don't have a personal preference, but if you're going to the effort of merging them, it would make more sense IMHO to get it into upstream BouncyCastle instead of having to maintain a persistent fork.

Also, you may want to wait until https://github.com/str4d/ed25519-java/pull/22 is merged, because my current master uses an older draft encoding that doesn't match the current draft (in WG last call, so nearly-final): https://datatracker.ietf.org/doc/draft-ietf-curdle-pkix/

str4d:
Scanning the linked commits, I'm guessing that you're merging in my code because you need it to be able to implement the OpenPGP support? That should also be doable by adding a compile dependency to `prov/build.gradle`, although since you're already maintaining a persistent fork, it doesn't really matter either way! My main concern is around limiting future divergence or staleness, but hopefully it's easy enough for you to keep pulling in code from upstream.

Valodim:
Hi, thanks for your feedback! We definitely planned to get the changes upstream into BouncyCastle, and our plan for that would be to bring it up on their mailing list to see what they think would be a suitable way to do so. Mostly I wanted to get this working as step one, then see how to do it properly in terms of code management later. :)

For the include vs dependency, the dep would have to be in BouncyCastle, not OpenKeychain, since the classes are used there (e.g. JcaPGPKeyConverter), so that would still result in a "import this here code" commit in our BC patch stack. The actual reason I did it this way is that when I started I didn't expect to be able to get away with such loose coupling :)

I have a question about the implementation: You're currently providing only SHA512withEd25519 as an algorithm, which directly uses the material to hash as an input, rather than use the output of a hashing function as input to the Eddsa function itself (which is itself parametrized with sha512).

It's a particularity of Ed25519 that the algorithm takes arbitrarily sized input, but at least in OpenPGP (and this makes senese for consistency), the input to the eddsa algorithm is still the output of an independent hash function, and doesn't care for that implementation detail. From my understanding, shouldn't the thing you offer actually be NONEwithEd25519? This is also reflected in the rather awkward ByteArrayOutputStream implementation you use to buffer all input.

dschuermann:
@str4d Werner's draft for OpenPGP (https://tools.ietf.org/html/draft-koch-eddsa-for-openpgp-04) uses Josefsson encoding (https://datatracker.ietf.org/doc/rfc8032/). Does this conflict with https://datatracker.ietf.org/doc/draft-ietf-curdle-pkix/ ?

Valodim:
both of those documents reference [irtf-cfrg-eddsa](https://tools.ietf.org/html/draft-irtf-cfrg-eddsa-08) for their encoding, so they should be the same?

Valodim:
@str4d ping? :)

str4d:
> For the include vs dependency, the dep would have to be in BouncyCastle, not OpenKeychain, since the classes are used there (e.g. JcaPGPKeyConverter), so that would still result in a "import this here code" commit in our BC patch stack. Note that bouncycastle has no dependencies otherwise, so this wouldn't really be viable. The actual reason I did it this way however, is that when I started I didn't expect to be able to get away with such loose coupling :)

Well, I do try to make my libraries usable ;P But yes, I agree that if this ended up being upstreamed to BouncyCastle, the code would most likely end up directly in their codebase rather than as a dependency.

> I have a question about the implementation: You're currently providing only SHA512withEd25519 as an algorithm, which directly uses the material to hash as an input, rather than use the output of a hashing function as input to the Eddsa function itself (which is itself parametrized with sha512).
>
> It's a particularity of Ed25519 that the algorithm takes arbitrarily sized input, but at least in OpenPGP (and this makes senese for consistency), the input to the eddsa algorithm is still the output of an independent hash function, and doesn't care for that implementation detail. This is what I did in SignatureDigestSpi. From my understanding, shouldn't the provided algorithm actually be called NONEwithEd25519? This is also reflected in the ByteArrayOutputStream implementation used to buffer all input.

I recall being very confused at the time by the JCA specifications on how to name these, and how the Ed25519 spec interacted with it. You might notice the presence of a JCA-incompatible "OneShotMode" that we added for efficiency, because we can't just push the data into the message digest over repeated updates. tl;dr I agree with you that Ed25519 should be NONEwith* and Ed25519ph should be SHA512with*, and I plan to fix this in my library (as well as implementing Ed25519ph).

However, I'm still not certain from the JCA spec whether they should be `NONEwithEd25519` or `NONEwithEdDSA` using `keysize = 256` - existing examples are not particularly clear. As it happens, one of the items on my todo list from a few weeks ago was to contact the BC people to get some more insight into this, specifically regarding their use in the OID mappings for the Provider.

Re: encoding, @dschuermann @Vadiml1024 I see your confusion. The encoding of the *points* are the same, using the Josefsson encoding you reference (which is now RFC 8032). The encoding of the *keys* (`net.i2p.crypto.eddsa.EdDSAPublicKey` and `net.i2p.crypto.eddsa.EdDSAPrivateKey`), however, are not:

- The [draft Josefsson encoding](https://tools.ietf.org/html/draft-josefsson-pkix-eddsa-04) (which is was all that existed when I first wrote my library) specifies a single OID for all EdDSA types, and uses a parameter to specify which variant is being used. It also only specified the public key format; we extrapolated a private key format from that and the PKCS#8 spec.
- The [Curdle WG encoding](https://tools.ietf.org/html/draft-ietf-curdle-pkix-03) (which is what will become an RFC) uses a separate OID for each EdDSA type, and does *NOT* allow any parameters at all, not even a null: `For all of the OIDs, the parameters MUST be absent. Regardless of the defect in the original 1997 syntax, implementations MUST NOT accept a parameters value of NULL.`
  - This is itself a problem, because Java's default keystore always puts a NULL in there. I have just sent an email to the Curdle WG asking about this...

dschuermann:
Just a small comment without diving further into it: The josefsson key encoding is also an RFC now. We definitely need to use the format that Werner proposes.

Valodim:
I have a unit test that imports a pulic key and verifies a signed message with it, both straight from gnupg's own test suite. So I'm not worried that we use the right point format or break that in the future.

dschuermann:
I am also not worried that we currently use the wrong key format. I am worried that @str4d s changes using the curdle RFC introduces the wrong key format.

dschuermann:
I would propose that we first do a PR at bouncy castle to get more reviews on this security critical code before merging this into OpenKeychain.

str4d:
@dschuermann Unfortunately, the word "key" is significantly overloaded in this context; I apologise for not being clearer. There are several distinct encodings:

- RFC 8032 specifies the encoding of the actual EdDSA private key material and public key curve points.
- The two drafts I linked in https://github.com/open-keychain/open-keychain/pull/2053#issuecomment-280846484 specify the encoding of RFC 8032 keys in ASN.1 PKI key certificates.
- Werner's draft specifies the encoding of RFC 8032 in OpenPGP packets.

So, for the purposes of this PR, you could completely ignore my comments about encoding, *assuming that open-keychain never stores keys in ASN.1 form* (which I do not know). However, the PR is not appropriate for merging my code upstream into BouncyCastle, because that will inevitably be used beyond OpenPGP and so the PKI encoding is very relevant. Hopefully the issues PKI side will be resolved soon, however!

Re: the OIDs, I am completely unfamiliar with the OID allocation process, but my guess is that the OIDs for OpenPGP are intentionally distinct from the OIDs for PKI, so Werner's draft will be correct there.

str4d:
FYI I've released v0.2.0 of my library, in which the PKI encoding has been specified.

Valodim:
rebased on current master and updated to v0.2.0

keks:
What's the status on this? Would love to see this merged :)

Valodim:
We decided to merge this. Sadly, ed25519 support isn't worth much without cv25519, and I was unable to implement that from the existing specs :(

keks:
@Valodim Oh yes, been there when I failed implementing it for the Go OpenPGP library.

There is some unmerged code from someone else, maybe you will find some insight in there. I don't know whether the serialization code is correct though, as the whole thing doesn't build. Anyway, maybe it helps. [Here](https://go.googlesource.com/crypto/+/refs/changes/95/22095/6/openpgp/packet/public_key.go#350) is the serialization code and [here](https://go-review.googlesource.com/c/22095/) is the entire changeset.

keks:
Maybe taking a look [here] is also worthwhile

[here]: https://github.com/benburkert/crypto/commits/openpgp-ecdh

Valodim:
the serialization is the same as for ECDH, that part already works. the issues start with actually parsing the MPI into an actual ECPoint

keks:
Now I remember. This is exactly the point where the Go implementation stalled as well. I sure hope Werner and the other finally manage to properly document (and maybe even specify) what they are doing in GnuPG.

I tried reading the ECDH code in libgcrypt (or how their crypto library is called) but the fact that write C like others write Lisp (I'm referring to S-Expressions) makes the whole thing unreadable to outsiders.

Valodim:
Haha, so did I. Nice to see somebody else has been at this exact same point :)

PR #2066, opened 2017-02-20
Refactor database access

Valodim:
I started refactoring our database access methods. The name "ProviderHelper" has been a thorn in my side for a while, but we also get two practical benefits:

- I separated write access from read access, so classes that don't write to the db can work with a superclass that doesn't implement methods that write to the db. As a very direct benefit, since both OpenPgpService and the `Pgp*` operations don't require write access, the class with write access can be package local.

- My plan is to centralize all access to the actual key blobs into as little code as possible, ideally only a single method per public and secret key reading and writing. This isn't finished, but I've made progress, most notably getting rid of the join for the blobs in the unified key query, which should reduce the attack surface to get to those blobs significantly.

The latter aspect is going to be useful down the line in two (and a half) ways:

- We can apply transformations (like encryption) more easily

- Since access happens in fewer places, it becomes feasible to store the blobs externally, sidestepping sqlite filesize restrictions

- We could choose to store *all* keys as blobs in internal storage, keeping none or only a minified version in the database. That would allow restoring the database from those files more easily, but also comes at a higher risk of consistency problems.

So far, this PR *should* contain no actual semantic changes, but merely changes the way things are loaded in some places.

I actually noticed after I did the renaming that Interactor is not a good term for what the classes do, I believe the correct term is "Repository", although "Data Access Object" would also be fine. However rebasing that rename is difficult so I'm going to do that as a last step when everything else is done here.


Valodim:
Updated PR.

Public keys are now always written into `files/public_keys/`, and only keys with filesize < 50kb are cached in the database for faster access. This fixes problems with large keys, tested with the [Edward the GnuPG bot](http://pool.sks-keyservers.net/pks/lookup?search=0x9ff2194cc09a61e8&fingerprint=on&op=vindex) key.

I also renamed the class to KeyRepository :)

dschuermann:
Read the code, LGTM! I still need to test.

Valodim:
This doesn't yet have a migration. There is no problem with that, having keys not on disk which would be written to disk now has no consequences. We just have to be aware of that if we ever do a migration or change code that does rely on large keys being available on disk (doesn't seem very likely).

This is no longer a WIP. Definitely warrants a beta release though.

Valodim:
I did some more testing what happens when the key on disk is gone. It's handled gracefully in all places I tested, most importantly updating the key still works (fixing the key in the process) and deletion as well.

PR #2072, opened 2017-03-09
handle unknown key algorithms using opaque keys

Valodim:
This PR adds handling of keys with unknown algorithms as opaque keys. Those will be removed during canonicalization so they never end up in the database, but at least those keys don't break our import completely anymore. Keys with eddsa primary key will simply fail to import, subkeys will be stripped so they won't be available.

This can lead to weird scenarios where signatures can't be verified because the specific subkeys aren't available, but it's a lot better than what we have now.

Relevant changes mostly in bouncycastle. I actually had handling for this in canonicalize all along, it just never got to that because bouncycastle would throw an IOEXception before we ever got there :)

(@dkg made me do it)


dschuermann:
👍

dschuermann:
@Valodim can you put the two commits at https://github.com/open-keychain/bouncycastle/tree/opaque-keys also onto https://github.com/open-keychain/bouncycastle/tree/openkeychain-master-old ? and remove the two branches for newline parsing and opaque keys.

Valodim:
why "master-old", and not "master"?

dschuermann:
Yeah, sry. master is my half-baked try for the newest bouncycastle. But currently we still use master-old.

Valodim:
done

PR #2083, opened 2017-03-22
Secure private key transfer

tobiasschuelke:
## Description

Private keys can be imported via a new actionbar item in the import keys fragment and exported via a new button in the view key fragment. The key is transferred over a wlan network. The connection between both devices can be established by scanning a QR Code or by entering the ip address and port number of the second device manually. Using the SecureDataSocket library (https://github.com/MrJabo/SecureDataSocket) guarantees a secure transmission of the private key. The private key can also be exchanged with a computer that has installed Kelivery (https://github.com/svenklaus/keylivery). This feature is available for devices with Android version Kitkat or higher. Older versions cannot be used with the SecureDataSocket library.

SecureDataSocket uses a commitment scheme based on hash commitments (http://crypto.cs.mcgill.ca/~crepeau/COMP647/2007/TOPIC07/Halevi-Micali.pdf) and SAS (https://www.iacr.org/archive/crypto2005/36210303/36210303.pdf). After the key exchange succeeded the private key will be send encrypted with AES-GCM 128 bit (based on diffie hellman) to the other device.

## Motivation and Context
 
 Many users need to transfer their private key from their computer to open-keychain. This feature simplifies this process. Moreover, the keys can also be transferred to other android devices such as a tablet.

## Screenshots

Import a key via QR Code:

qr

Export a key (3 screenshots on left side) and import on another device (3 screenshots on the right side) without QR Code:

manual

## Types of changes
✅ New feature (non-breaking change which adds functionality)


dschuermann:
Hey, thanks for the big pull request.
Will need to free up some time to review it.

The first things I notice:
* Don't change the submodules
* Travis fails due to dependencies. Try changing your build files or the travis file accordingly.

Valodim:
So first of all, I love the general idea of this, especially if it is able to interact with a desktop client! Thanks for your work!

However I have a couple of questions about the concrete implementation:

- What is the advantage of the hash commitment scheme over a simple authenticated diffie hellman?

- What are the circumstances where the user is asked to type the ip and port manually? Only if there is no camera? I tend to think we'd rather fail than ask the user to manually type this kind of information.

I'm also unsure about including the SecureDataSocket library from a maintenance perspective. From the looks of it, it's an academic implementation, started last month, that hasn't gone through security review. It's also hard to tell whether it will be maintained a year from now.

0xmatthias:
Thanks for your comment

First of all the hash commitment is only used, if there is no camera (then the user has to type port and ip manually). We chose a hash commitment to simplify the key exchange. An authenticated diffie hellman is based on DSA which requires to verify the DSA public keys which then authenticates the diffie hellman. I would like to point out, that in this case we do not have a trusted third party or a secure channel (http://etutorials.org/Programming/secure+programming/Chapter+8.+Authentication+and+Key+Exchange/8.18+Using+Diffie-Hellman+and+DSA+Together/). You can take a look at https://github.com/MrJabo/SecureDataSocket/blob/master/protocol_high_level_overview for more information.

dschuermann:
So, we are not sure if the hash commitment based fallback with IP+port is really useful. It serves only a very small user base that have no camera.

At this point in development, we need to make sure that OpenKeychain does not grow in its code size too much. Also, code that sits in a different git repository is often overlooked. Having code in OpenKeychain's main repo makes code much more visible to the developers. We learned this the hard way in the past and are trying to improve here.

What I would **definitely** accept as a contribution:
* remove dependency on the lib
* remove hash commitment fallback with IP+port
* Include everything else into the main repo. Maybe in a subpackage in OpenKeychain
* The server side app should be available and downloadable on openkeychain.org. This requires a PR to https://github.com/open-keychain/open-keychain.github.io

I really like this feature, we can also improve the UX after having the basic structure as described above. Hopefully, we find a good middle way that also satisfies you guys.

tobiasschuelke:
We understand your issues with code size and external repositories. That´s why we changed our code according to your points. Moreover, the builds of Travis now pass. A PR of the PC program to the website of openkeychain will be following soon.

dschuermann:
Nice work. Two nitpicks before we proceed:
* Please format the added files according to our code style (at least adhere to the whitespaces instead of tabs)
* Add license headers to the files you added

tobiasschuelke:
The files are now formatted and they have license headers.

Valodim:
Tested this! :)

I tried to transfer a key. Selected, scanned QR code, OpenKeychain said "Bad Data" and I got this stacktrace from the keylivery application, no feedback in the UI though:

```
Prepare Server at: 192.168.179.59:4711
com.cryptolib.SecureDataSocketException: com.cryptolib.CryptoSocketException: not the same sharedSecret is used. maybe the wrong device connected to you
com.cryptolib.CryptoSocketException: not the same sharedSecret is used. maybe the wrong device connected to you
	at com.cryptolib.CryptoSocket.listen(CryptoSocket.java:159)
	at com.cryptolib.FDESocket.listen(FDESocket.java:69)
	at com.cryptolib.SecureDataSocket.setupServerWithClientCamera(SecureDataSocket.java:130)
	at keylivery.server.ServerService$1.call(ServerService.java:37)
	at keylivery.server.ServerService$1.call(ServerService.java:29)
	at javafx.concurrent.Task$TaskCallable.call(Task.java:1423)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at javafx.concurrent.Service.lambda$null$6(Service.java:725)
	at java.security.AccessController.doPrivileged(Native Method)
	at javafx.concurrent.Service.lambda$executeTask$7(Service.java:724)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

	at com.cryptolib.SecureDataSocket.setupServerWithClientCamera(SecureDataSocket.java:132)
	at keylivery.server.ServerService$1.call(ServerService.java:37)
	at keylivery.server.ServerService$1.call(ServerService.java:29)
	at javafx.concurrent.Task$TaskCallable.call(Task.java:1423)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at javafx.concurrent.Service.lambda$null$6(Service.java:725)
	at java.security.AccessController.doPrivileged(Native Method)
	at javafx.concurrent.Service.lambda$executeTask$7(Service.java:724)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
SERVER: DONE!
Prepare Server at: 192.168.179.59:4711
com.cryptolib.SecureDataSocketException: java.net.BindException: Address already in use (Bind failed)
java.net.BindException: Address already in use (Bind failed)
	at java.net.PlainSocketImpl.socketBind(Native Method)
	at java.net.AbstractPlainSocketImpl.bind(AbstractPlainSocketImpl.java:387)
	at java.net.ServerSocket.bind(ServerSocket.java:375)
	at java.net.ServerSocket.<init>(ServerSocket.java:237)
	at java.net.ServerSocket.<init>(ServerSocket.java:128)
	at com.cryptolib.CryptoSocket.listen(CryptoSocket.java:117)
	at com.cryptolib.FDESocket.listen(FDESocket.java:69)
	at com.cryptolib.SecureDataSocket.setupServerWithClientCamera(SecureDataSocket.java:130)
	at keylivery.server.ServerService$1.call(ServerService.java:37)
	at keylivery.server.ServerService$1.call(ServerService.java:29)
	at javafx.concurrent.Task$TaskCallable.call(Task.java:1423)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at javafx.concurrent.Service.lambda$null$6(Service.java:725)
	at java.security.AccessController.doPrivileged(Native Method)
	at javafx.concurrent.Service.lambda$executeTask$7(Service.java:724)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

	at com.cryptolib.SecureDataSocket.setupServerWithClientCamera(SecureDataSocket.java:132)
	at keylivery.server.ServerService$1.call(ServerService.java:37)
	at keylivery.server.ServerService$1.call(ServerService.java:29)
	at javafx.concurrent.Task$TaskCallable.call(Task.java:1423)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at javafx.concurrent.Service.lambda$null$6(Service.java:725)
	at java.security.AccessController.doPrivileged(Native Method)
	at javafx.concurrent.Service.lambda$executeTask$7(Service.java:724)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

Valodim:
We talked this through some more, also getting feedback from Oliver earlier today. Summary of what we're thinking right now:

### key exchange protocol

We came to the conclusion that a custom protocol for secure data transfer just isn't going to work. Even for a simple authenticated diffie-hellman, we would have a hard time justifying why we decided to include an implementation that is used nowhere else, and that we hardly reviewed ourselves. Apart from that, availability as a java library is not optimal for adoption by other tools.

The solution we came up with is [TLS-PSK](https://en.wikipedia.org/wiki/TLS-PSK). This is basically a type of TLS ciphersuite that uses a symmetric PSK instead of asymmetric certificates, which in its EDH variants offers mutual authentication and forward secrecy. It comes [natively with android](https://developer.android.com/reference/android/net/PskKeyManager.html), and is also available in [gnutls](https://github.com/defuse/gnutls-psk) - which is an indirect dependency of gpg already. I was able to hammer out a proof of concept on top of this PR using tls-psk in a couple of hours, that worked between OpenKeychains and a simple invocation of gnutls-cli.

### user workflow

We'd like to have a user workflow that allows transfer of keys in either direction, while the QR code can be scanned from either side. Rough draft, this is what we imagine:



This is a two-step workflow rather than single-step, an idea which is driven by two thoughts:
- Either device can scan the QR code. It's no issue technically to send a key in either direction, but choosing which device should send their key a priori is not easy to combine with a bidirectional exchange.
- The additional dialog before sending after a connection is established makes it very explicit and clear when the sending happens and who is doing it, reducing the chance of giving up one's secret key by accident.

Valodim:
I implemented what we had in mind in #2117, which supersedes this PR. Thanks a lot for your work on this! :+1:

PR #2088, opened 2017-04-04
Use KeychainExternalProvider for resolving encryption keys in api

Valodim:
This PR unifies the way key lookup (user id -> key id) is done. This affects primarily OpenPgpServiceKeyIdExtractor, which previously had its own logic, but now queries the external provider for key ids instead.

This PR *also* introduces EXTRA_DRY_RUN, which causes a SIGN_ENCRYPT operation to abort once all input parameters are determined, and returns a status of keys that would be used in the process. This change is quite simple and isolated in a commit of its own.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/KeychainExternalProvider.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/KeychainExternalProvider.java
@@ -34,6 +34,7 @@ import android.os.Binder;
 import android.support.annotation.NonNull;
 import android.text.TextUtils;
 
+import org.sufficientlysecure.keychain.BuildConfig;
 import org.sufficientlysecure.keychain.Constants;
 import org.sufficientlysecure.keychain.provider.ApiDataAccessObject;
 import org.sufficientlysecure.keychain.provider.KeychainContract;
@@ -50,6 +51,7 @@ import org.sufficientlysecure.keychain.util.Log;
 
 public class KeychainExternalProvider extends ContentProvider implements SimpleContentResolverInterface {
     private static final int EMAIL_STATUS = 101;
+    private static final int EMAIL_STATUS_INTERNAL = 102;
     private static final int API_APPS = 301;
     private static final int API_APPS_BY_PACKAGE_NAME = 302;
 
@@ -78,8 +80,9 @@ public class KeychainExternalProvider extends ContentProvider implements SimpleC
          * </pre>
          */
         matcher.addURI(authority, KeychainExternalContract.BASE_EMAIL_STATUS, EMAIL_STATUS);
+        matcher.addURI(authority, KeychainExternalContract.BASE_EMAIL_STATUS + "/*", EMAIL_STATUS_INTERNAL);
 
-        matcher.addURI(KeychainContract.CONTENT_AUTHORITY, KeychainContract.BASE_API_APPS, API_APPS);
+        // can only query status of calling app - for internal use only!
         matcher.addURI(KeychainContract.CONTENT_AUTHORITY, KeychainContract.BASE_API_APPS + "/*", API_APPS_BY_PACKAGE_NAME);
 
         return matcher;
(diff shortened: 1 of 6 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceKeyIdExtractor.java
```diff
@@ -53,150 +47,223 @@ private OpenPgpServiceKeyIdExtractor(ContentResolver contentResolver,
...
-            result.putExtra(OpenPgpApi.RESULT_CODE, OpenPgpApi.RESULT_CODE_ERROR);
-            return new KeyIdResult(result);
-        }
-
-        return new KeyIdResult(encryptKeyIds);
+        return result;
     }
 
-    private KeyIdResult returnKeyIdsFromEmails(Intent data, String[] encryptionUserIds, boolean isOpportunistic) {
+    private KeyIdResult returnKeyIdsFromEmails(Intent data, String[] encryptionUserIds, String callingPackageName) {
```
Valodim: This method no longer determines the recipient status itself, but instead calls out to KeychainExternalProvider. The "callingPackageName" parameter allows making these queries calling package sensitive, which is not used right now but can be for application-specific key/identity bindings at a later point.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/KeychainExternalProvider.java
```diff
@@ -134,9 +137,19 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
...
+                if (!BuildConfig.APPLICATION_ID.equals(callingPackageName)) {
+                    throw new AccessControlException("This URI can only be called internally!");
+                }
+
+                // override package name to use any external
+                // callingPackageName = uri.getLastPathSegment();
+
             case EMAIL_STATUS: {
-                boolean callerIsAllowed = mApiPermissionHelper.isAllowedIgnoreErrors();
+                boolean callerIsAllowed = (match == EMAIL_STATUS_INTERNAL) || mApiPermissionHelper.isAllowedIgnoreErrors();
```
dschuermann: yup, makes sense to me.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/OpenPgpServiceKeyIdExtractor.java
```diff
@@ -53,150 +47,223 @@ private OpenPgpServiceKeyIdExtractor(ContentResolver contentResolver,
...
+                throw new IllegalStateException("Internal error, received null cursor!");
+            }
+
+            try {
+                while (cursor.moveToNext()) {
+                    String queryAddress = cursor.getString(INDEX_EMAIL_ADDRESS);
+                    Long masterKeyId = cursor.isNull(INDEX_MASTER_KEY_ID) ? null : cursor.getLong(INDEX_MASTER_KEY_ID);
+                    int verified = cursor.getInt(INDEX_EMAIL_STATUS);
+
+                    KeyRow row = new KeyRow(masterKeyId, verified == 2);
```
dschuermann: this is a little bit difficult to understand, can you make it more clear or add a comment?

dschuermann: To make my comment more clear ;) It is difficult to see if the new code in ``OpenPgpServiceKeyIdExtractor`` has the provides the same functionality and works for the same corner cases as before.

Valodim: Sure, I'll revisit the code and add a comment or make it more readable.

Note that this behavior was nailed down with tests before the changes, so there is some confidence that the special cases (no keys, duplicate, missing keys) are handled equivalently. I tried to keep the tests intact except for the "assertion" part, which is more readable with the more explicit output now.

dschuermann:
After this, I will do a new beta

Valodim:
Rebased this on current master, and I tried to make the code in `OpenPgpServiceKeyIdExtractor` more readable

dschuermann:
Travis does not find the commit in the submodule, can you look into it?

PR #2093, opened 2017-04-24
Key Status Card

Valodim:
This is a follow-up to #1450. It adds a "key status" card to key view, that displays a summary of key health. I think it came together rather nicely.

I'm fairly happy with the summary info at this point, and if there are problems with the primary key (revoked,expired, insecure, ...) then subkey state is never displayed. The per-subkey state could use some more work, and there is no distinction yet between the user's own keys or other keys.

One more thing we should also rethink is the status string up in the toolbar, which is now redundant information (see screenshots).

## Types of changes


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSecurityConstants.java
```diff
@@ -117,26 +134,53 @@ public static boolean isSecureHashAlgorithm(int id) {
             TeleTrusTNamedCurves.getOID("brainpoolP512r1").getId()
     ));
 
-    public static boolean isSecureKey(CanonicalizedPublicKey key) {
-        switch (key.getAlgorithm()) {
+    static KeySecurityProblem checkForSecurityProblems(CanonicalizedPublicKey key) {
+        // TODO: checks on master key?
```
dschuermann: can you explain this TODO?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignatureChecker.java
```diff
@@ -134,10 +137,11 @@ boolean initializeOnePassSignature(Object dataChunk, OperationLog log, int inden
     }
 
     private void checkKeySecurity(OperationLog log, int indent) {
-        // TODO: checks on signingRing ?
```
dschuermann: has the other TODO something to do with this one? 😬 
I think currently it is only checked on the signing subkey, but not if the master key is insecure.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/widget/SubkeyStatusLoader.java
```diff
@@ -0,0 +1,182 @@
+package org.sufficientlysecure.keychain.ui.widget;
```
dschuermann: license ;)

Valodim:
Agreed on all three points, but I would not do any of that in this PR. It's large enough as it is :)

Oh one of the things I'm not happy with anymore is the term "Special" for partially broken keys. This should really be something more negative, since the keys really are partially defunct. I couldn't come up with anything good to describe this in one or two words. Any ideas? Just "Broken" seems too blunt. Oh also, need to change that icon.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSecurityConstants.java
Valodim: no :)

I suppose what I had in mind was "check primary key, not only signing subkey", but that doesn't really belong here.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpSignatureChecker.java
Valodim: yes :)

and I think this is the better place for the todo, too.

dschuermann:
What about "Key has Issues" or just "Issues" instead of "Special"?
And this icon: https://materialdesignicons.com/icon/heart-broken 😬

Valodim:
Haha, I like the icon. I called the status "Defective" for now because I couldn't think of anything better. This only occurs for keys where the best candidate signing or encryption subkey (but not the primary key) is expired or revoked, which should be super rare in practice.

dschuermann:
LGTM for merge

Valodim:
removed status text from toolbar for revoked/expired/insecure keys

PR #2097, opened 2017-04-29
Security Warnings API

Valodim:
This PR adds support to the api to display extended information about "insecure" decryption/verification state. It does so by adding a PendingIntent result that can be called to display a dialog. The client app can also indicate that it supports an "override" mechanism, which will show a suppress button in the dialog. If the user does this, the client app will be informed of this right there, and also receive extra in the result if the same security problem is encountered at a later point.

Two things missing still: some security problems have no identity yet, which is no big deal. the bigger issue is the wording and layouts, which I'm sure can be improved in many ways. I'm not quite sure how to approach this :)

## How Has This Been Tested?
Only manually - there's no unit tests yet :(

## Types of changes


Valodim:
I worked on the wording some more and I think it's good enough now, also added identity (by session key). Also added a bunch of tests for insecure symmteric cipher, no mdc and rsa 1024.

Valodim:
no longer WIP

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
```diff
@@ -306,6 +314,11 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                     return;
                 }
             */
+            case 20:
+                db.execSQL(CREATE_OVERRIDDEN_WARNINGS);
+                if (oldVersion == 20) {
+                    return;
+                }
```
dschuermann: currently, (As far as I understand) if someone is upgrading from 18 or 19 to 21, your SQL will never be executed. We really need to fix our upgrade method...

dschuermann: It is already returning in ``case 19``

Valodim: nice catch! we had the same problem with an upgrade from version 14 to anything newer! I guess we were lucky this never came up before

dschuermann: So we should also remove that return in ``case 14`` right?

Valodim: oh, whoops. I had done that but not staged it for the commit. fixed.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
```diff
@@ -316,7 +308,8 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
             */
             case 20:
                 db.execSQL(CREATE_OVERRIDDEN_WARNINGS);
-                if (oldVersion == 20) {
+                if (oldVersion == 14 || oldVersion == 18 || oldVersion == 19 || oldVersion == 20) {
```
dschuermann: If we are coming from 14 we want a consolidate 😬

PR #2099, opened 2017-04-30
Security token: Show bind activity when importing key from file

### LINKED ISSUE #2098: Yubikey: Wizard should allow to bind token after public key is imported (from file)
I understood why I thought the import of a Yubikey failed (in #2075): the wizard brings you back to the key list.

## Expected Behavior
After importing the public key (from a file), the wizard should show you the `ViewKeyActivity` which shows the button "Bind key"

## Current Behavior
At the moment, you're brought back to the main list. This may lead users to think the import went wrong (#2075), because only the public key is imported.

**Workaround:** Go through the "Manage my keys" wizard one more time, and `ViewKeyActivity` will be displayed in lieu of step 3.

## Possible Solution
Modify `ImportKeysActivity.handleResult()` to redirect the activity to `ViewKeyActivity` instead of going back to the main list, in this particular case.

1. Go to "Manage my keys" (or start the first time usage wizard)
1. Tap "Use security token"
1. Select "Import key information from file" and select the file
1. After the key data is display, tap "import"

Reproduction rate: 10/10

## Context
I thought the usage of a Yubikey was impossible in my case, and it seems I'm [not the only one](https://github.com/open-keychain/open-keychain/issues/1681#issuecomment-280689658) 😃 

## Your Environment

* Android Version: LineageOS 14.1, 20170418
* OpenKeychain Version: Debug version built from master, which is currently at 6c2c8b248073864f9750e7a4dbd54c5282aa7dfd. This is also reproducible from the Google Play beta version (still 4.4-beta1)
* Yubikey Firmware: 4.3.4

JohanLorenzo:
Fixes #2098.

## How Has This Been Tested?
Tested manually. Steps are the same described in #2098.

## Types of changes

## Note

Changes were heavily copied from `BaseSecurityTokenActivity.onSecurityTokenPostExecute()`. I don't mind refactoring these changes to avoid the duplication, but I would love to know where to put the common parts.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateSecurityTokenImportResetFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/CreateSecurityTokenImportResetFragment.java
@@ -229,7 +229,7 @@ public class CreateSecurityTokenImportResetFragment
 
     public void importFile() {
         Intent intent = new Intent(getActivity(), ImportKeysActivity.class);
-        intent.setAction(ImportKeysActivity.ACTION_IMPORT_KEY_FROM_FILE);
+        intent.setAction(ImportKeysActivity.ACTION_IMPORT_KEY_FROM_FILE_AND_BIND);
         startActivity(intent);
     }
 
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -36,6 +36,11 @@ import org.sufficientlysecure.keychain.keyimport.processing.ImportKeysOperationC
 import org.sufficientlysecure.keychain.keyimport.processing.LoaderState;
 import org.sufficientlysecure.keychain.operations.ImportOperation;
 import org.sufficientlysecure.keychain.operations.results.ImportKeyResult;
+import org.sufficientlysecure.keychain.pgp.exception.PgpKeyNotFoundException;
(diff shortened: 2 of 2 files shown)


JohanLorenzo:
```java
org.sufficientlysecure.keychain.pgp.UncachedKeyringTest > testArmorIdentity FAILED
    java.lang.RuntimeException
        Caused by: java.lang.NegativeArraySizeException at UncachedKeyringTest.java:81

org.sufficientlysecure.keychain.pgp.UncachedKeyringTest > testDecodeEncodeMulti FAILED
    java.lang.Exception
        Caused by: java.lang.NullPointerException at UncachedKeyringTest.java:125
```

Hmmm, I'm not sure whether I broke these tests. If I did, I currently don't see how 😕

Valodim:
Nice, thanks! The build error was just our unit tests acting up. I restarted the travis process, looks good now.

Code looks fine but I didn't get to test yet

Valodim:
Hmm just tested this and it didn't work for me. The call to `securityTokenHelper.getFingerprints()` in handleResult tries to communicate with the security token to obtain the fingerprint, which doesn't work. But that mechanism feels wrong in the first place - we already now the expected fingerprint from the previous dialog, that information should be passed in via an extra.

Going the extra mile here, we could do a check for the expected fingerprint among imported keys, display only that key or an error if it's not actually in the selected file, and change the button to "Import and Bind to Token" (maybe find better wording, heh) to match user expectations.

Just passing the `EXTRA_SECURITY_TOKEN_*` values through would work just fine as well though I reckon. Thanks for your work so far! :) :+1:

Valodim:
superseded by #2157

PR #2103, opened 2017-05-08
TYPO: singing replaced to signing

atsb:
Replaced singing with signing

I have replaced the incorrect word 'singing' to 'signing'

## Description
Previously singing was present in the text.  I corrected this to signing.

## Motivation and Context
Correct spelling helps make the program look more professional.  I use this on a regular
basis and I want to help make it better.

## How Has This Been Tested?
Plain-text change that doesn't require recompilation to test.

## Types of changes
Text change (non-code) change.


dschuermann:
Nice catch :P

PR #2109, opened 2017-05-16
set cache notification to low priority

Valodim:
Super small PR: This sets the notification priority for our passphrase cache service to "min", which (at least on my phone) hides the notification from the lock screen and also the icon bar on top.

This seems appropriate to me, given we require no attention whatsoever from the user. And from what I can tell, it doesn't mess with foreground status of the service.


dschuermann:
nice

PR #2110, opened 2017-05-20
Keyserver status

Valodim:
We *almost* save enough information to know whether keys are published on keyservers or not. This is a thing I had in the back of my mind for a while and I felt like writing it down into code today :)

What this basically does is it saves a flag on whether a key has (ever) been seen on keyservers during import, in addition to the last updated status. This information is displayed on a card, but it's currently not actionable.

This is still rough around the edges and I really didn't think the user workflows through. Is this even useful for users? If so, for what? Should there be actions associated with this? What about different keyservers?

Putting this here for discussion, for now :)

## Types of changes


Valodim:
note to self, add better migration support. at the moment keys that were updated will initially show up as "Not Published", which is obviously incorrect.

Valodim:
I moved the status into the key health card, it is now simply part of "Key Status"

Valodim:
I fixed all remaining issues I had with this: smaller layout issues, wording, loader refresh when key is updated, correct handling of "null" state.

It's still only informational and has no associated actions, but this PR should be good to merge now.

dschuermann:
Code looks great. I also just tested this. One thing I propose before merging this:
When the default keyserver is changed via preferences, all keyserver statuses should be resetted.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainDatabase.java
```diff
@@ -308,7 +308,11 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
             */
             case 20:
                 db.execSQL(CREATE_OVERRIDDEN_WARNINGS);
-                if (oldVersion == 18 || oldVersion == 19 || oldVersion == 20) {
+
+            case 21:
+                db.execSQL("ALTER TABLE updated_keys ADD COLUMN seen_on_keyservers INTEGER;");
```
Valodim: Note to self, check if this column is also in the create table statement

Valodim:
Done. I'm resetting both last updated time and keyserver status entirely now.

Valodim:
updated screenshot:

keyserverstatus

PR #2111, opened 2017-05-23
Auto value

Valodim:
The *Operation classes take parcelables as both their input/output and output. At the moment, those classes aren't properly immutable (in fact we have several hacks in place that rely on their mutability), don't use builder patterns in most cases, and add a *lot* of boilerplate for implementing the Parcelable interface.

This PR introduces the [AutoValue](https://github.com/google/auto/tree/master/value) library to improve on all three of those points. I didn't yet convert all classes, and this needs more testing because it's not a pure semantics-preserving refactoring, but I got rid of a whoopin' 500 lines of pure boilerplate so far while moving many classes to a proper builder pattern.


nmikhailov:
Nice to see auto-value-parcel included. :+1:

Valodim:
Done. I tested all of these parcels in the operations I could think of from UI (and fixed a couple of unrelated bugs in the process), this should be good to go now.

PR #2115, opened 2017-05-27
Advanced password

Valodim:
This PR is a first step towards an app lock centered security model: It moves the "change password" operation from key view overflow menu into a tab under advanced key view. 

I had first considered adding more methods here, but we don't really want to stick with per-key passphrases for much longer anyways :)


Valodim:
Rebased and polished this PR. I think it's a step we can take along the way for #2163, since it deemphasizes lock methods, and gives us more space to later place "please use app lock instead"

PR #2117, opened 2017-05-30
Key transfer using TLS-PSK

Valodim:
Inspired by #2083, this PR introduces a "Secure Device Setup" mechanism, which allows bi-directional transfer of secret key material that is authenticated via qr code, transferred via local network, and forward-secret via tls-psk.

This requires Android 5+, lower versions will display an error screen. The user also gets an error if wifi is disabled, and I tried to gracefully handle other error states.


Valodim:
Special thanks to @tobiasschuelke, @0xmatthias, @svenklaus and @MrJabo for their work on this, working from their PR made this whole thing *much* easier! :+1:

Valodim:
Wrote a short spec for this: https://github.com/open-keychain/open-keychain/wiki/OpenPGP-Key-Transfer

Valodim:
Interoperability with the standalone C version that @dkg is working on, based only on the spec above, succeeded at the first attempt this weekend! :1st_place_medal:

dschuermann:
@Valodim awesome! Can't wait to deploy this!

Valodim:
added a ton more polish: the qr code can now be scanned from the key list's "Scan QR Code" action, the fragment is available in the first time wizard, a "Done" button is shown once a connection is established, connecting has a loading screen and improved timeout error handling, and there is now a warning dialog before any key is actually sent:



Such material.

Also, after discussion with @dkg, the keys are now delimited by their "natural" delimiters, which is the ascii armor markers (`-----BEGIN PGP SECRET KEY BLOCK-----`)

Valodim:
added a specialized error screen if conecting to the wrong wifi:



not super happy with the wording yet, suggestions welcome. this was a suggestion of @dkg's, I gotta say I'm not sure if it's worth it including the SSID in the QR code for the slightly more specific error message myself. then again, the qr code is still at an acceptable size, so why not.

User research showed that people who have a PGP background will look for transfer options in secret key view. Now this doesn't quite match the flow, since the connection is established first, but I just added an entry point to the overflow menu in ViewKeyActivity for secret keys, to pick up those people.

the only thing that's still missing is a page on openkeychain.org that describes the mechanism for users, and that I can link to from the "More Info" button. we can also then link this from our FAQs.

PR #2118, opened 2017-06-01
Fix Constants.java Orbot proxy settings

xorxornop:
Fix settings to correct default Orbot proxy port and type

## Description
Change Orbot proxy port and type to `9050` and `SOCKS`, respectively.

## Motivation and Context
Does not correctly connect to Orbot because the proxy settings do not match those set by default by Orbot. Failure occurs silently; to the user, appears as if Tor connection is not active.

## Types of changes


dschuermann:
We use Orbot's default HTTP proxy instead of its SOCKS proxy. The SOCKS proxy is not fully supported by OkHttp. I cleared up some confusion in https://github.com/open-keychain/open-keychain/commit/0e6ec82823c996e5d88cba025d7db7aba9ef9994

Can you open an issue report about your problems with Orbot? I wasn't able to reproduce them. For me it works with the HTTP proxy.

Issue #2049, opened 2017-02-09
Decryption no longer works with Yubikey Neo: SecureMessagingException


nh2:
When I try to decrypt with my Yubikey Neo over NFC (which in the past, worked fine at some point), I get an error. For example, when using the decrypt-from-clipboard feature, or when using another program like https://github.com/zeapo/Android-Password-Store/issues/271.

## Expected Behavior

Decryption succeeds.

## Current Behavior

I now get `Error: Transceive failed` in the UI and the following stack trace in `adb logcat`:

```
E/Keychain(13043): failed to establish secure messaging
E/Keychain(13043): org.sufficientlysecure.keychain.securitytoken.SecureMessagingException: failed to retrieve secure messaging key attributes
E/Keychain(13043): 	at org.sufficientlysecure.keychain.securitytoken.SCP11bSecureMessaging.establish(SCP11bSecureMessaging.java:298)
E/Keychain(13043): 	at org.sufficientlysecure.keychain.securitytoken.SecurityTokenHelper.connectToDevice(SecurityTokenHelper.java:211)
E/Keychain(13043): 	at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity.handleSecurityToken(BaseSecurityTokenActivity.java:444)
E/Keychain(13043): 	at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity$1.doInBackground(BaseSecurityTokenActivity.java:172)
E/Keychain(13043): 	at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity$1.doInBackground(BaseSecurityTokenActivity.java:162)
E/Keychain(13043): 	at android.os.AsyncTask$2.call(AsyncTask.java:288)
E/Keychain(13043): 	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
E/Keychain(13043): 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
E/Keychain(13043): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
E/Keychain(13043): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
E/Keychain(13043): 	at java.lang.Thread.run(Thread.java:841)
```

## Possible Solution

Not sure, but this is definitely a regression, it worked in the past.

[...]

PR #2122, opened 2017-06-13
fix issue #2049

af-anssi:
Temporary workaround to fix issue #2049.

## Description

The temporary workaround proposed does not rely only on the "secure messaging" bit from the "extended capabilities", but also on the "secure messaging key size" declared in the "extended capabilities" to trigger on SmartPGP secure messaging only if secure messaging is configured for AES (appeared in specification v3) and not DES.

## Motivation and Context

A regression with OpenPGP applets other than SmartPGP that support secure messaging was introduced by commit a6b7b2bf4ec204cd647db9f0bead53476069d34f, including Yubikey NEO (issue #2049).

## How Has This Been Tested?

No test as been done on Yubikey NEO to verify the issue is fixed.

## Types of changes


dschuermann:
Thanks! Interestingly, my test YubiKey did not trigger this, i.e., it did not have the messaging capability.

PR #2131, opened 2017-07-06
Update SecureMessaging/SCP11b wrt OpenPGP card v3.3

af-anssi:
OpenPGP card specification v3.3 introduces a bit for secure messaging using SCP11b. It should now be used to avoid any conflict with the AES secure messaging described in the specification.

## Description
Use the correct secure messaging type to trigger SCP11b establishment.
As the AES key sizes cannot be inferred from the secure messaging type, they are now fixed and determined by the size of the field of the curve used for secure messaging, as recommended in SCP11 specification.

## Motivation and Context
Correction of the existing workaround which can trigger a SCP11b establishment while the token is not supporting it.

## How Has This Been Tested?
Tested on SmartPGP version 1.3 with SCP11b enabled.
Should not impact other cards (the value 03 for the secure messaging type was unused until OpenPGP card v3.3).

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
@@ -358,8 +358,10 @@ class SCP11bSecureMessaging implements SecureMessaging {
                 throw new SecureMessagingException("No key in token for secure messaging");
             }
 
-            final int fieldSize = pkcard.getParams().getCurve().getField().getFieldSize();
+            final EllipticCurve curve = pkcard.getParams().getCurve();
+            final int fieldSize = curve.getField().getFieldSize();
             int keySize;
+
             if(fieldSize < 512) {
                 keySize = 16;
             } else {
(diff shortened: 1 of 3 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SCP11bSecureMessaging.java
```diff
@@ -365,6 +358,14 @@ public static void establish(final SecurityTokenHelper t, final Context ctx)
                 throw new SecureMessagingException("No key in token for secure messaging");
             }
 
+            final int fieldSize = pkcard.getParams().getCurve().getField().getFieldSize();
+            int keySize;
+            if(fieldSize < 512) {
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -206,7 +206,7 @@ public void connectToDevice(final Context ctx) throws IOException {
         mPw1ValidatedForDecrypt = false;
         mPw3Validated = false;
 
-        if (mOpenPgpCapabilities.isHasAESSM()) {
+        if (mOpenPgpCapabilities.isHasSCP11bSM()) {
```
Valodim: so before this change, this routine was triggered on values 1 and 2, now it is triggered only by 3. is this related to the bug you mentioned? can you say a bit more about this?

af-anssi: Before this change, establishement of SCP11b was tried some with tokens that do not support which generated a stack trace and an error message in the log (example [here](https://github.com/open-keychain/open-keychain/issues/2049)), but without any other inconvenient as the execution continued normally. With this change, establishement of SCP11b secure messaging is triggered only with tokens that support it, which is now possible thanks to an update of the OpenPGP card specification (v3.3).

Valodim:
:+1:

PR #2138, opened 2017-07-24
refactor keyserver class structure, use AutoValue for ParcelableHkpKeyserver

Valodim:
what it says on the tin. these changes were made almost entirely with refactoring operations, plus fixing errors that occured here and there. I changed some namings while I was at it, and extracted the actual querying logic from the parcelable data class.

I also noticed the EmailKeyHelper class isn't used... was this ever in use somewhere?


dschuermann:
The class was never used in production. The idea was to prefetch all keys where an email address exists in the addressbook. We can remove that as this is no longer a good idea.

Valodim:
You think so? If we are looking up an email address directly it might be nice to have, especially if we teach add WKD support. I'd leave it in for now.

PR #2145, opened 2017-08-02
Ditch Android Beam feature

Valodim:
## Motivation and Context

I got this stacktrace when opening the (huge!) Edward the GPG Bot key:

```
java.lang.RuntimeException: An error occurred while executing doInBackground()
	at android.os.AsyncTask$3.done(AsyncTask.java:325)
	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
	at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
	at java.util.concurrent.FutureTask.run(FutureTask.java:242)
	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
	at java.lang.Thread.run(Thread.java:761)
Caused by: java.lang.IllegalStateException: Couldn't read row 0, col 0 from CursorWindow.  Make sure the Cursor is initialized correctly before accessing data from it.
	at android.database.CursorWindow.nativeGetBlob(Native Method)
	at android.database.CursorWindow.getBlob(CursorWindow.java:403)
	at android.database.AbstractWindowedCursor.getBlob(AbstractWindowedCursor.java:45)
	at android.database.CursorWrapper.getBlob(CursorWrapper.java:147)
	at org.sufficientlysecure.keychain.provider.KeyRepository.getGenericData(KeyRepository.java:129)
	at org.sufficientlysecure.keychain.provider.KeyRepository.getGenericDataOrNull(KeyRepository.java:93)
	at org.sufficientlysecure.keychain.provider.KeyRepository.loadPublicKeyRingData(KeyRepository.java:310)
	at org.sufficientlysecure.keychain.util.NfcHelper$3.doInBackground(NfcHelper.java:133)
	at org.sufficientlysecure.keychain.util.NfcHelper$3.doInBackground(NfcHelper.java:128)
```
this is caused by a key blob that is too large for the database to handle. The underlying problem is fixed since #2066, however this can apparently still occur for old keys. Regardless, it led me to investigate that piece of code.

## Changes

This PR does two things:

* rewrite of the NfcAdapter class. it is now nice™
* delete the NfcAdapter class, removing the Android Beam feature

I first just wanted to handle that error case by disabling the feature instead of crashing (this should happen very rarely anyways). Then I noticed the class was quite old and ugly, so I rewrote it. And when I tested the feature, I found that it simply doesn't work. The ViewKeyActivity where it is used in has an NfcForegroundDispatcher enabled which listens for security tokens, which entirely overrides the registered NDEF tag.

Android beam is an awkward feature that people very rarely use. We never even received a bug report about this being broken. Time to get rid of it.

But hey, at least the class is very nice after the first commit.

## Types of changes


dschuermann:
nice, your work will never be recognized by future devs ;)

Issue #1681, opened 2016-01-13
Token import fails when token public key is not a the master key


Nekroze:
There is a common issue that occurs with gnupg smartcard style devices when the smartcard holds only subkeys with the master key being seperate/offline.

This [issue has been reported to gnupg issue tracker](https://bugs.gnupg.org/gnupg/issue1828), however, it will likely not be repaired for some time and will likely require pre-existing smartcards to have the keys on the device recreated/replaced.

A better option might be to modify the way you import a token to allow you to specify a file or url to the public key for the whole key not just the key that the card things is the public key.

Due to this issue I cannot use my Yubikey together with some of the great apps that integrate this app for security.

Nekroze:
Reading through the issues, I think this fix can be incorporated into #1587

dschuermann:
I don't really get the issue, could you describe it in more detail?
OpenKeychain is not based in gnupg, we don't share any code.

Nekroze:
My apologies. 

This is a bug in GnuPG that causes a smart card to list a sub key as the public key for all of the keys on that smart card. This is an issue because it seems that openkeychain finds the public key I'd from the smart card and its pub which is a sub key not the primary key so when openkeychain searches the key servers it does not download the full public key and thus fails to import. 

This is a bug it seems with all smart cards and is just the way GnuPG sets them up. As such there are many cards that are out there affected by the bug making their card unusable with openkeychain. 

This bug only affects smart cards storing sub keys alone with the primary GnuPG key stored separately or offline. This strategy has long been the standard for higher safety with GnuPG as storing the primary key in a safe or offline storage device and only using it from an offline computer when the sub keys need renewing is a solid security strategy that makes it feasible to keep your primary key for a very long time. 

Unfortunately this bug is punishing users who adopt this higher security configuration way more then an inability to use the fetch command from `gpg - - card-edit` in that openkeychain cannot import the smart cards key at all. 

While this is a bug in GnuPG itself I think we should address it in openkeychain as a patch to fix this will take a long time to become ubiquitous and does nothing to help people who already set their smart card up and are not in a position to wipe the smart card and start again. 

Does that make more sense? I am not good at making sense sadly.

amingilani:
I have the same problem. Adopting higher levels of security, I generated and saved three subkeys on my Yubikey specifically for the token.

I cannot seem to import these from OpenKeyChain. Despite having pushed the key to a keyserver, and having imported it manually via a file.

[...]

Issue #1587, opened 2015-11-11
Redesign Security Token Import (Yubikey etc)


Valodim:
Multiple users have expressed confusion about the import yubikey dialog. A couple of points have come up in irc while discussing the issue:
- The matching public key seems to not be up on keyservers at import time quite often. This either happens because it was created very recently and just isn't there yet (this delay is a general point of confusion with keyservers), or because people don't use keyservers.
- There do seem to be people around who have urls on their yubikeys which we can use. This is encouraged in sjoseffson's [blog post on yubikeys](http://blog.josefsson.org/2014/06/23/offline-gnupg-master-key-and-subkeys-on-yubikey-neo-smartcard/), and quite a few people seem to follow this.
- Users are _not_ aware of what the activity is actually trying to do, which is import the public key. Even users who would have no problem manually importng their public key by some means miss that, simply because we do not indicate it.

I have a suggestion in mind for a design overhaul, will write about it later.

Valodim:
Here's the UI flow I had in mind: The import dialog does not initially offer the user a choice of where to obtain the public key from, but progresses through a list of possible sources. I envision this similar to the UI in the GitHub Linked Identity creation activity, here's a mockup:

![import yubikey mockup](https://dl.pushbulletusercontent.com/UFdUBjyNwWFiDmaeJgFuZplTuYBXuXIA/layout-2015-11-11-161503.png)

The options to "Retry" (makes sense e.g. due to keyserver connection errors) and "From File" show up when all attempts failed.

Things left:
- how can the user "reset" his yubikey and/or create a new key for it?

s0crat:
@Valodim , I'm in the case you suggest in your first message : 
- yubikey with subkeys stored on it
-  URL on yubikey that point to the public master key

Have you been able to deal with this specific import case ?

jdanders:
After two days of searching trying to figure out why my key wouldn't import from my yubikey, I finally found this page. Your opening description perfectly matches my situation -- I presume it's failing because I have not used a keyserver. Before messing around with the yubikey I had no idea what a keyserver was, and I do not really want to upload my key to one because I am still just messing around trying to learn and apparently there's no "delete" button on the keyserver.

I don't entirely understand the PGP approach to life, but if you could add a way to allow my key to be loaded without upload to a keyserver, I would definitely appreciate it. And the app definitely does not help an ignorant user like myself understand why it is failing.

To help google find this page, I'm going to paste my exact error below:
```No keys found!```
[...]

PR #2157, opened 2017-09-05
new Security Token import flow

Valodim:
This is a rewrite of our security token import ui flow. It assumes more things than the previous, and uses specialized import mechanisms instead of reusing the key import fragment.

This follows up on the ideas in #1587, and hopefully fixes (or makes easily fixable) the issues mentioned in #1681. It adds support for fetching from the security token uri, has a nice logging capability and hopefully provides an extensible basis for more features in this direction.

## What's missing?

- [x] consider entry points to this ui flow
- [x] remove "Bind Key" in view key activity
- [x] finish "reset security token" flow
- [x] deal with blank token
- [x] deal with non-trivial key constellations
- [x] test with usb tokens

There is probably more that I'm not thinking of. I'm not sure all of these need to be done, either - I think this really isn't missing much until it's an improvement over what we have, so I would aim for a quick beta release to gather feedback and find bugs early.

## Screenshots:
output

## Types of changes
- ✅ New feature


Valodim:
This PR is really coming along quite nicely! I added a bunch more bells and whistles, dealing with more token states. In particular, the dialog can now be used to unlock a locked token:

output

Valodim:
I added handling of blank tokens, and made sure everything played well with usb tokens. This PR is *almost* done at this point, I only need to make sure this works well with non-standard subkey constellations (e.g. token signing key as subkey of offline primary key).

Issue #1844, opened 2016-04-29
Security Token: Admin PIN handling


dschuermann:
We assume that the Admin PIN is 12345678 by default when creating new keys or moving existing keys to security tokens.
A problem can occur when the token has no keys on it, but the user already changed the Admin PIN or the token has been shipped with a different Admin PIN.

Thus, we should let the user input the Admin PIN if 12345678 did not work. I would strongly favor for a solution that still uses 12345678 by default to keep user interaction to a minimum.

top-gun007:
@dschuermann where can I learn openpgp as I really wish to contribute in this project

dschuermann:
Not important for now, closing.

PR #2160, opened 2017-09-12
remove "Move to Card" action from subkey edit ui

Valodim:
This PR removes the "Move to Card" feature from the subkey edit UI. This is due to a number of reasons:

- as implemented, the operation invariably led to an NPE. from what I could see in the git history, this has been the case for a long time, so it's unlikely this feature was actually used
- beyond the crash, this would only work if the admin pin was set to 12345678, and we don't have a very good concept yet of what to do if this isn't true (see also #1844). this is not quite trivial since we only have three attempts at wrong pins. consequently, assuming it's the default first will use up one of these attempts, but assuming otherwise (which should rarely be the case) makes the operation as a whole much less user friendly.
- even if this had worked as intended, the feature would have been very difficult to use correctly, and dangerous if used incorrectly, e.g. involuntarily deleting subkeys in openkeychain or overwriting them on the token

If we do want to support this, we should probably spend time on the whiteboard to think it through some more. From my point of view though, setting up subkeys and security tokens in a custom manner seems like a use case we can safely defer to GunPG.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EditKeyFragment.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/EditKeyFragment.java
@@ -17,6 +17,9 @@
 
 package org.sufficientlysecure.keychain.ui;
 
+
+import java.util.Date;
+
 import android.app.Activity;
 import android.content.Intent;
 import android.database.Cursor;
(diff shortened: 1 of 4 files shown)


Valodim:
relevant NPE, for reference:

```
Caused by: java.lang.NullPointerException: Attempt to get length of null array
    at org.sufficientlysecure.keychain.securitytoken.SecurityTokenHelper.modifyPin(SecurityTokenHelper.java:256)
    at org.sufficientlysecure.keychain.ui.SecurityTokenOperationActivity.doSecurityTokenInBackground(SecurityTokenOperationActivity.java:276)
    at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity.handleSecurityToken(BaseSecurityTokenActivity.java:435)
    at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity$1.doInBackground(BaseSecurityTokenActivity.java:146)
    at org.sufficientlysecure.keychain.ui.base.BaseSecurityTokenActivity$1.doInBackground(BaseSecurityTokenActivity.java:136)
```

PR #2162, opened 2017-09-19
Reformat public key blocks when loaded from clipboard

### LINKED ISSUE #2161: Copy And pasting keys issue
I try and copy&paste a pgp key and it never reads the paste from Android. Is there a specific text editor I could use and what exact format does a pgp key and a pgp msg look exactly like?

Valodim:
This PR adds some logic that attempts to reformat public key blocks read from the clipboard. It checks for patterns in the first twelve bits of encoded key data, so it makes more assumptions than previously about the structure of these blocks.

This allows recovering keys from arbitrary whitespace changes. It also drops ascii armor headers, but we never use those for public keys so that shouldn't be an issue.

Fixes #2161

## How Has This Been Tested?
Unit tests included

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpHelper.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/PgpHelper.java
@@ -44,6 +44,7 @@ public class PgpHelper {
     public static final Pattern PGP_PUBLIC_KEY = Pattern.compile(
             ".*?(-----BEGIN PGP PUBLIC KEY BLOCK-----.*?-----END PGP PUBLIC KEY BLOCK-----).*",
             Pattern.DOTALL);
+    private static final Pattern KEYDATA_START_PATTERN = Pattern.compile("\\s(m[A-Q])");
 
     /**
      * Fixing broken PGP MESSAGE Strings coming from GMail/AOSP Mail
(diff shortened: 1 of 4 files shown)


dschuermann:
Hmmm. Currently the fix works only for keys from clipboard, but also encrypted data, or keys from files could be mangled. Couldn't we change https://github.com/open-keychain/bouncycastle/blob/openkeychain-master-old/pg/src/main/java/org/bouncycastle/bcpg/ArmoredInputStream.java and skip whitespaces in general?

Valodim:
Whitespaces are already skipped, that was never the issue. The issue is with mangled header lines, in particular the empty line that separates the begin line from the data. We have a real loss of information here, and without any information about the structure it's dangerous to guess the original structure of the mangled message.

```
-----BEGIN-----
Version: GnuPG 6

base64datadatadata
lalala=
-----END-----
```

vs 

```
-----BEGIN----- Version: GnuPG 6 base64datadatadata lalala= -----END-----
```

This works for public keys because a) we don't use the headers for anything (unlike with pgp messages), and b) we can make assumptions about the keydata (see comment in diff). It might be possible to do this for other message types as well, but we can't just trivially reuse the same method.

I'd put this in a release first and see how that goes, then maybe improve input behavior for the other cases later.

dschuermann:
Ah, thanks for the explanation.

PR #2166, opened 2017-09-21
Fix EC decipher template

af-anssi:
This commit reflects a [recent commit](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=commitdiff;h=ff7ccd284c327a5b1c89603f157089177dac9d13) in GnuPG to fix incorrect encoding of data sent to a security token for the deciphering using ECC with large parameters (Brainpool 512 or NIST p521).

## Description

For large EC parameters (Brainpool 512 or NIST p521) the length of TLV encoded data for the decipher command is encoded on two bytes instead of one. It is actually systematically encoded on one byte in the existing code, which is wrong. This wrong encoding scheme, also present in GnuPG, as been [recently fixed](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=commitdiff;h=ff7ccd284c327a5b1c89603f157089177dac9d13).

## Motivation and Context

With large EC parameters (Brainpool 512 or NIST p521) the length of TLV encoded data for the decipher command sent to a security token is above 127 bytes so it must be encoded on two bytes and not just one.

## How Has This Been Tested?

It has been tested with SmartPGP applet (version 1.9).
This change breaks compatibility with previous versions of the SmartPGP applet, and possibly other applet/tokens, but only when a deciphering key on Brainpool 512 or NIST p521 is used.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
@@ -297,6 +297,7 @@ public class SecurityTokenHelper {
         }
 
         byte[] data;
+        byte[] dataLen;
         int pLen = 0;
 
         X9ECParameters x9Params;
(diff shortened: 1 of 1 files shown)


dschuermann:
Thanks! Just a small nitpick: Please add brackets around the if else clauses.

PR #2167, opened 2017-09-22
Ditch "consolidate" mechanism

Valodim:
I have been annoyed with the consolidate mechanism for a while now. We used to use it for migrations, but this has always caused problems and we have actually had it disabled for a very long time now. Now a bug was introduced that lost public keys during the consolidate operation (fixed in 318005dbb6318a68bc9728aa35920bf603084b35), and this finally got me to sit down and get rid of it.

Other than migration, the other use of consolidate was to refresh the trust status of all keys after the user imported or deleted some secret key. With 200 keys this took an unacceptably long time (two minutes-ish?) on my Nexus 5X, where interrupting the operation caused it to start over at best.

To replace this, I introduced a table that maintains the key ids of all foreign signatures, for all keys. That way, we can easily list keys that were signed by a particular key, and reimport only those keys that were actually signed by the changed secret keys. The method that does this is much simpler than consolidate.

This is a WIP, it's still missing logging and polish.


Valodim:
I think this is done now. Losing some 400 lines of code, and the amount of keys that need to be reimported when a secret key is added or deleted should be dramatically lower now. :+1:

Please test & merge.

PR #2172, opened 2017-09-29
WIP: SSH Authentication API

hagau:
## Description
add SSH authentication API and implementation

## Types of changes


dschuermann:
some todos:
* include openssh-lib as folder instead of submodule
* unit tests for operation
* renaming of operation etc.
* inform me if I should publish a version of the API lib on JCenter

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ui/SelectAuthKeyIdActivity.java
```diff
@@ -0,0 +1,134 @@
...
+
+import org.openintents.openpgp.util.OpenPgpApi;
+import org.openintents.ssh.SSHAgentApiConstants;
+import org.sufficientlysecure.keychain.Constants;
+import org.sufficientlysecure.keychain.R;
+import org.sufficientlysecure.keychain.operations.results.OperationResult;
+import org.sufficientlysecure.keychain.ui.base.BaseActivity;
+import org.sufficientlysecure.keychain.util.Log;
+
+public class SelectAuthKeyIdActivity extends BaseActivity {
```
Valodim: please rewrite this based off RemoteDeduplicateActivity. looks way nicer :)

Valodim:
Also, this branch is rather cluttered at this point. Please rebase it off current master

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/SSHAuthenticationOperation.java
```diff
@@ -0,0 +1,296 @@
...
+        long opTime = 0;
+        long startTime = System.currentTimeMillis();
+
+        updateProgress(R.string.progress_extracting_signature_key, 0, 100);
+
+        byte[] signature = null;
+
+        byte[] challenge = sshAuthenticationParcel.getChallenge();
+        if (challenge == null) {
+            Log.d(TAG, "No challenge!");
```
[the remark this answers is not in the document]
hagau: AuthenticationParcel.getChallenge is `@Nullable` (since the getters in PgpSignEncryptInputParcel are too and PgpSignEncryptOperation handles these errors; I used those as reference)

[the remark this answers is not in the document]
Valodim: If null is not a useful input value, why is it nullable? :)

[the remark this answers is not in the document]
hagau: Good point, I will move the error handling for the challenge to SSHAgentService

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ui/SelectAuthKeyIdActivity.java
Valodim: Actually, looking over that activity I noticed it's far easier to do it without the use of fragments. I'll look into that soonish. Put this on hold for now.

hagau: Already done :). I did a simple copy&paste RemoteDeduplicateActivity, but it doesn't overlay properly over the calling app.

Valodim: heh, alright then.

as for the overlay issues: set the of the activity to transparent in the manifest file :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/SSHAuthenticationOperation.java
```diff
@@ -0,0 +1,296 @@
...
+ * <p/>
+ * The operation of this class takes a SSHAuthenticationParcel
+ * as input, and signs the included challenge as parametrized in the
+ * SSHAuthenticationData object. It returns its status
+ * and a possible signature as a SSHAuthenticationResult.
+ * <p/>
+ *
+ * @see SSHAuthenticationParcel
+ */
+public class SSHAuthenticationOperation extends BaseOperation<SSHAuthenticationParcel> {
```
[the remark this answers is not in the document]
hagau: Done

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -252,6 +253,49 @@ public PGPSignatureGenerator getCertSignatureGenerator(Map<ByteBuffer, byte[]> s
...
+        // We explicitly create a signature creation timestamp in this place.
+        // That way, we can inject an artificial one from outside, ie the one
+        // used in previous runs of this function.
+        if (creationTimestamp == null) {
+            // to sign using nfc PgpSignEncrypt is executed two times.
+            // the first time it stops to return the PendingIntent for nfc connection and signing the hash
+            // the second time the signed hash is used.
+            // to get the same hash we cache the timestamp for the second round!
+            creationTimestamp = new Date();
+        }
```
dschuermann: As far as I understand: You do not use creation time for authentication, thus we can get rid of ``creationTimestamp`` parameter and this whole block.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -252,6 +253,49 @@ public PGPSignatureGenerator getCertSignatureGenerator(Map<ByteBuffer, byte[]> s
...
+
+        PGPContentSignerBuilder contentSignerBuilder = getContentSignerBuilder(hashAlgo, signedHashes);
+
+        int signatureType;
+        if (cleartext) {
+            // for sign-only ascii text (cleartext signature)
+            signatureType = PGPSignature.CANONICAL_TEXT_DOCUMENT;
+        } else {
+            signatureType = PGPSignature.BINARY_DOCUMENT;
+        }
```
dschuermann: Get rid of ``cleartext`` parameter and always use binary.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -252,6 +253,49 @@ public PGPSignatureGenerator getCertSignatureGenerator(Map<ByteBuffer, byte[]> s
...
+
+        try {
+            PGPAuthSignatureGenerator signatureGenerator = new PGPAuthSignatureGenerator(contentSignerBuilder);
+            signatureGenerator.init(signatureType, mPrivateKey);
+
+//            PGPSignatureSubpacketGenerator spGen = new PGPSignatureSubpacketGenerator();
+//            spGen.setSignerUserID(false, mRing.getPrimaryUserIdWithFallback());
+//            spGen.setSignatureCreationTime(false, creationTimestamp);
+//            signatureGenerator.setHashedSubpackets(spGen.generate());
```
dschuermann: remove commented lines.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
hagau: Yes, thanks. I have now removed unecessary timestamp parameters in other places too.

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -955,6 +959,19 @@
             </intent-filter>
         </service>
 
+        <!-- SSHAgent Remote API, this service has explicitly no permission requirements
+                     because we are using our own package based allow/disallow system -->
+        <service
+            android:name=".remote.SSHAgentService"
+            android:enabled="true"
+            android:exported="true"
+            android:process=":remote_api_3"
```
dschuermann: we cann call this ``:remote_ssh_api``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
```diff
@@ -0,0 +1,341 @@
...
+import org.sufficientlysecure.keychain.service.input.CryptoInputParcel;
+import org.sufficientlysecure.keychain.service.input.RequiredInputParcel;
+import org.sufficientlysecure.keychain.ssh.AuthenticationParcel;
+
+import java.security.PublicKey;
+import java.util.Date;
+import java.util.HashSet;
+
+
+public class SSHAgentService extends Service {
```
dschuermann: Regarding the naming: I would propose will call this ``SSHService``

dschuermann: Or better: ``SshService``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
```diff
@@ -0,0 +1,341 @@
...
+            return errorResult;
+        }
+    }
+
+    private Intent executeIntentInternal(Intent intent) {
+        Log.d(TAG, "===>>> executeIntentInternal 0: " + intent);
+        switch (intent.getAction()) {
+            case SSHAgentApiConstants.ACTION_SIGN:
+                return sign(intent);
+            case SSHAgentApiConstants.ACTION_SELECT_KEY:
```
dschuermann: same for the API: I would propose we simply call this ``SshApi``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -756,6 +756,126 @@ public String getUserId() throws IOException {
         return signature;
     }
 
+    /**
+     * Call INTERNAL AUTHENTICATE command and returns the MPI value
+     *
+     * @param hash the hash for signing
+     * @return a big integer representing the MPI for the given hash
+     */
+    public byte[] calculateAuthSignature(byte[] hash, int hashAlgo) throws IOException {
```
dschuermann: to prevent code duplication: everything that is similar to what we have in ``calculateSignature`` should move into a different method, like the construction of the ``dsi``

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
Valodim: Well, the service is not doing ssh, it's doing ssh agent or ssh auth stuff. I would propose SshAgentService or SshAuthService

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
hagau: I agree, I'm working on this right now and will move the constants into a new class SSHApi and add some helper functions for versioning and async execution.
(Altough, SSHApi sounds a bit generic, but it seems the best tradeoff)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
dschuermann: I think agent is the term that is used on desktop system. No need to use it in our implementation. The API itself is the SSH API living in the org.openintents.ssh package, so that is why I propose to simply call it SshService because it is implementing this SSH API. Maybe in the future it could do more than just authentication, we never know.

dschuermann: Hmm, maybe I see your pont but then we should rename the whole API.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
Valodim: Acronyms aren't capitalized, please make that SshApi

dschuermann: As an explanation: Bouncy Castle has the coding style to use upper case acronyms. But in Android and in OpenKeychain acronyms are not capitalized. So that's why in some parts of the code it is mixed style.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SSHAgentService.java
Valodim: I don't see how desktop/mobile makes a difference. The purpose of naming is to tell the reader what something does, and "agent" is a perfectly established term for the exact thing this service is doing.

I see the argument about calling it "SshService" because it implements the "SshApi", but my conclusion there would be that "SshApi" is also not a well chosen name, for the same reasons.

dschuermann: I would be okay with ``SshAuthenticationService``, but then I would also rename the package to ``org.openintents.sshauthentication``.

Valodim: sounds good to me :+1: (ftr, I missed your second reply while writing that last one)

hagau: I moving the package to org.openintents.ssh.authentication makes sense regardless of other naming issues (btw. while doing some cleaning, I've been renaming classes and methods from *auth* to *authentication*, local variables can stay *auth*)

dschuermann: Great! I am not sure about ``org.openintents.ssh.authentication`` in comparison to ``org.openintents.sshauthentication``. What do you think @Valodim ?

Valodim: `.ssh.authentication` sounds good to me

dschuermann: That's how it should be!

hagau: SshAuthenticationService makes sense. Renaming the API to sshauthentication-api and SshAuthenticationApi seems unwieldy, but would be logical.
There is of course nothing really specific about SSH to the API (at least right now), really it is just a challenge-response authentication API...

dschuermann: Hmh, I actually like ``SshAuthenticationService``, ``sshauthentication-api`` and ``SshAuthenticationApi``.

As said before I would prefer SSH in the name to make the purpose clear for an outsider, like SSH client developers.

Valodim: I agree. Also, googleability of the term.

hagau: SshAuthenticationService it is, then. For the API itself, I don't know yet :)

dschuermann:
@hagau Can you check in the ssh lib as a subdirectory instead of git submodules? It helps to have them here already for review.

hagau:
@dschuermann I would suggest delaying the review for a couple of days since the sshagent-api lib is not really ready yet. I will include it as a subdirectory when it is stable.

Valodim:
There is no point to that. Please just move it here, regardless of its state.

dschuermann:
@hagau Hey, great work. Can you add a list of TODOs that are left from your side?

hagau:
TODO:
- [ ] api library v1
- [ ] SSH-compatible public key export

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/AuthenticationData.java
```diff
@@ -0,0 +1,71 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.sufficientlysecure.keychain.ssh;
```
hagau: AuthenticationOperation should be moved to .operations
AuthenticationResult should be moved to .operations.results
Where shall this and AuthenticationParcel go?

hagau:
@dschuermann The library should be finished shortly. The name should be 'sshauthentication-api' then?
For the key export I will add a new row to ViewKeyAdvShareFragment.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/AuthenticationData.java
dschuermann: Hmm, not really sure, but I think it's not that important right now. You can leave it in your ssh package.

dschuermann:
@hagau ACK regarding the naming and export.

dschuermann:
Please rebase, I disabled the android tests for now, so Travis at least runs for the unit tests again.

dschuermann:
Arg, this is an unrelated test problem on travis. I just restarted the job.

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/SshAuthenticationApi.java
```diff
@@ -0,0 +1,174 @@
...
+    public static final int SHA384 = 3;
+    public static final int SHA512 = 4;
+    public static final int RIPEMD160 = 5;
+
+    /* signature algorithms */
+    public static final int RSA = 0;
+    public static final int ECDSA = 1;
+    public static final int EDDSA = 2;
+    // TODO: remove?
+    public static final int DSA = 3;
```
hagau: I can't decide whether to keep DSA support in the library. Besides the fact that OpenKeychain doesn't support DSA and OpenSSH has it disabled in default config, I see no reasons to not support it.

dschuermann: I think we still support DSA keys if you import them, but we don't support generation of DSA keys/subkeys. IMHO You can include it.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SshAuthenticationService.java
```diff
@@ -0,0 +1,408 @@
...
+
+        // carries the metadata necessary for signing the challenge
+        AuthenticationData.Builder authData = AuthenticationData.builder();
+        authData.setAuthenticationMasterKeyId(masterKeyId);
+
+        CachedPublicKeyRing cachedPublicKeyRing = mKeyRepository.getCachedPublicKeyRing(masterKeyId);
+
+        long authSubKeyId = 0;
+        try {
+            // get first usable subkey capable of signing
```
dschuermann: capable of authentication

OpenKeychain/src/main/AndroidManifest.xml
```diff
@@ -955,6 +959,19 @@
             </intent-filter>
         </service>
 
+        <!-- SSHAgent Remote API, this service has explicitly no permission requirements
```
dschuermann: SSH Authentication API

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SshAuthenticationService.java
```diff
@@ -0,0 +1,408 @@
...
+        CachedPublicKeyRing cachedPublicKeyRing = mKeyRepository.getCachedPublicKeyRing(masterKeyId);
+
+        String description = "";
+        long authSubKeyId = Constants.key.none;
+        try {
+            authSubKeyId = cachedPublicKeyRing.getSecretAuthenticationId();
+            description += cachedPublicKeyRing.getPrimaryUserId();
+        } catch (PgpKeyNotFoundException e) {
+            // should never happen, since this is always called after existence of the subkey is assured
+            e.printStackTrace();
```
dschuermann: Log.e or maybe throw a AssertionError if this should never happen.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SshAuthenticationService.java
```diff
@@ -0,0 +1,408 @@
...
+    private Intent createErrorResult(int errorCode, String errorMsg) {
+        Log.d(TAG, errorMsg);
+        Intent result = new Intent();
+        result.putExtra(SshAuthenticationApi.EXTRA_ERROR, new SshAuthenticationApiError(errorCode, errorMsg));
+        result.putExtra(SshAuthenticationApi.EXTRA_RESULT_CODE, SshAuthenticationApi.RESULT_CODE_ERROR);
+        return result;
+    }
+
+    private Intent createExceptionErrorResult(int errorCode, String errorMsg, Exception e) {
+        e.printStackTrace();
```
dschuermann: Log.e

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SshAuthenticationService.java
```diff
@@ -0,0 +1,408 @@
...
+        Intent result = mApiPermissionHelper.isAllowedOrReturnIntent(data);
+        if (result != null) {
+            return result;
+        }
+
+        return null;
+    }
+
+    private Intent createErrorResult(int errorCode, String errorMsg) {
+        Log.d(TAG, errorMsg);
```
dschuermann: leftover?

sshauthentication-api/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+
+android {
+    compileSdkVersion 25
+    buildToolsVersion '25.0.2'
+    defaultConfig {
+        minSdkVersion 9
+        targetSdkVersion 25
+        versionCode 1
+        versionName "1.0"
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
```
dschuermann: not needed here

sshauthentication-api/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+    }
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
```
dschuermann: I think we don't need proguard for the lib

sshauthentication-api/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+dependencies {
+    compile fileTree(include: ['*.jar'], dir: 'libs')
+    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
+        exclude group: 'com.android.support', module: 'support-annotations'
+    })
+    compile 'com.android.support:appcompat-v7:25.3.1'
+    testCompile 'junit:junit:4.12'
```
dschuermann: can we remove everything from the lib to keep it small? Maybe even the appcompat is not required.

sshauthentication-api/proguard-rules.pro
```diff
@@ -0,0 +1,25 @@
+# Add project specific ProGuard rules here.
```
dschuermann: remove this file

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/SshAuthenticationApi.java
```diff
@@ -0,0 +1,174 @@
...
+    private Intent createErrorResult(int errorCode, String errorMsg) {
+        Log.d(TAG, errorMsg);
+        Intent result = new Intent();
+        result.putExtra(SshAuthenticationApi.EXTRA_ERROR, new SshAuthenticationApiError(errorCode, errorMsg));
+        result.putExtra(SshAuthenticationApi.EXTRA_RESULT_CODE, SshAuthenticationApi.RESULT_CODE_ERROR);
+        return result;
+    }
+
+    private Intent createExceptionErrorResult(int errorCode, String errorMsg, Exception e) {
+        e.printStackTrace();
```
dschuermann: Log.e

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/SshAuthenticationConnection.java
```diff
@@ -0,0 +1,122 @@
...
+     * @return false if not found or error, else true
+     */
+    private boolean bindToService() {
+        if (mSSHAgent == null) {
+            Intent intent = new Intent(SshAuthenticationApi.SERVICE_INTENT);
+            intent.setPackage(mTargetPackage);
+            boolean result = mContext.bindService(intent, mServiceConnection, Context.BIND_AUTO_CREATE);
+            return result;
+        } else {
+			mOnBoundListener.onBound(mSSHAgent);
```
dschuermann: formatting

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/SshAuthenticationConnection.java
```diff
@@ -0,0 +1,122 @@
...
+		unbind();
+	}
+
+    private void unbind() {
+        mSSHAgent = null;
+        mContext.unbindService(mServiceConnection);
+    }
+
+    public ISshAuthenticationService getService() {
+		return mSSHAgent;
```
dschuermann: formatting

hagau:
TODO:
- [ ] api library v1
 - requires testing
- [ ] SSH-compatible public key export

hagau:
@dschuermann 
- The key export should be done shortly.
- The api library is practically done, but has a lot of untested (but trivial) code that I'm not using right now in either OpenKeychain or ConnectBot.
- I will be reviewing the api lib code tomorrow together with the rest.

dschuermann:
@hagau great thanks. IMHO we are in a state were we can merge the code and you can switch to work on ConnectBot. If we find bugs in the API later, it is fine to patch them in future pull requests.
That also means we can merge first and later publish v1 of the API.

hagau:
@dschuermann I would suggest holding off the merge at least until I have done the review tomorrow and have more confidence in the code.
There are also issues with ConnectBot and Ed25519 X.509 keys (CB uses eddsa-0.1.0, which doesn't support the new ASN.1 encoding for Ed25519 keys, and using eddsa-0.2.0 produces invalid signatures) which might force the format for public keys for the action GET_PUBLIC_KEY to the SSH public key format.

dschuermann:
@hagau OK. We can disable support for Ed25519 for our SSH Authentication API for now.

hagau:
okc_ssh_key_export

Is this acceptable for the key export UI?

dschuermann:
@hagau Yep, totally fine for this use case. Please show a Notify/Toast if the key has no authentication subkey.

hagau:
The SSH export option only appears if there is a autentication key in the selected keyring

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ViewKeyAdvShareFragment.java
```diff
@@ -380,6 +413,24 @@ public void onLoadFinished(Loader<Cursor> loader, Cursor data) {
 
         }
         setContentShown(true);
+        showSshShareIfApplicable();
+    }
+
+    private void showSshShareIfApplicable() {
```
hagau: @dschuermann The SSH export option only appears if there is a authentication key in the selected keyring

dschuermann: great

Valodim: Please show it regardless of auth key, and display an error if there is none. An option that's just missing for complicated reasons breaks user expectations in a very frustrating way.

hagau: Ok, that makes sense. I was concerned it might clutter up the UI with a feature that will not be used by the majority of the users. I have changed it to display a Notify instead.

hagau:
TODO:
- [ ] api lib v1
    [ ] javadoc

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -366,7 +366,7 @@
     <string name="no_keys_exported">"No keys exported."</string>
     <string name="key_creation_el_gamal_info">"Note: only subkeys support ElGamal."</string>
     <string name="key_not_found">"Couldn't find key %08X."</string>
-    <string name="auth_subkey_not_found">"This key does not have a authentication subkey."</string>
+    <string name="authentication_subkey_not_found">"This key does not have a authentication subkey."</string>
```
Valodim: "an"

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/util/key/SshEd25519PublicKey.java
```diff
@@ -65,17 +65,13 @@ public void setAByte(byte[] aByte) {
...
-        if (yIn.length > 32 && yIn[0] == 0) {
-            // BigInteger.toByteArray prepends a byte for sign bit if MSB of Y is set,
-            // but Y will always be positive, so remove the superfluous byte
-            yIn = Arrays.copyOfRange(yIn, 1, 33);
-        }
         byte[] y = new byte[32];
-        for (int i = 0; i < yIn.length; i++) {
-            y[i] = yIn[yIn.length-1 - i]; // convert to little endian
+        for (int i = 0; i < 32; i++) {
+            y[i] = yIn[31 - i]; // convert to little endian
```
Valodim: did you consider using ByteBuffer for this?

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/util/key/SshEd25519PublicKey.java
```diff
@@ -64,11 +66,10 @@ public void setAByte(byte[] aByte) {
     // TODO: remove?
     // encode point according to RFC8032 section 5.1.5
     private byte[] encodePoint() {
-        byte[] yIn = mY.toByteArray();
-        byte[] y = new byte[32];
-        for (int i = 0; i < 32; i++) {
-            y[i] = yIn[31 - i]; // convert to little endian
-        }
+        ByteBuffer buffer = ByteBuffer.allocate(32).order(ByteOrder.LITTLE_ENDIAN);
```
Valodim: `ByteBuffer.wrap(mY.toByteArray()).order(ByteOrder.LITTLE_ENDIAN).array()` not tested but I'd expect that to work

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/util/key/SshEd25519PublicKey.java
hagau: Thanks. Regarding this whole function: Is it correct to take the LSB (x_0) of the big endian coordinate x? Comparing with other implementations this seems to be correct, but RFC8032 doesn't clearly specify it. It does mention in the decoding to set x = p-x if the encoded bit x_0 != x mod 2.

hagau: I could remove the whole function since most users will probably use a crypto library that returns the already encoded point. It's mostly a product of investigating whether bouncycastle returns the correct format needed for SSH public key blobs.

sshauthentication-api/src/main/java/org/openintents/ssh/authentication/util/key/SshEd25519PublicKey.java
hagau: Neither work. After some trial&error I noticed that ByteBuffer doesn't have a concept of little-endianness when it comes to byte[] (since that is a bulk operation). This really requires a loop since little-endian BigIntegers are not a thing in Java apparently.

hagau: I have a test for this function with point data extracted from a public key blob produced by OpenSSH and I'm fairly certain it is correct :). I have not included it since it adds junit&robolectric as dependecies to the api library.

dschuermann:
@hagau What is the state of this PR? Can we merge?

Additional documentation and further fixes can easily be done with new PRs.

PR #2174, opened 2017-09-29
Update support libraries to 25.3.1 and update gradle-witness hashes

hagau:
## Description
Updated support libraries and ran ./gradlew calculateChecksums

## How Has This Been Tested?
Compiles just fine. Deployment to test device apparently needs multidex.

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -9,12 +9,12 @@ dependencies {
     // NOTE: libraries are pinned to a specific build, see below
 
     // from local Android SDK
-    compile 'com.android.support:support-v4:25.0.1'
-    compile 'com.android.support:appcompat-v7:25.0.1'
-    compile 'com.android.support:design:25.0.1'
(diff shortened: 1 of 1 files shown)


dschuermann:
Can you enable multidex in this PR for debug builds only?

OpenKeychain/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     compile 'com.android.support:recyclerview-v7:25.3.1'
     compile 'com.android.support:cardview-v7:25.3.1'
     compile 'com.android.support:support-annotations:25.3.1'
+    compile 'com.android.support:multidex:1.0.1'
```
Valodim: this is pointless if we don't inherit our application class from `MultiDexApplication`. it's also only needed to build with multidex for minsdk 20 or lower, so we can just drop this.

hagau: I apparently did not read the documentation thoroughly enough, thanks.

hagau:
This now passes all tests.

PR #2186, opened 2017-10-13
Refactor OpenPGP applet communication code

Valodim:
Another refactoring PR. Clarity, testability, less magic constants, factories, etc etc.

A bunch of things are changed around, this should probably receive some testing before it gets into a stable release.


dschuermann:
Yubikey NEO and Nitrokey Pro is no longer working for me correctly.
In this branch, when I use the wizard and choose "use security token" I get an "Error: initialization failed".
In current master it works and proceeds to the next screen.
I get the same when signing.

Valodim:
sorry, I should have mentioned that I only did marginal actual testing with the last five commits

dschuermann:
@Valodim Should I test again?

Valodim:
Yes. This should work ok now

dschuermann:
Not sure if this is related to the changes in this PR: After resetting a yubikey, the fragement reloads but again is searching for a key. Instead it should show an empty token. This works after taking away the token and holding it again against the device.

Valodim:
Might be related to the last commit, when the application related data needs to be refreshed

PR #2189, opened 2017-10-18
Disable unsupported USB tokens

dschuermann:
I propose we disable unsupported USB devices for now until we properly support them.

I also worked on the [wiki page about supported tokens](https://github.com/open-keychain/open-keychain/wiki/Security-Tokens) which is now linked from our FAQ page: https://github.com/open-keychain/open-keychain.github.io/commit/53ee5c718af59897dec992ac606e1bd2de932416


Valodim:
I wonder if we should display a "device not supported" error, instead of simply not doing anything with the usb token? That way the user knows they're on the right track and that we do support usb, just not with the particular token they are using.

PR #2190, opened 2017-10-18
Accept dash character (-) in GitHub user names when checking for Gist URL

wiktor-k:
## Description

GitHub allows dash / minus characters in usernames (e.g. "wiktor-k" or
"open-keychain"). This change extends the regular expression to capture
this missing character.

## Motivation and Context

It allows displaying Github Linked Identity resources that contain dash character.

## How Has This Been Tested?

It has not been tested at all. I imagine one character change will not break anything but if it's required I can setup the project and build it to see if it really works :)

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/linked/resources/GithubResource.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/linked/resources/GithubResource.java
@@ -192,7 +192,7 @@ public class GithubResource extends LinkedTokenResource {
             return null;
         }
 
-        Pattern p = Pattern.compile("https://gist\\.github\\.com/([a-zA-Z0-9_]+)/([0-9a-f]+)");
+        Pattern p = Pattern.compile("https://gist\\.github\\.com/([a-zA-Z0-9_-]+)/([0-9a-f]+)");
         Matcher match = p.matcher(uri.toString());
         if (!match.matches()) {
             return null;


dschuermann:
Yep, please test this properly. Not sure about the minus sign in the regex, maybe you need to escape it.

wiktor-k:
@dschuermann Minus sign need to be escaped only if it's surrounded by other characters (not at the end). (see https://stackoverflow.com/questions/154031 unfortunately there is no *official* docs on minus character at the end on Oracle Java site).

An example:

```java
Pattern p = Pattern.compile("https://gist\\.github\\.com/([a-zA-Z0-9_-]+)/([0-9a-f]+)");
Matcher match = p.matcher("https://gist.github.com/Valo-dim/313bb04162420753639f");
if (!match.matches()) {
	System.out.println("does not match");
	return;
}
String handle = match.group(1);
String gistId = match.group(2);
System.out.println("handle: " + handle + "; gistId: " + gistId);
```

Prints:
```
handle: Valo-dim; gistId: 313bb04162420753639f
```
Without this patch it would print `does not match`.

Escaping it there also does no harm, I can add it if you want. Building Open Keychain can take a little bit of time so if you still want me to do that this PR will need to wait.

wiktor-k:
Okay, I did my homework, here is the same key visible in OpenKeychain without the fix, you can see only one Linked GitHub identity:


okc2


Here is the built after applying the fix, you can see also identity with dash in the name:

okc1

dschuermann:
thanks!

PR #2193, opened 2017-10-23
Fix showing of some notifications

hagau:
## Description
Add missing .show() to some Notify.create() calls.

## Types of changes


Valodim:
Nice catch. While you're at it, could you add the `@CheckResult` annotation to the create calls? That should prevent this error in the future.

hagau:
Hm, Android Studio doesn't have a linter for CheckResult and there are no warnings in the build log even when I remove the .show().

Valodim:
Just tested and this works for me. When I remove a `.show()` on a method with `@CheckResult`, I get an error "The result of create is not used" not only just a warning.

hagau:
I mean I don't see the the category in the code inspection results. There is one for javax.annotation.CheckReturnValue but not for android.support.annotations.CheckResult.
I do get the squiggly red underlining in the IDE at least.
Anyway, I have found most missing .show() by greping :)
Have I missed a way to inspect for CheckResult violations?

Valodim:
Not sure. But the warning should be clear enough that this won't happen again

PR #2195, opened 2017-10-23
Change USB filtering

dschuermann:
Instead of filtering by the USB xml, we will move to a method of filtering inside OpenKeychain, so we can display helpful messages to the user.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
@@ -955,10 +955,18 @@ public class SecurityTokenConnection {
         if (transportType == TransportType.USB) {
             tokenType = mTransport.getTokenType();
         } else {
-            tokenType = isFidesmoToken() ? TokenType.FIDESMO : TokenType.UNKNOWN;
+            tokenType = isFidesmoToken() ? TokenType.FIDESMO : TokenType.UNKNOWN_NFC;
         }
 
-        return SecurityTokenInfo.create(transportType, tokenType, fingerprints, aid, userId, url, pwInfo[4], pwInfo[6]);
+        // TODO: bail out earlier on unsupported tokens to not execute other commands
+
+        SecurityTokenInfo info = SecurityTokenInfo.create(transportType, tokenType, fingerprints, aid, userId, url, pwInfo[4], pwInfo[6]);
+
+        if (! info.isSecurityTokenSupported()) {
+            throw new UnsupportedSecurityTokenException();
+        }
+
+        return info;
     }
 
     public static double parseOpenPgpVersion(final byte[] aid) {
@@ -966,4 +974,5 @@ public class SecurityTokenConnection {
         while (minv > 0) minv /= 10.0;
         return aid[6] + minv;
     }
+
 }
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
@@ -76,8 +76,7 @@ import java.util.List;
 public class SecurityTokenConnection {
     private static final int APDU_SW1_RESPONSE_AVAILABLE = 0x61;
 
-    // Fidesmo constants
-    private static final String FIDESMO_APPS_AID_PREFIX = "A000000617";
+    private static final String AID_PREFIX_FIDESMO = "A000000617";
 
     private static final byte[] BLANK_FINGERPRINT = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
 
@@ -91,6 +90,7 @@ public class SecurityTokenConnection {
     private final Passphrase mPin;
     private final OpenPgpCommandApduFactory commandFactory;
 
+    private TokenType tokenType;
     private CardCapabilities mCardCapabilities;
     private OpenPgpCapabilities mOpenPgpCapabilities;
     private SecureMessaging mSecureMessaging;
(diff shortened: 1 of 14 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/usb/UsbTransport.java
```diff
@@ -211,36 +230,36 @@ public TransportType getTransportType() {
     @Override
     public TokenType getTokenType() {
         switch (usbDevice.getVendorId()) {
-            case 4176: {
+            case VID_YUBIKEY: {
```
Valodim: How about `VENDOR_` and `DEVICE_`? I find particularly `PID` misleading

dschuermann: 👍 I changed to ``VENDOR`` and ``PRODUCT`` as in the terminology of USB.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/usb/UsbTransport.java
```diff
@@ -202,6 +222,49 @@ public int hashCode() {
...
+                    case PRODUCT_NITROKEY_PRO:
+                        return TokenType.NITROKEY_PRO;
+                    case PRODUCT_NITROKEY_START:
+                        return TokenType.NITROKEY_START;
+                    case PRODUCT_NITROKEY_STORAGE:
+                        return TokenType.NITROKEY_STORAGE;
+                }
+                break;
+            }
+            case VENDOR_GNUK: {
```
Valodim: this is FSIJ, the Free Software Initiative Japan

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/usb/UsbTransport.java
```diff
@@ -202,6 +222,49 @@ public int hashCode() {
...
 
+    @Override
+    public TransportType getTransportType() {
+        return TransportType.USB;
+    }
+
+    @Override
+    public TokenType getTokenType() {
+        switch (usbDevice.getVendorId()) {
+            case VENDOR_YUBIKEY: {
```
Valodim: `YUBICO` :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java
```diff
@@ -1003,9 +1005,26 @@ public SecurityTokenInfo getTokenInfo() throws IOException {
...
         byte[] pwInfo = getPwStatusBytes();
 
-        return SecurityTokenInfo.create(fingerprints, aid, userId, url, pwInfo[4], pwInfo[6]);
+        TransportType transportType = mTransport.getTransportType();
+        TokenType tokenType;
+        if (transportType == TransportType.USB) {
+            tokenType = mTransport.getTokenType();
+        } else {
+            tokenType = isFidesmoToken() ? TokenType.FIDESMO : TokenType.UNKNOWN_NFC;
+        }
```
dschuermann: @Valodim I found a very strange problem here only happening for Fidesmo cards: In the current order of commands everything works, but if I first get the transport type including ``isFidesmoToken()`` and only afterwards get the fingerprints ``rawFingerprints`` will be null. Do you have an idea what could cause this?

dschuermann:
@Valodim Also, please review.

Currently, the handling of not-supported USB devices is not perfect: I tried for example my smart card reader, which is also class 11, and it will error out much earlier inside UsbTransport.connect().

OpenKeychain/src/main/res/values/strings.xml
```diff
@@ -1580,6 +1580,8 @@
     <string name="security_token_error_header">"Security Token reported invalid %s byte."</string>
     <string name="security_token_error_tag_lost">"Security Token has been taken off too early. Keep the Security Token at the back until the operation finishes."</string>
     <string name="security_token_error_iso_dep_not_supported">"Security Token does not support the required communication standard (ISO-DEP, ISO 14443-4)"</string>
+    <string name="security_token_error_generic">"Generic Error: Most probably, the Security Token has been taken off too early."</string>
```
Valodim: "Generic" is a very technical term

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenInfo.java
```diff
@@ -74,9 +80,41 @@ public static SecurityTokenInfo newInstanceDebugLockedHard() {
...
+        return SecurityTokenInfo.create(TransportType.NFC, TokenType.UNKNOWN_NFC,
                 new byte[][] { KeyFormattingUtils.convertFingerprintHexFingerprint("4700BA1AC417ABEF3CC7765AD686905837779C3E") },
                 Hex.decode("010203040506"), "yubinu2@mugenguild.com", "http://valodim.stratum0.net/mryubinu2.asc", 0, 0);
     }
 
+    public enum TransportType {
+        NFC, USB
+    }
+
+    public enum TokenType {
```
Valodim: Is there a particular reason you put UNKNOWN_NFC and UNKNOWN_USB here redundantly? This information is already available from the TransportType.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
```diff
@@ -948,12 +950,29 @@ public SecurityTokenInfo getTokenInfo() throws IOException {
         String url = getUrl();
         byte[] pwInfo = getPwStatusBytes();
 
-        return SecurityTokenInfo.create(fingerprints, aid, userId, url, pwInfo[4], pwInfo[6]);
+        TransportType transportType = mTransport.getTransportType();
```
dschuermann: @Valodim I found a very strange problem here only happening for Fidesmo cards: In the current order of commands everything works, but if I first get the transport type including ``isFidesmoToken()`` and only afterwards get the fingerprints ``rawFingerprints`` will be null. Do you have an idea what could cause this?

dschuermann:
@Valodim done

Valodim:
I'm still unsure whether the fidesmo check is really necessary. But it's fine if it's in for now.

Also, I changed the "generic error" string to "communication error" on master.

PR #2197, opened 2017-10-25
Fix NPE upon dialog dismissal in RemoteDeduplicatePresenter

hagau:
## Description
Upon dialog dismissal the view is set to null, consequently there is no view when the loader is reset as result of the activity being destroyed.
I found this since I based RemoteSelectAuthenticationKeyPresenter on RemoteDeduplicatePresenter.

## How Has This Been Tested?
Not tested in RemoteDeduplicatePresenter, but works in RemoteSelectAuthenticationKeyPresenter.

## Types of changes


dschuermann:
thanks!

PR #2199, opened 2017-10-26
WIP: GNUK Support (Rebase of #1914)

dschuermann:
see https://github.com/open-keychain/open-keychain/pull/1914


Valodim:
I really can't say for sure whether this behavior is more or less true to the spec, `7.2.9 GET RESPONSE` is very sparse on the details unfortunately. Looking at code that is meant to communicate with the gnuk, I can't find this behavior either: https://github.com/RaymiiOrg/gnuk/blob/2b784cb3b957e83479173d735373a20576515c7e/tool/gnuk_token.py#L227

dschuermann:
@Valodim interesting. Maybe it is not needed. The other related commit in https://github.com/open-keychain/open-keychain/pull/1914 wasn't easy to rebase so I stopped due to time constraints. Maybe we can do a in-person session, reapply the changes, and test some other stuff.

Issue #1937, opened 2016-10-12
Nitrokey Pro / Storage: Key generation (Put key command) fails ("invalid data")


szszszsz:
Application shows USB transmission error when trying to use Nitrokey Storage as a security token.
Please let me know how can I create / extract the logs.

App version: 4.2 beta 1 (42000)
Android 6.0.1, Galaxy S6
Nitrokey Storage v0.43 (OTG USB)
Frequency: Always (5/5)

Scenario:
1. Insert Storage stick
2. Open application
3. Tap `Manage my keys`
4. Tap `USE SECURITY TOKEN`
5. Tap `USE THIS SECURITY TOKEN`
6. Enter arbitrary name, email and PIN and tap `Next`
7. Uncheck `Synchronize with the internet` and tap `CREATE KEY`
-  A Bar with key creation progress is shown
- Image is shown with instructions to holding a security token against the NFC marker
- Red LED on the device blinks a couple of times
- After next 10 seconds the error is shown:

```
Error: USB error - failed to transmit data -1
Take away the Security Token now and touch TRY AGAIN
```
1. After tapping `TRY AGAIN` error is shown again after another 10 seconds delay
2. After reinsertion application crashes but returns to the screen where key is created
3.  Tapping `CREATE KEY` another time red LED on device is flashing for about 10 seconds
4. After the delay same error with `-1` code is shown.

Calling possibly interested: @jans23 @nmikhailov

szszszsz:
Hi! I have just retested the issue on latest version (even if not marked as solved; by the way of other check).
On 4.4.1 (44100) I have another error window now, that the PIN is too short and it should be at least 6 characters (despite entering 9 or later exactly 6 digits). I have reset the token earlier with the menu (there were some keys on it before).
Storage Firmware: v0.47.1

techge:
Same behaviour on Nitrokey Pro.

szszszsz:
Retested by the occasion of #1936 with:
```
4.8-beta.1 (48000)
Nitrokey Pro v0.7
```

Valodim:
wait, nitrokey pro?

szszszsz:
Yes, I assumed the issue cause for both is the same. Would you like me to retest it on Storage anyway?

Valodim:
No, I was just confused because of the title. You're right, this issue happened on all nitrokeys all along.

Valodim:
```
[...]

PR #2201, opened 2017-10-29
Gnuk

Valodim:
This is my take on the gnuk branch. I rebased the voltage selection commit from #1914, and only afterwards noticed that the actual GNUK token I have does support voltage auto selection. I tested with that token, and signing and decryption work as expected with no further changes.

I also added a bunch of tests for CcidReceiver, fixing two bugs in the process.


Valodim:
automatic voltage has been in gnuk since virtually forever (version 0.5 from 2010): https://github.com/RaymiiOrg/gnuk/commit/206b17fef7d7249afc1c92e8cf83c966716cb402#diff-08f862d95a2b3a782e65396928d9f70dR99

Valodim:
This PR, specifically 778fb8e94acd1483c163aefc02517f8676301798, also fixes key import for the Nitrokey Pro. This closes #1937 :)

PR #2206, opened 2017-11-02
More Gnuk improvements

dschuermann:
## Description
* Fix for version check
* Check for life cycle management support and use this to decide if token can be resetted

## Motivation and Context
* Gnuk still not working

## How Has This Been Tested?
* new unit tests
* with all tokens on my table

## Types of changes


dschuermann:
Forgo to mention:
* reset procedure has been tested with GNUK 1.2.6 with life cycle management enabled. It works
* PUT KEY works, so the resulting key is usable. Even with my PIN.

dschuermann:
Gnuk is now working. I needed to prevent it from going into ["admin less mode"](http://www.fsij.org/doc-gnuk/gnuk-passphrase-setting.html#set-up-pw1-pw3-and-reset-code) by setting Admin PIN first and then PIN.

@Valodim Please review.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SecurityTokenOperationActivity.java
```diff
@@ -293,9 +292,12 @@ protected void doSecurityTokenInBackground(SecurityTokenConnection stConnection)
                     mInputParcel = mInputParcel.withCryptoData(subkeyBytes, tokenSerialNumber);
                 }
 
-                // change PINs afterwards
-                stConnection.resetPin(newPin, adminPin);
+                // First set Admin PIN, then PIN.
+                // Order is important for Gnuk, otherwise it will be set up in "admin less mode".
+                // http://www.fsij.org/doc-gnuk/gnuk-passphrase-setting.html#set-up-pw1-pw3-and-reset-code
                 stConnection.modifyPw3Pin(newAdminPin, adminPin);
+                stConnection.resetPw3Validation();
```
Valodim: maybe just do that at the end of modifyPw3Pin?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenInfo.java
```diff
@@ -150,6 +155,60 @@ public static String parseGnukVersionString(String serialNo) {
         if (!matcher.matches()) {
             return null;
         }
-        return matcher.group(1);
+        return new Version(matcher.group(1));
+    }
+
+    public static class Version implements Comparable<Version> {
```
Valodim: this class can be simplified with @AutoValue

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenInfo.java
```diff
@@ -104,20 +107,21 @@ public static SecurityTokenInfo newInstanceDebugLockedHard() {
...
-            TokenType.GNUK_NEWER_1_25
+            TokenType.GNUK_1_25_AND_NEWER
     ));
 
     private static final HashSet<TokenType> SUPPORTED_USB_RESET = new HashSet<>(Arrays.asList(
             TokenType.YUBIKEY_NEO,
             TokenType.YUBIKEY_4,
             TokenType.NITROKEY_PRO,
-            TokenType.GNUK_NEWER_1_25
+            TokenType.GNUK_1_25_AND_NEWER
```
Valodim: reset and put key are now the same set of devices. do we expect this to ever change again? Hopefully not? :)

dschuermann:
@Valodim thanks and done.

Valodim:
lgtm. note we still need to address the problem of cached pins and SecurityTokenInfo objects before a release

dschuermann:
@Valodim ack :)

PR #2210, opened 2017-11-05
Enable multidex for release build

hagau:
## Description
Due to [Add sshauthentication-api v1 support](https://github.com/open-keychain/open-keychain/commit/2619cb1db38f2b9c21a49e2c928cb48d4bc16c8a) multidex is now needed for release build

## Types of changes

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -195,6 +195,8 @@ android {
             minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules-release.pro'
 
+            multiDexEnabled true
+
             // Reference them in the java files with e.g. BuildConfig.ACCOUNT_TYPE.
             buildConfigField "String", "ACCOUNT_TYPE", "\"org.sufficientlysecure.keychain.account\""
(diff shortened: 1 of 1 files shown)


dschuermann:
That is not enough because we target SDK 15, see https://developer.android.com/studio/build/multidex.html

I will see if we can add some more classes to ProGuard.

dschuermann:
😬 https://github.com/open-keychain/open-keychain/commit/1ea5a67f6cafdc78ea8c39b76c95feb1007e010f

hagau:
Ah yes, I didn't consider the minSDK version nor using ProGuard.

dschuermann:
np, I think in the long term we need better ProGuard rules or multidex. Let's see.

Valodim:
What's the issue with using the multidex support lib? It works for minSdk 15, right?

It's nicer to get by without it, of course, but this won't be feasible for very much longer

dschuermann:
[Careful testing](https://developer.android.com/studio/build/multidex.html#limitations) is required, bit yup we could use it in the long term. I read the page yesterday and decided to not go that route for this release.

We could also use much better proguard rules. Currently it is a blacklist of packages that are removed. Normally proguard just does its thing and you declare where it went to far and whitelist some classes. Looking at signal's proguard files here. This also requires some work.

PR #2211, opened 2017-11-06
Nitrokey Start/Storage Support

dschuermann:
* Removes GNUK_UNKNOWN, not used anywhere
* Add support for Nitrokey Start and Storage

## How Has This Been Tested?
Tested with Nitrokey Start and Storage

## Types of changes


Valodim:
Even though full setup is blocked, resetting keys is an option that also shows up for non-blank tokens. I think this more specific error case is still useful to have.

dschuermann:
@Valodim Ah, great, I haven't seen that. Fixed.

PR #2214, opened 2017-11-10
Add/modify helper for getting authentication provider information

hagau:
## Motivation and Context
I'm using these in ConnectBot;while they are rather trivial, they are neccessary in all provider implementations.

## How Has This Been Tested?
Works in ConnectBot.

## Types of changes


dschuermann:
LGTM 👍

PR #2216, opened 2017-11-12
Add tests for RSA, EdDSA & DSA keys in AuthenticationOperation

hagau:
## Description
Add missing tests for RSA, EdDSA & DSA.

## How Has This Been Tested?
Works for RSA & DSA. EdDSA fails, but this is most likely a bug (not in the test, but in the signature generation). 

## Types of changes
- ✅ Bug fix (missing tests)


hagau:
I went through the signature generation for EdDSA trying to find the error, but to no avail (yet).
EdDSA should be disabled for a beta (and the test commented out).

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/AuthenticationOperationTest.java
```diff
@@ -146,13 +242,107 @@ public void testAuthenticate() throws Exception {
...
+        }
+        { // verify signature
+            CanonicalizedPublicKey canonicalizedPublicKey = keyRepository.getCanonicalizedPublicKeyRing(masterKeyId)
+                                                                         .getPublicKey(authSubKeyId);
+            PublicKey publicKey = canonicalizedPublicKey.getJcaPublicKey();
+
+            EdDSAEngine engine = new EdDSAEngine(MessageDigest.getInstance("SHA-512"));
+			engine.initVerify(publicKey);
+			engine.setParameter(EdDSAEngine.ONE_SHOT_MODE);
+			engine.update(challenge);
```
hagau: Hm, apparently this does not work as expected. Feeding SHA512(challenge) works though. I'm investigating further.

hagau:
I suggest delaying merging this until I have implemented a fix for the EdDSA issue, which needs a custom JcaPGPContentSignerBuilder.
This should also be a problem wherever EdDSA is used, since there doesn't seem be a way to not hash the input to the PGPContentSigner. EdDSA (specifically PureEdDSA) always hashes it's input message anyway.

hagau:
See (Fix EdDSA signature generation for authentication & clean up authentication signature generator)[https://github.com/open-keychain/open-keychain/pull/2217] for a fix for the EdDSA issue.

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/AuthenticationOperationTest.java
Valodim: EdDSA itself is parametrized with a hashing algorithm, and unlike other (most?) signature schemes can take arbitrary input, not just fixed length. This led to some confusion for me as well, because the interface was inconsistent - JCA assumes signing algorithms are paired with hashing algos, since for a generic interface you have to assume fixed length. A simple eddsa with sha512 parametrization is more equivalent to "RSAwithNONE" than "RSAwithSHA512", at least in openpgp context it was.

hagau: Yes, and whats makes this worse is that there is a variant HashEdDSA mentioned in [eddsa-20150704.pdf](http://ed25519.cr.yp.to/eddsa-20150704.pdf) (but not in the original paper) that does define a prehash parameter. I was not aware of that until I saw this [issue]( https://github.com/str4d/ed25519-java/issues/34) in ed25519-java for implementing HashEdDSA.

Valodim:
add to #2217

PR #2217, opened 2017-11-12
Fix EdDSA signature generation for authentication & clean up authentication signature generator

hagau:
## Description
- Fix EdDSA signature generation for AuthenticationOperation by using a custom EdDsaAuthenticationContentSignerBuilder to feed the raw input directly to the signature engine without hashing it first (EdDSA hashes it anyway).
- Remove PGP specific code from PGPAuthenticationSignatureGenerator and rename to AuthenticationSignatureGenerator

## How Has This Been Tested?
Works at least for RSA, ECDSA & DSA. See [pull request for tests](https://github.com/open-keychain/open-keychain/pull/2216).

## Types of changes
- ✅ Bugfix
- ✅ Cleanup


dschuermann:
LGTM 👍

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/pgp/CanonicalizedSecretKey.java
```diff
@@ -253,17 +247,33 @@ public PGPSignatureGenerator getCertSignatureGenerator(Map<ByteBuffer, byte[]> s
         }
     }
 
-    public PGPAuthenticationSignatureGenerator getAuthenticationSignatureGenerator(int hashAlgorithm,
-                                                                                   Map<ByteBuffer, byte[]> signedHashes)
+
+    private PGPContentSignerBuilder getAuthenticationContentSignerBuilder(int hashAlgorithm, Map<ByteBuffer,
+            byte[]> signedHashes) {
+        if (getAlgorithm() == PublicKeyAlgorithmTags.EDDSA) {
```
hagau: This should probably also test for PRIVATE_KEY_STATE_DIVERT_TO_CARD in case of a key on a token, but I don't know how the Gnuk (or any other EdDSA-capable token) handle input data

libkeychain/src/main/java/org/bouncycastle/openpgp/AuthenticationSignatureGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+            sigValues = new MPInteger[1];
+            sigValues[0] = new MPInteger(new BigInteger(1, contentSigner.getSignature()));
+        } else if (contentSigner.getKeyAlgorithm() == PublicKeyAlgorithmTags.EDDSA) {
+            byte[] sig = contentSigner.getSignature();
+
+            sigValues = new MPInteger[2];
+
+            sigValues[0] = new MPInteger(BigIntegers.fromUnsignedByteArray(sig, 0, 32));
+            sigValues[1] = new MPInteger(BigIntegers.fromUnsignedByteArray(sig, 32, 32));
+        } else {
```
Valodim: add `if` and additional else to UnsupportedOperationException or IllegalStateException

libkeychain/src/main/java/org/bouncycastle/openpgp/AuthenticationSignatureGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+            sigValues = PGPUtil.dsaSigToMpi(contentSigner.getSignature());
+        }
+
+        return new PGPSignature(new SignaturePacket(sigType, contentSigner.getKeyID(), contentSigner.getKeyAlgorithm(),
```
Valodim: just directly encode to a `byte[]` and return that here

Valodim:
Did this fix DSA as expected?

hagau:
Nope, DSA authentication still fails.
Right now, sshlib expects a ASN.1 encoded signature and picks it apart to generate a RFC4253 compatible signature. I am thinking about changing this, it should happen in the SshAuthenticationService (the AuthenticationOperation should remain service-agnostic) and the service should return the SSH-compatible signature blob directly.

I would suggest merging this anyway as all further changes will be in SshAuthenticationService.

PR #2224, opened 2017-11-26
Encode signature to SSH compatible format in SshAuthenticationService

hagau:
## Description
Instead of returning the signature in native encoding, encode it according to RFC4253, RFC5656 & draft-ietf-curdle-ssh-ed25519

## How Has This Been Tested?
Tested with RSA, ECDSA & EdDSA. Testsuite finishes successfully.

## Types of changes


hagau:
DSA still fails.
- OpenSSH only supports DSA-(1024,160) (with SHA1...)
- OpenKeychain only supports DSA with a modulus > 1024bit
This PR implements (a fixed rough) signature encoding for DSA-(2048,256). I'm unsure whether supporting DSA at all makes any sense in this context.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/signature/SshSignature.java
```diff
@@ -0,0 +1,148 @@
...
+import org.bouncycastle.util.BigIntegers;
+import org.sufficientlysecure.keychain.pgp.UncachedPublicKey;
+import org.sufficientlysecure.keychain.ssh.key.SshEncodedData;
+import org.sufficientlysecure.keychain.ssh.utils.SshUtils;
+
+import java.io.IOException;
+import java.math.BigInteger;
+import java.security.NoSuchAlgorithmException;
+
+public class SshSignature {
```
Valodim: this whole class is only once instantiated and a single getter called on it. this would be more nicely done with a `SshSignatureConverter` that has a single public (optionally static) method `convertToSshSignature`. `SshSignature` sounds like it would be used for its data somewhere

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/signature/SshSignature.java
```diff
@@ -0,0 +1,148 @@
...
+        mPublicKey = publicKey;
+    }
+
+    private String getSignatureType() throws NoSuchAlgorithmException {
+        switch (mAlgorithm) {
+            case PublicKeyAlgorithmTags.RSA_SIGN:
+            case PublicKeyAlgorithmTags.RSA_GENERAL:
+                return "ssh-rsa";
+            case PublicKeyAlgorithmTags.ECDSA:
+                return "ecdsa-sha2-"+ SshUtils.getCurveName(mPublicKey);
```
Valodim: Why pass the whole UncachedPublicKey here, *and* the algorithm? Either just pass the public key, or pass the algorithm + curve oid. I'd say the latter is nicer

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/signature/SshSignature.java
```diff
@@ -0,0 +1,148 @@
...
+        SshEncodedData rsBlob = new SshEncodedData();
+        rsBlob.putMPInt(r);
+        rsBlob.putMPInt(s);
+
+        return getFinalSignature(rsBlob.getBytes());
+    }
+
+    private BigInteger getR(byte[] rawSignature) throws IOException {
+        ASN1Sequence asn1Sequence = getASN1Sequence(rawSignature);
+        return ASN1Integer.getInstance(asn1Sequence.getObjectAt(0)).getValue();
```
Valodim: do we ever expect those to fail? It's kind of weird for a "convert this to that" method to throw an `IOException`, if this is an "should never happen" deal you can just throw an IllegalStateException

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/SshAuthenticationService.java
```diff
@@ -141,12 +147,22 @@ private Intent authenticate(Intent data) {
...
         long authSubKeyId;
+        int authSubKeyAlgorithm;
+        CanonicalizedPublicKey authSubPublicKey = null;
         try {
             // get first usable subkey capable of authentication
             authSubKeyId = cachedPublicKeyRing.getSecretAuthenticationId();
+            // needed for encoding the resulting signature
+            authSubKeyAlgorithm = getPublicKey(masterKeyId).getAlgorithm();
+            if (authSubKeyAlgorithm == PublicKeyAlgorithmTags.ECDSA) {
+                authSubPublicKey = getPublicKey(masterKeyId);
```
Valodim: using implicit knowledge here that `SshSignature` only requires the key object for ECDSA couples these classes tighter than necessary. this should go away if you juts pass algorithm + curveOid

hagau: Yes, thanks. SshSignature was rather badly designed. It's now SshSignatureConverter with a static methods for converting ECDSA signatures and a static method for the rest.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/utils/SshUtils.java
```diff
@@ -0,0 +1,57 @@
...
+                return     "1.3.132.0.16";
+            case "1.3.132.0.36":
+                return     "1.3.132.0.36";
+            case "1.3.132.0.37":
+                return     "1.3.132.0.37";
+            case "1.3.132.0.38":
+                return     "1.3.132.0.38";
+
+            default:
+                return null;
```
Valodim: this will invariably lead to a NullPointerException, won't it?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/utils/SshUtils.java
```diff
@@ -0,0 +1,57 @@
...
+            case "1.3.132.0.1":
+                return     "1.3.132.0.1";
+            case "1.2.840.10045.3.1.1":
+                return  "1.2.840.10045.3.1.1";
+            case "1.3.132.0.33":
+                return     "1.3.132.0.33";
+            case "1.3.132.0.26":
+                return     "1.3.132.0.26";
+            case "1.3.132.0.27":
+                return     "1.3.132.0.27";
```
Valodim: whitespace

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ssh/signature/SshSignatureConverter.java
```diff
@@ -0,0 +1,137 @@
...
+import org.bouncycastle.bcpg.PublicKeyAlgorithmTags;
+import org.bouncycastle.util.BigIntegers;
+import org.sufficientlysecure.keychain.ssh.key.SshEncodedData;
+import org.sufficientlysecure.keychain.ssh.utils.SshUtils;
+
+import java.io.IOException;
+import java.math.BigInteger;
+import java.security.NoSuchAlgorithmException;
+
+public class SshSignatureConverter {
```
Valodim: care to add a bunch of tests to this class? just pin it down with a couple of input/output pairs

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/ssh/signature/SshSignatureConverterTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.sufficientlysecure.keychain.KeychainTestRunner;
+
+
+@RunWith(KeychainTestRunner.class)
+public class SshSignatureConverterTest {
+
+    private final static String sCurveOidNistP256 = "1.2.840.10045.3.1.7";
+
+    private final static int[] sRawEcDsaSignatureInt = {
+            0x30, 0x46,
```
Valodim: these can be nicely represented as hex strings with `Hex.decode()`, which also saves the `int[]` hack

hagau: Ah yes thanks, in hindsight I should have known there would be such a utility function used somewhere in OpenKeychain

PR #2227, opened 2017-11-28
Use different approach for whitelisting usb security tokens

Valodim:
The previous approach to usb whitelisting (added in 0920d9757228930ddc44ae802b6a18fb0f4e96a1, removed again in 1f7c7f49d481efd8e59e08b3d23023e2fc92cf75) didn't actually work, since communication errors happened before the check was ever called. This PR adds a similar check, but much earlier in the procedure and in the UsbTransport only.

## How Has This Been Tested?
Not at all, so far :)

## Types of changes


dschuermann:
👍 Manually tested with devices.

PR #2228, opened 2017-11-28
Do not re-enable contacts sync if first time wizard is not finished

### LINKED ISSUE #1996: Contact synchronization notification
My problem is related to issue #1720 

I am still experiencing these notifications "Access to contacts required. Touch to configure linking to contacts",
they keep coming, as I did not enable the "Link keys to contacts" feature.

    Nexus 5 (Android 6.0.1)
    OpenKeychain 4.2 (installed via F-Droid)

How can I help to pinpoint the problem?

dschuermann:
Hopefully fixes #1996

## Description
If the first time wizard is not finished or properly finished, the contacts sync notification will pop-up again and again because the service keeps re-enabling on application start.

## Motivation and Context
#1996

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
@@ -119,7 +119,10 @@ public final class Constants {
         public static final String LANGUAGE = "language";
         public static final String KEY_SERVERS = "keyServers";
         public static final String PREF_VERSION = "keyServersDefaultVersion";
-        public static final String FIRST_TIME = "firstTime";
+        // false if first time wizard has been finished
+        public static final String FIRST_TIME_WIZARD = "firstTime";
+        // false if app has been started at least once (also from background etc)
+        public static final String FIRST_TIME_APP = "firstTimeApp";
         public static final String CACHED_CONSOLIDATE = "cachedConsolidate";
         public static final String SEARCH_KEYSERVER = "search_keyserver_pref";
         public static final String SEARCH_KEYBASE = "search_keybase_pref";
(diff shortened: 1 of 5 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
```diff
@@ -99,18 +98,21 @@ public void onCreate() {
         // Add OpenKeychain account to Android to link contacts with keys and keyserver sync
         createAccountIfNecessary(this);
 
-        if (Preferences.getKeyserverSyncEnabled(this)) {
-            // will update a keyserver sync if the interval has changed
+        if (Preferences.getPreferences(this).isAppExecutedFirstTime()) {
```
Valodim: preferences object belongs into variable

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/KeychainApplication.java
```diff
@@ -99,18 +98,21 @@ public void onCreate() {
...
+            ContactSyncAdapterService.enableContactsSync(this);
+
+            Preferences.getPreferences(this).setPrefVersionToCurrentVersion();
         }
 
-        // if first time, enable keyserver and contact sync
-        if (Preferences.getPreferences(this).isFirstTime()) {
+        if (Preferences.getKeyserverSyncEnabled(this)) {
+            // will update a keyserver sync if the interval has changed
             KeyserverSyncAdapterService.enableKeyserverSync(this);
```
Valodim: this will trigger twice now on first time. maybe `} else if (...) {` it?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
```diff
@@ -119,7 +119,8 @@
         public static final String LANGUAGE = "language";
         public static final String KEY_SERVERS = "keyServers";
         public static final String PREF_VERSION = "keyServersDefaultVersion";
-        public static final String FIRST_TIME = "firstTime";
+        public static final String FIRST_TIME_WIZARD = "firstTime";
```
Valodim: the difference is not obvious, please add a comment

PR #2229, opened 2017-11-30
Show notification for Clipboard data

Valodim:
## Motivation and Context

Play Store reviews and conversations with users suggest that decrypting data and importing keys from the clipboard is a very common use case. The Android clipboard has some [security problems](https://plus.google.com/+BitcoinWalletApp/posts/dRAPrwhgLgF), which is why we didn't like it much as a workflow so far. However, people do use this regardless, so we should at least try and do our best to work with that.

## Description
This PR adds a listener that monitors clipboard data. If OpenPGP data (encrypted message or keys) are found, a notification is displayed that allows the user to immediately list/import the keys or decrypt the message. 

The haptics of the notification involved a bunch of experimental fiddling around with, and it feels good to me now. I actually quite like the workflow. Still, second opinions welcome.

Some more details:
* clipboard data from OpenKeychain itself is ignored, unless it's a debug build
* the monitor service is bound to screen on / screen off broadcast intents
* notifications vibrate a little bit, which is required to show the notification as heads-up (rather than just in the background)
* after 7 seconds, the notification will be cleared. the assumption is that users will either click them immediately, or don't intend to. this avoids lingering notifications
* the content from the clipboard is immediately copied into a cache file to make sure the stuff we see first is the stuff that we actually import/decrypt
* the decrypt activity has its own task affinity now. This changes task stack behavior in a way that pressing "back" in the DecryptActivity willl not bring the user to the key list, but back to the task that launched the decryption.

## How Has This Been Tested?
Manually :)

## Screenshots
image

image

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -277,21 +277,15 @@
             android:name=".ui.DisplayTextActivity"
             android:configChanges="orientation|screenSize|keyboardHidden|keyboard"
             android:label="@string/title_decrypt"
-            android:parentActivityName=".ui.MainActivity"
-            android:windowSoftInputMode="stateHidden">
-            <meta-data
-                android:name="android.support.PARENT_ACTIVITY"
-                android:value=".ui.MainActivity" />
+            android:windowSoftInputMode="stateHidden"
+            android:taskAffinity="org.sufficientlysecure.keychain.decrypt">
         </activity>
         <activity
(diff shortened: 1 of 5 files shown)


Valodim:
I *love* the workflow here. Unfortunately, monitoring the clipboard has become very unstable, there is no way to keep the background service running without annoying the user in one way or another. This issue also became more prominent with every Android version, especially in Android O. All in all we don't think this can work reliably enough that we would want to introduce and maintain it.

PR #2243, opened 2017-12-15
Remove experimental word confirm

dschuermann:
Removes experimental word confirm feature

## Motivation and Context
Not really used outside of OpenKeychain.


Valodim:
yup.

Issue #2179, opened 2017-10-03
Using Yubikey Neo (brand new) Can sign but not decrypt (apdu error)


Morthawt:
I go to decrypt a test message and it fails with "Failed to chain apdu"

## Expected Behavior
The message should be decrypted.

## Current Behavior
Error: Failed to chain apdu

Take away the Security Token now and touch TRY AGAIN.

## Possible Solution
I do not know

## Steps to Reproduce (for bugs)
1. Click my key
2. Encrypt a message (not signed) to myself
3. Go to main screen
4. Encrypt/Decrypt
5. Read from clipboard
6. Hold Yubikey Neo to the back of the phone when requested.

## Context
I am unable to use my key for decrypting messages. All I seem to be able to do is sign.

## Your Environment

* Android Version: 6.0.1
* OpenKeychain Version: 4.7
* From Google Play or F-Droid?: F-Droid

Valodim:
Huh. This happens [https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenHelper.java#L798](here), when a command returns an error before it is fully transferred.

ping @nmikhailov, also tentative ping @jas4711 do you know if maybe something changed on new yubikeys lately?

Morthawt:
I am not aware of any changes. What is strange is that the Yubico
authenticator app which obtains TOTP codes via NFC works, signing
information through your app using NFC works. Decryption is the thing which
is not working. I am unaware of what the problem is. I even tried
generating the key internally instead of making a key myself as I did
before and it just errors. You have the best GPG app for android, Yubikeys
are (hands down) the best hardware for personal security and privacy. It
would be nice if perhaps they could help you resolve this or supply you one
for testing? But all I know is, this problem is my only stumbling block so
far with your amazing app.




----------------------------------------
For secure PGP/GnuPG users, if it is *seriously *needed/personal etc
(I rarely use crypto due to more work involved, no ability to search past
emails etc)
The download link for my public key
<https://keyserver2.pgp.com/vkd/DownloadKey.event?keyid=0xDF945F5C614D98E6>
Key Fingerprint: 4C23 C140 E8BE E33B 7858  D986 DF94 5F5C 614D 98E6
----------------------------------------

Valodim:
[...]

PR #2252, opened 2018-01-12
Refactor SecurityToken ops

Valodim:
While investigating #2179 and #2049, I looked at how our decryption code worked in SecurityTokenConnection and noticed there were a couple of points I didn't understand and that weren't documented - so I wrote unit tests for them, and refactored until I knew what was going on :)

This PR now does a few things:
- extract particular use cases from SecurityTokenConnection into their own UseCase classes
- add unit tests for decryption on token
- refactor the decryption use case a bit for readability (this is kind of a WIP, not sure if I want to spend more time)
- refactor SecurityTokenConnection, in particular split up and reorder some methods

I also added a test case against that compares what we currently emit against ee8cd3862f65de580ed949bc838628610e22cd98. This is not currently identical, but could be made so if we find out it's more compatible (see #2049). But I read the openpgp card implementation code, and can't figure out any reason why this wouldn't work, so it's a bit of poking in the dark.

All of these are true refactorings - there should be no semantic changes in here.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/GenerateKeyUseCase.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/GenerateKeyUseCase.java
@@ -1,0 +1,50 @@
+package org.sufficientlysecure.keychain.securitytoken;
+
+
+import java.io.IOException;
+
+import org.sufficientlysecure.keychain.util.Passphrase;
+
+
+public class GenerateKeyUseCase {
+    private final SecurityTokenConnection connection;
+
+    public static GenerateKeyUseCase create(SecurityTokenConnection connection) {
+        return new GenerateKeyUseCase(connection);
+    }
+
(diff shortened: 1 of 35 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/GenerateKeyUseCase.java
```diff
@@ -0,0 +1,50 @@
+package org.sufficientlysecure.keychain.securitytoken;
+
+
+import java.io.IOException;
+
+import org.sufficientlysecure.keychain.util.Passphrase;
+
+
+public class GenerateKeyUseCase {
```
nmikhailov: Looks like this is not used anywhere

Valodim: It's not, never was iirc. I guess we left it in just to keep the code close :)

dschuermann:
Please add license headers

Valodim:
done :+1:

PR #2253, opened 2018-01-14
Autocrypt 1.0

Valodim:
This PR changes the logic from behavior Autocrypt-circa-July to the shiny new Autocrypt 1.0. These changes should be compatible with current K-9, but I havn't done a lot of testing yet.

Still missing is a way to deal with "external" keys. I don't have a good plan for that yet. It might make sense to save a state similar to having received an autocrypt header when we see a signature by some key from an email address (careful with mailing lists though). There are a lot of variations to this concept, and it's actually not terribly important for now to get this 100% right. It does need to handle deduplication of keys though, that's broken right now and needs to be fixed.


Valodim:
Ok this took a lot longer than expected to polish a bit because I ran into performance issues. But this works for the external provider now, and I made sure it's compatible with stable K-9. Unfortunately, it currently breaks our own activities, heh. But we're getting there-

PR #2254, opened 2018-01-19
Make predicate names in KeychainProvider consistent and allow export of SSH public keys without associated private key in keyring

hagau:
## Description
Make predicate names in KeychainProvider consistent and allow export of SSH public keys from keyrings with authentication subkey but without associated private key.

## Motivation and Context
This allows for use cases where the user needs to copy the SSH public key blob of someone else (or of a keyring with secret key material on another device) manually into an app not using the sshauthentication-api (of which the implementation in OpenKeychain also supports export of SSH public keys for keyrings without private key with this PR) or a browser.

## How Has This Been Tested?
Works for me.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
@@ -353,6 +353,8 @@ public class KeychainProvider extends ContentProvider {
                         "kE." + Keys.KEY_ID + " AS " + KeyRings.HAS_ENCRYPT);
                 projectionMap.put(KeyRings.HAS_SIGN_SECRET,
                         "kS." + Keys.KEY_ID + " AS " + KeyRings.HAS_SIGN_SECRET);
+                projectionMap.put(KeyRings.HAS_AUTHENTICATE,
+                        "kA." + Keys.KEY_ID + " AS " + KeyRings.HAS_AUTHENTICATE);
                 projectionMap.put(KeyRings.HAS_AUTHENTICATE_SECRET,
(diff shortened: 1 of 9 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/provider/KeychainProvider.java
```diff
@@ -420,6 +422,16 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
                                     + " AND ( kA." + Keys.EXPIRY + " IS NULL OR kA." + Keys.EXPIRY
                                     + " >= " + new Date().getTime() / 1000 + " )"
                                     + ")" : "")
+                        + (plist.contains(KeyRings.HAS_AUTHENTICATE_PUBLIC) ?
```
Valodim: What makes this "public"? It will include secret keys as well, won't it? Is this distinction really necessary?

hagau: Yes, I'm not too sure about the naming. I chose to keep to keep HAS_AUTHENTICATE as having secret keys available to keep in line with HAS_SIGN etc. An earlier version changed HAS_AUTHENTICATE to be inclusive and added HAS_AUTHENTICATE_SECRET, would that be more appropriate?

Valodim: Ugh, this whole thing is messier than I had in mind. I don't have a good answer, the `HAS_` things are already inconsistent as they are.

hagau: No worries. I was thinking about making it all consistent by specialising into `HAS_*` / `HAS_*_SECRET` since I couldn't find a good reason for the naming as it is; I'll look into that later.

PR #2256, opened 2018-01-21
Gradle 4.4.1 + build sdk 27

Valodim:
Update to gradle 4.4.1, and build sdk 27. This greatly improves build times for me, but might introduce non-obvious incompatibilities.

This PR includes a hack to retain access to PresharedKeyManager. I think this should be fine, since conscrypt will likely not drop the class while apps can build against sdk 25 and use it legitimately. We are also not doing any reflection hacks ourselves, just rely on the duck typing (indeed) in conscrypt to do the right thing for us :)

### DIFF
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -66,7 +66,7 @@ dependencies {
     // http://robolectric.org/getting-started/
     // http://www.vogella.com/tutorials/Robolectric/article.html
     testCompile 'junit:junit:4.12'
-    testCompile ('org.robolectric:robolectric:3.2.2') {
+    testCompile ('org.robolectric:robolectric:3.6.1') {
         exclude group: 'org.bouncycastle', module: 'bcprov-jdk16'
     }
     testCompile 'org.mockito:mockito-core:1.10.19'
--- a/OpenKeychain/build.gradle
+++ b/OpenKeychain/build.gradle
@@ -167,8 +167,8 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_7
-        targetCompatibility JavaVersion.VERSION_1_7
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 
     testOptions.unitTests.all {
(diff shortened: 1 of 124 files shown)


OpenKeychain/build.gradle
```diff
@@ -89,77 +88,67 @@ dependencies {
...
+        'com.fidesmo:nordpol-core:296e71b12884a9cd28cf00ab908973bbf776a90be1f23ac897380d91604e614d',
+        'android.arch.lifecycle:runtime:d0b36278878c82b838acc4308595bec61a3b5f6e7f2acc34172d7e071b2cf26d',
+        'android.arch.lifecycle:common:ff0215b54e7cbaaa898f8fd00e265ed6ea198859e10604bc1c5e78477df48b5c',
+        'android.arch.core:common:5192934cd73df32e2c15722ed7fc488dde90baaec9ae030010dd1a80fb4e74e1',
     ]
 }
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
-    buildToolsVersion rootProject.ext.buildToolsVersion
```
dschuermann: Can you explain this and related changes? Is it no longer required to define build tools?

dschuermann: Would be great to have the same build tools in all projects. Contributors then only need to install this one version.

build.gradle
```diff
@@ -3,14 +3,15 @@ apply plugin: 'com.github.ben-manes.versions'
...
     dependencies {
         // NOTE: Always use fixed version codes not dynamic ones, e.g. 0.7.3 instead of 0.7.+, see README for more information
-        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.android.tools.build:gradle:3.0.1'
         classpath files('gradle-witness.jar')
         // bintray dependency to satisfy dependency of openpgp-api lib
-        classpath 'com.novoda:bintray-release:0.2.7'
+        classpath 'com.novoda:bintray-release:0.8.0'
 
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.6.3'
```
dschuermann: I think we can also remove the two dependencies below this line

OpenKeychain/build.gradle
Valodim: the property is apparently deprecated, it defaults to the build tools version now. I'm not sure if still using it causes problems, but it does yield a bunch of warnings during compilation.

ah, I also forgot to update the submodules

Valodim:
Finally got everything working! After a ton of trial & error, everything looks in order now with new gradle, build sdk, robolectric, and travis.

dschuermann:
Great work!

PR #2264, opened 2018-02-07
Fix SCP11b (problem since merged pull request #2186)

af-anssi:
SCP11b secure messaging is not working at all since pull request #2186 is merged.

## Description

Remove zeroization of the unencrypted command data buffer in SCP11.

## Motivation and Context

Since the APDU command refactoring, the command data buffer is not systematically copied by the `create` method, so it can be an alias to (real/concrete) data holded by other class instances, so it must not be zeroized.

## How Has This Been Tested?

SCP11b is completely functional with this modification.

## Types of changes


Valodim:
Thanks for the fix. I already had this in a local branch, but didn't push them. We'll try to get this into a release asap :)

This fix works around an implementation detail of CommandApdu that shouldn't have been there in the first place. I fixed CommandApdu instead, see bc4ab57a0aa78efa4d28faf095b0515d582362a4.

PR #2265, opened 2018-02-07
Fix decrypt with ECDH

af-anssi:
Fix broken ECDH decryption.

## Description

During the refactoring/creation of the `decryptSessionKeyEcdh` method in the `PsoDecyptTokenOp class` (see commit 139735f0e1c3460cdb66567b1d9431eae0634f50), a `System.arrayCopy` as been changed to `Array.copyOfRange`, but the second parameter has not been adjusted leading to a 2-byte truncated array: in `System.arrayCopy` the second parameter is the length of copied data, while in `Array.copyOfRange` it is the `to` index (not included) so the offset of copied data must be added.

## Motivation and Context

Fixes ECDH decryption

## How Has This Been Tested?

ECDH decryption is working correctly within OpenKeychain (issue #2261 still needs to be fixed).

## Types of changes


Valodim:
I had this fixed already (8e31615f7e5c9151666b5a0d1106d5d23d4ab517), but originally didn't push the commits because I wanted to add unit tests for it. If you have a debug version all communication is logged, could you extract some test vectors for me for this feature (both working / non-working)?

af-anssi:
I didn't check other branches for pending commit, sorry for the redundancy. I will thus close this pull request.
I will try to find some time to generate some test vectors.

Valodim:
Awesome, thanks :+1:

Issue #2266, opened 2018-02-08
ECDH (encrpytion) subkey w/25519


J-J-Chiarella:
Are ECDH encryption subkeys with Curve 25519 usable? I know EdDSA support was added and that uses Curve25519 as well.

* Android Version: Marshmallow
* Device Model: BLU Life One X (2016)
* OpenKeychain Version: 4.8
* From Google Play or F-Droid?: Google Play
 
Key created via GnuPG 2.1.15 and exported unencrypted to test.

Error message is only "No valid keys."

Thank you for all the great work, of course.

Valodim:
I looked at this a couple of times, and just two days ago finally figured out how to actually get it to work :) I just need to get it into a mergable state, but it shouldn't be too long until we support this.

Valodim:
I have a working variant of this in a branch, but there is some amount of confusion about the spec. I asked for clarification on [gnupg-devel](https://lists.gnupg.org/pipermail/gnupg-devel/2018-February/033437.html), we'll include this feature as soon as that's cleared up.

PR #2269, opened 2018-02-14
Cv25519

Valodim:
This PR adds support for cv25519 encryption and decryption. Yay.

The way scalars are used is kind of weird here. This PR is built for interoperability with gnupg, but I feel like it doesn't really do what the spec says. I asked about this on [GnuPG-devel](https://lists.gnupg.org/pipermail/gnupg-devel/2018-February/033437.html), let's wait for that.

Related #2266 and #2178, also possibly interesting for @Keks

## How Has This Been Tested?
Unit tests included, encrypt/decrypt and decryption from gnupg

## Types of changes


Valodim:
Alright, I'm fairly sure at this point the issue I brought up on gnupg-devel is a spec and interoperability one, not about the implementation. Here we go~

PR #2275, opened 2018-02-26
Use Constants.DEBUG instead of BuildConfig.DEBUG

dschuermann:
Use Constants.DEBUG instead of BuildConfig.DEBUG to allow debugging of release builds

## Motivation and Context
Debugging a problem caused by ProGuard requires me building a release build but with logging enabled.

## Types of changes


Zainlezark:
#2275

PR #2283, opened 2018-03-05
Update bottomsheet to 3.1.3, fixes #2278

### LINKED ISSUE #2278: Can't save a file after decryption.
As of version 4.9.1 a file cannot be saved after the decryption process finishes.  The app crashes.  Wiping and reinstalling doesn't fix the problem.  However the issue isn't present after downgrading to version 4.8.1 on F-droid.

dschuermann:
Update bottomsheet to 3.1.3, fixes #2278

upstream: https://github.com/soarcn/BottomSheet/issues/132

## Types of changes


Valodim:
accidentally based on the same branch? also, this changes a lot of the verification hashes, was that intentional?

dschuermann:
done in https://github.com/open-keychain/open-keychain/commit/b086cd09e732ee439574c3a42e1ee24462b5d5e7 and https://github.com/open-keychain/open-keychain/commit/f0e0ed4a4838f55701c100ed62928698c5596d60

PR #2301, opened 2018-04-03
New identity chooser dialog for API

Valodim:
This adds a new key id selection dialog:

* dialog theme, better preserves visual context of calling app
* immediate key generation, no further questions asked
* key generation can be backed out of
* reports name and creation to the api caller
* implicit app registration and permission when a key is selected

## How Has This Been Tested?
I tested combinations of new and old K-9 and OpenKeychain. It's pretty much just UI stuff, so no unit tests included :(

## Screenshots:

screenshot_openkeychain_ debug _20180403-015408

## Types of changes


chriswoolfe:
Great work @Valodim ! This is much simpler than the old key create wizard. If I still wanted to put a passphrase on the private key, would I be able to from this dialog via a configuration? Otherwise, perhaps I could reuse the existing reset password dialog or set a passphrase onClickNoKeysGenerate() via builder.setSecurityTokenPin ?

PR #2309, opened 2018-04-25
Use correct signature format identifier when creating SSH signatures for RSA with SHA256 & SHA512

hagau:
[RFC8332](https://tools.ietf.org/html/rfc8332) defines new signature format identifier for RSA signatures with SHA256 & SHA512. Use those for correct SSH signature creation.

## How Has This Been Tested?
Works with ConnectBot.
Passes test.

## Types of changes


hagau:
- [ ] Add tests for SHA256 & SHA512

Valodim:
lgtm

Issue #2270, opened 2018-02-14
Feature: Support for WKS/WKD key discovery


wiktor-k:
[Web Key Directory is a draft](https://datatracker.ietf.org/doc/draft-koch-openpgp-webkey-service/) for a new key discovery scheme (implemented e.g. in new gnupg). It is similar to HKP but instead of querying the keyserver it queries the domain from the UID's e-mail address. The difference is that WKS returns binary data and HKP ASCII-armored.

For example, when looking for key for `Joe.Doe@Example.ORG` the key would be fetched from `https://example.org/.well-known/openpgpkey/hu/iy9q119eutrkn8s1mk4r39qejnbu3n5q`.

Open Keychain could support this in several ways:
  * register as a handler for URLs containing `.well-known/openpgpkey/hu`,
  * when searching e-mails additionally query WKS,

Several services (like kernel.org) already have WKS support, for example Linus' key is at: https://kernel.org/.well-known/openpgpkey/hu/pf113mfnx1f3eb1yiwhsipa91xfc7o4x. Some third party services like posteo.de [support WKS too](https://posteo.de/.well-known/openpgpkey/). WKS is also [supported by Enigmail](https://www.gnupg.org/blog/20170803-web-key-in-enigmail.html).

More details here (in German): https://www.golem.de/news/web-key-service-openpgp-schluessel-ueber-https-verteilen-1609-123194.html

If this is in scope of Open Keychain I could work on a PR.

Valodim:
WKD is definitely in scope for OpenKeychain, and looking up wkd entries when the user searches for an e-mail address makes sense. It's not particularly high on my own list of priorities, but if you make a nice PR we would probably merge it :+1:

That said, I'm not fully convinced WKD is the way to go for most encrypted mail use cases, and I have yet to see any actual user workflow considerations based on it.

wiktor-k:
Thanks @Valodim for 👍.

I know there is certain [conflict of interests](https://autocrypt.org/) 😉 but I'm mostly interested in WKD as a quite simple *key* discovery mechanism, I didn't think of encrypted mail in particular.

I'll see what can I reuse from existing KeyserverClients...

Valodim:
Don't worry about that. We know there are use cases that Autocrypt doesn't cover, and even then testing multiple approaches to see what sticks is a good thing! ;)

bam80:
>If this is in scope of Open Keychain I could work on a PR.

[...]

PR #2314, opened 2018-05-08
Search for keys in Web Key Directory

wiktor-k:
## Description
If a search pattern that looks like an email address is found
an additional query to Web Key Directory will be performed.

I'd gladly appreciate any feedback!

## Motivation and Context

Fixes partially #2270.

## How Has This Been Tested?

Search for `test-wkd@metacode.biz`, a key that is reachable only via WKD succeeds with this change.

screenshot_1525788892

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/CloudSearch.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/CloudSearch.java
@@ -51,6 +51,9 @@ public class CloudSearch {
         if (cloudPrefs.searchFacebook) {
             servers.add(FacebookKeyserverClient.getInstance());
         }
+        if (cloudPrefs.searchWebKeyDirectory) {
+            servers.add(WebKeyDirectoryClient.getInstance());
+        }
 
         int numberOfServers = servers.size();
         final ImportKeysList results = new ImportKeysList(numberOfServers);
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright (C) 2017 Schürmann & Breitmoser GbR
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.sufficientlysecure.keychain.keyimport;
+
+
+import android.support.annotation.Nullable;
+
+import org.sufficientlysecure.keychain.network.OkHttpClientFactory;
+import org.sufficientlysecure.keychain.pgp.UncachedKeyRing;
+import org.sufficientlysecure.keychain.pgp.exception.PgpGeneralException;
+import org.sufficientlysecure.keychain.util.ParcelableProxy;
+import org.sufficientlysecure.keychain.util.ZBase32;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.Proxy;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import timber.log.Timber;
+
+
+/**
+ * Searches for keys using Web Key Directory protocol.
+ *
+ * @see <a href="https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-05#section-3.1">Key Discovery</a>
+ */
+public class WebKeyDirectoryClient implements KeyserverClient {
+
+    public static WebKeyDirectoryClient getInstance() {
+        return new WebKeyDirectoryClient();
+    }
+
+    private WebKeyDirectoryClient() {
+    }
+
+    private static final Pattern EMAIL_PATTERN = Pattern.compile("^\\s*(.+)@(.+)\\s*$");
+
+    @Override
+    public List<ImportKeysListEntry> search(String name, ParcelableProxy proxy)
+            throws QueryFailedException {
+        URL webKeyDirectoryURL = toWebKeyDirectoryURL(name);
+
+        if (webKeyDirectoryURL == null) {
+            Timber.d("Name not supported by Web Key Directory Client: " + name);
+            return Collections.emptyList();
+        }
+
+        Timber.d("Web Key Directory import: " + name + " using Proxy: " + proxy.getProxy());
+
+        byte[] data = query(webKeyDirectoryURL, proxy.getProxy());
+
+        if (data == null) {
+            Timber.d("No Web Key Directory endpoint for: " + name);
+            return Collections.emptyList();
+        }
+
+        // if we're here that means key retrieval succeeded,
+        // would have thrown an exception otherwise
+        try {
+            UncachedKeyRing ring = UncachedKeyRing.decodeFromData(data);
+            return Collections.singletonList(new ImportKeysListEntry(null, ring));
+        } catch (PgpGeneralException | IOException e) {
+            Timber.e(e, "Failed parsing key from Web Key Directory during search");
+            throw new QueryFailedException("No valid key found on Web Key Directory");
+        }
+    }
+
+    @Override
+    public String get(String name, ParcelableProxy proxy) {
+        throw new UnsupportedOperationException("Returning armored key from Web Key Directory not supported");
+    }
+
+    @Nullable
+    private byte[] query(URL url, Proxy proxy) throws QueryFailedException {
+        try {
+            Timber.d("fetching from Web Key Directory with: %s proxy: %s", url, proxy);
+
+            Request request = new Request.Builder().url(url).build();
+
+            OkHttpClient client = OkHttpClientFactory.getClientPinnedIfAvailable(url, proxy);
+            Response response = client.newCall(request).execute();
+
+            if (response.isSuccessful()) {
+                return response.body().bytes();
+            } else if (response.code() == HttpURLConnection.HTTP_NOT_FOUND) {
+                return null;
+            } else {
+                throw new QueryFailedException("Error while fetching key from Web Key Directory. " +
+                        "Response:" + response);
+            }
+
+        } catch (UnknownHostException e) {
+            Timber.e(e, "Unknown host at Web Key Directory key download");
+            return null;
+        } catch (IOException e) {
+            Timber.e(e, "IOException at Web Key Directory key download");
+            throw new QueryFailedException("Cannot connect to Web Key Directory. "
+                    + "Check your Internet connection!"
+                    + (proxy == Proxy.NO_PROXY ? "" : " Using proxy " + proxy));
+        }
+    }
+
+    @Override
+    public void add(String armoredKey, ParcelableProxy proxy) {
+        throw new UnsupportedOperationException("Uploading keys to Web Key Directory is not supported");
+    }
+
+    @Nullable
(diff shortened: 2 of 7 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/CloudSearch.java
```diff
@@ -52,6 +52,8 @@
             servers.add(FacebookKeyserverClient.getInstance());
         }
 
+        servers.add(WebKeyDirectoryClient.getInstance());
```
wiktor-k: This looks like it should also be gated with `cloudPrefs.searchXXX` right?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/processing/ImportKeysListCloudLoader.java
```diff
@@ -83,8 +83,10 @@ public ImportKeysListCloudLoader(Context context, CloudLoaderState loaderState,
 
         // Now we have all the data needed to build the parcelable key ring for this key
         for (ImportKeysListEntry e : mEntryList) {
-            e.setParcelableKeyRing(ParcelableKeyRing.createFromReference(e.getFingerprint(), e.getKeyIdHex(),
-                    e.getKeybaseName(), e.getFbUsername()));
+            if (e.getParcelableKeyRing() == null) {
```
wiktor-k: Is this a good idea to introduce this here? I'm returning a key ring from search (search already includes entire key) and don't want to have it overridden here... If there is a better idea please tell.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import timber.log.Timber;
+
+
+public class WebKeyDirectoryClient implements KeyserverClient {
```
wiktor-k: This is adapted from FacebookKeyserverClient.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,203 @@
...
+
+    public static WebKeyDirectoryClient getInstance() {
+        return new WebKeyDirectoryClient();
+    }
+
+    private WebKeyDirectoryClient() { }
+
+    private static Pattern EMAIL_PATTERN = Pattern.compile("^\\s*(.+)@(.+)\\s*$");
```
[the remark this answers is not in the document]
MauroMombelli: regex cannot validate 100% an email address.

see here for a correct implementatio (apache commons) https://github.com/apache/commons-validator/blob/trunk/src/main/java/org/apache/commons/validator/EmailValidator.java

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,203 @@
...
+        List<ImportKeysListEntry> entry = new ArrayList<>(1);
+
+        byte[] data = query(name, proxy);
+        // if we're here that means key retrieval succeeded,
+        // would have thrown an exception otherwise
+        try {
+            try {
+                entry.add(getEntry(name, data));
+            } catch (UnsupportedOperationException e) {
+                Timber.e("Parsing retrieved Web Key Directory key failed!");
```
MauroMombelli: why here you still want to return an empty entry array?

MauroMombelli: why here you still want to return an empty entry array?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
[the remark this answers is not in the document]
wiktor-k: @MauroMombelli thanks, unfortunately it doesn't seem to have extraction of local/domain part and I need that.
My point is if it's really necessary to actually validate e-mail here, the e-mail may be valid but using some obscure things like comments etc. that I would need to strip out. So it could be a lot of code (more code -> more bugs) for a (probably) marginal benefit.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,203 @@
...
+                    .url(url)
+                    .build();
+
+            OkHttpClient client = OkHttpClientFactory.getClientPinnedIfAvailable(url, proxy.getProxy());
+            Response response = client.newCall(request).execute();
+
+            if (response.isSuccessful()) {
+                return response.body().bytes();
+            } else {
+                // probably a 404 indicating that the key does not exist
```
MauroMombelli: maybe check for a 5xx type error and return a different kind of message (server overloaded)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final char[] ALPHABET = "ybndrfg8ejkmcpqxot1uwisza345h769".toCharArray();
+    private static final int SHIFT = Integer.numberOfTrailingZeros(ALPHABET.length);
+    private static final int MASK = ALPHABET.length - 1;
+
+    private static String encodeZBase32(byte[] data) {
+        if (data.length == 0) {
+            return "";
+        }
+
+        StringBuilder result = new StringBuilder();
```
MauroMombelli: Maybe you can safetly initialize the size data.length() + SHIFT to avoid realloc?

wiktor-k: Honestly I'd keep it as it is. The code for ZBase32 is already quite opaque and this function is not inside any tight loop (it's called only once per search) so that the no-reallocating buffer version wouldn't show any observable benefits.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/CloudSearch.java
wiktor-k: I've modified the source so in case a non-supported name is searched for WebKeyDirectoryClient just returns an empty list so no `cloudPrefs.searchWKD` is needed... I think.

Valodim: I think a setting would still be nice, if only for discoverability of the fetaure.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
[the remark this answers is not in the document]
Valodim: I think it's fine. Worse case it'll die in dns resolution or request a nonexisting hash :)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/processing/ImportKeysListCloudLoader.java
Valodim: This stuff was never particularly pretty anyways. So yeah it's fine

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+        catch (NoSuchAlgorithmException e) {
+            throw new AssertionError("SHA-1 should always be available");
+        }
+    }
+
+    private static final char[] ALPHABET = "ybndrfg8ejkmcpqxot1uwisza345h769".toCharArray();
+    private static final int SHIFT = Integer.numberOfTrailingZeros(ALPHABET.length);
+    private static final int MASK = ALPHABET.length - 1;
+
+    private static String encodeZBase32(byte[] data) {
```
Valodim: Please move this into a `util.ZBase32` class, it doesn't belong here. A link to the spec would also be nice

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+            throws QueryFailedException, QueryNeedsRepairException {
+        URL webKeyDirectoryURL = toWebKeyDirectoryURL(name);
+
+        if (webKeyDirectoryURL == null){
+            Timber.d("Name not supported by Web Key Directory Client: " + name);
+            return Collections.emptyList();
+        }
+
+        Timber.d("Web Key Directory import: " + name + " using Proxy: " + proxy.getProxy());
+        List<ImportKeysListEntry> entry = new ArrayList<>(1);
```
Valodim: `Collections.singletonList`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public String get(String name, ParcelableProxy proxy) throws QueryFailedException {
+        throw new UnsupportedOperationException("Returning armored key from Web Key Directory not supported");
+    }
+
+    private byte[] query(URL url, ParcelableProxy proxy) throws QueryFailedException {
+        try {
+            Timber.d("fetching from Web Key Directory with: " + url + " proxy: " + proxy.getProxy());
```
Valodim: Use interpolation instead of string concatenation:

```
Timber.d("fetching from Web Key Directory with: %s proxy: %s, url, proxy.getProxy());
```

(applies to some other places as well)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+    private WebKeyDirectoryClient() { }
+
+    private static Pattern EMAIL_PATTERN = Pattern.compile("^\\s*(.+)@(.+)\\s*$");
+
+    @Override
+    public List<ImportKeysListEntry> search(String name, ParcelableProxy proxy)
+            throws QueryFailedException, QueryNeedsRepairException {
+        URL webKeyDirectoryURL = toWebKeyDirectoryURL(name);
+
+        if (webKeyDirectoryURL == null){
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Web Key Directory returns the entire key even during our searching phase.
+     *
+     * @throws UnsupportedOperationException if the key could not be parsed
+     */
+    @NonNull
+    private static ImportKeysListEntry getEntry(byte[] data)
+            throws UnsupportedOperationException, IOException, PgpGeneralException {
+        ImportKeysListEntry entry = new ImportKeysListEntry();
```
Valodim: why not construct from the uncached key ring entry? I'm not sure if it does everything you need, but a `ImportKeysListEntry.from(UncachedKeyRing)` method seems more appropriate here

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/WebKeyDirectoryClient.java
```diff
@@ -0,0 +1,220 @@
...
+
+        try {
+            entry.setBitStrength(key.getBitStrength());
+            final int algorithm = key.getAlgorithm();
+            entry.setAlgorithm(KeyFormattingUtils.getAlgorithmInfo(algorithm, key.getBitStrength(),
+                    key.getCurveOid()));
+        } catch (NumberFormatException | NullPointerException e) {
+            Timber.e(e, "Conversion for bit size, algorithm, or creation date failed.");
+            // can't use this key
+            throw new UnsupportedOperationException(
```
Valodim: `UnsupportedOperationException` is not the right exception to throw if something went wrong during parsing of the key. If it's a thing that shouldn't occur (e.g. you expect the to always be correct) it's an `AsseritonError`, if it's just malformed input data it should be some kind of `IOException` or perhaps `PgpGeneralException`

chriscow1:
Yea

Valodim:
Just tested and this works perfectly, thanks for your contribution! :+1:

PR #2318, opened 2018-05-10
Extend QR scanner to support VCards with KEY field

wiktor-k:
## Description

This change will extend QR scanner with simple VCard support. When a VCard is scanned it will extract URI from `KEY` field and process it just like it would be a URI scanned directly. When a `openpgp4fpr` URI is put there the UI would search and import the key and show the confirmdialog.

Example VCard with this URI:

    BEGIN:VCARD
    FN:Test WKD
    EMAIL:test-wkd@metacode.biz
    KEY:OPENPGP4FPR:74EC8D3DA82A79DAA25DF10C6BA55ED83ABAE1BB
    END:VCARD

## Motivation and Context

Currently QR scanner supports only `openpgp4fpr` URIs. VCard specification allows embedding public key information as an URI in [`KEY` field](https://tools.ietf.org/html/rfc6350#section-6.8.1). Two schemes used with this field - `https` and `data` are [either insecure or not practical](https://www.av8n.com/computer/htm/distributing-keys.htm#sec-fing). As the value of `KEY` field is a URI one can use `openpgp4fpr` URI there to have both secure and small links.

This idea is hinted at in [Distributing PGP Keys and Fingerprints](https://www.av8n.com/computer/htm/distributing-keys.htm#sec-fing) with this text:
> It would be nice if .vcf supported PGP fingerprints, as opposed to entire keys or mere URLs, but there is no sign that it does.

## How Has This Been Tested?
Built a debug Fdroid APK and tested on live phone scanning QR code from https://metacode.biz/@wiktor
The QR code can be decoded in browser with: https://qrcode.online/reader using URL https://metacode.biz/@wiktor/card.png

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
@@ -49,6 +49,8 @@ import org.sufficientlysecure.keychain.util.Preferences;
 
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Proxy activity (just a transparent content view) to scan QR Codes using the Barcode Scanner app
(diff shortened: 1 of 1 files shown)


Valodim:
Sounds reasonable. I don't see it being used terribly often in practice, but it's a simple change so why not :)

wiktor-k:
Thank you! It's not used because it wasn't supported by open Keychain obviously! :)

PR #2323, opened 2018-05-15
Add support for NDEF tags with `openpgp4fpr` URIs

wiktor-k:
## Description

This change allows importing keys from NDEF tags that have fingerprint encoded as an `openpgp4fpr` URI when the application is in background.

After scanning a tag the `Import keys` activity is shown with results of the search for key fingerprint encoded in the tag.

## Motivation and Context

Cheap NFC tags can be used as an alternative to barcodes when exchanging public keys. They can also be used as a business card, that, when scanned, will open OpenKeychain and look for the key.

## How Has This Been Tested?
Built a fdroid variant, OpenKeychain moved to background (by pressing Android Home key), prepared and scanned NDEF tag with a sample URI (`openpgp4fpr:73EE2314F65FA92EC2390D3A718C070100012282`). Import keys window is shown.

After scanning the tag only standard Import Keys window is shown.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -20,6 +20,7 @@ package org.sufficientlysecure.keychain.ui;
 import android.app.Activity;
 import android.content.Intent;
 import android.net.Uri;
+import android.nfc.NfcAdapter;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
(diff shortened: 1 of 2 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -131,7 +132,11 @@ protected void handleActions(@NonNull Intent intent) {
                 // delegate action to ACTION_IMPORT_KEY
                 action = ACTION_IMPORT_KEY;
             }
+        } else if (NfcAdapter.ACTION_NDEF_DISCOVERED.equals(action) && "openpgp4fpr".equalsIgnoreCase(scheme)) {
```
Valodim: use FINGERPRINT_SCHEME constant

Valodim:
I wonder if this is going to see a lot of use :) but it's such a simple change, why the hell not.

PR #2325, opened 2018-05-19
Adaptive icons

lokesh-krishna:
## Description
This adds support for adaptive icons in devices running Android 8.0 and above while continuing to display the legacy icon on devices running older versions of Android.

## Motivation and Context
Starting from November 2018, apps need to target API 26 in order to be able to push updates to the Play Store. When an app targets API 26 but does not provide an adaptive icon, the existing icon is put on a white background and masked according to whatever choice the users or OEM make in their launcher.

By providing an adaptive icon, this situation is avoided. An app doesn't need to target API 26 in order to provide an adaptive icon and merely building against SDK 26 should suffice, which means that users can enjoy an adaptive icon even as the team works on other changes that will let them target API 26.

## How Has This Been Tested?
As I cannot use Android Studio on my present laptop, I've not actually been able to test this change by building it. I however used [Adapticon](https://adapticon.tooo.io/) to test the icon itself and it seems to be fine.

I am no designer and the icon was made by messily playing around with the SVG provided and I understand if this might not be up to the team's standards. This is however the best that I can do and somebody else will need to step in to do a better job if this won't do.

screenshot_2018-05-19 adapticon
screenshot_2018-05-19 adapticon 1
screenshot_2018-05-19 adapticon 2
screenshot_2018-05-19 adapticon 3

## Types of changes


ArchangeGabriel:
I my opinion, you’re missing some kind of shadow from the key onto the background. One random example: https://lh3.googleusercontent.com/_4zBNFjA8S9yjNB_ONwqBvxTvyXYdC7Nh1jYZ2x6YEcldBr2fyijdjM2J5EoVdTpnkA=s360.

lokesh-krishna:
I am sorry but I am no graphic designer and I don't think I can implement a drop shadow properly. However it isn't uncommon for an adaptive icon to not have a drop shadow. A lot of Google's own apps (YouTube, Play Games, News, YouTube Kids and even the main Google app).

Of course, it would be better to have a drop shadow but I don't think I can do it.

kowalski7cc:
I think it would be better to provide a vectorial adaptive icon as well multiple DPI variants raster icons like here: https://github.com/edotassi/AmazMod/commit/242974747873a2722550ac11978da08e2cc429f3

I'll try tomorrow to come up to something with Android Studio

kowalski7cc:
Check out #2335

Valodim:
this seems to be superseded by #2335, thanks for kicking this off @lokesh-krishna !

PR #2327, opened 2018-05-20
Improve key search with HKP servers

### LINKED ISSUE #2061: Mailvelope keyserver does not work
Add Mailvelope Keyserver with cert at https://github.com/mailvelope/keyserver/issues/7

Search for thomas@mailvelope.com

Compare
* https://keys.mailvelope.com/pks/lookup?op=index&options=mr&search=thomas@mailvelope.com
* https://pgp.mit.edu/pks/lookup?op=index&options=mr&search=thomas@mailvelope.com

jatoko:
## Description

This updates the handling of key queries with hkp servers by trying to conform to [OpenPGP HKP](https://tools.ietf.org/html/draft-shaw-openpgp-hkp-00) and allowing to use all keyservers for search.

Any feedback is welcome!
## Motivation and Context

- If a key search query is unsuccessful (no keys, or error) one had to change keyserver selection manually. For convenience, a button was added, suggesting to search other servers.
- Partly fixes #2061 as the handling of hkp server responses did not respect valid results as defined in [OpenPGP HKP](https://tools.ietf.org/html/draft-shaw-openpgp-hkp-00), especially disrespecting empty field.

## How Has This Been Tested?

Tested with hkps://keys.mailvelope.com and default hkp servers.
**Added a tag to other servers as "fallback" servers.**

**Added a button, in case the search on the server was unsuccessful.**

**Added a tag for the keys origin in case it is not the primary key server.**

Added a tag for the keys origin in case it is not the primary key server.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
@@ -18,6 +18,14 @@
 package org.sufficientlysecure.keychain.keyimport;
 
 
+import android.support.annotation.NonNull;
+import okhttp3.*;
+import org.sufficientlysecure.keychain.network.OkHttpClientFactory;
+import org.sufficientlysecure.keychain.pgp.PgpHelper;
+import org.sufficientlysecure.keychain.ui.util.KeyFormattingUtils;
+import org.sufficientlysecure.keychain.util.ParcelableProxy;
+import timber.log.Timber;
+
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.Proxy;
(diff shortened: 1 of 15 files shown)


Valodim:
Hi, thanks for contributing!

All of those sound like good ideas, great work! Could I trouble you to separate the unrelated parts of this PR (hkp fix, fallback keyserver setting, fallback keyserver button, display keyserver on top of card) into separate commits? That makes review easier and our commit history stays nice and clear.

jatoko:
Thank you for your feedback, I tried to adopt to the suggested seperation of commits.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
```diff
@@ -102,19 +99,33 @@
...
+     * @see <a href="http://tools.ietf.org/html/draft-shaw-openpgp-hkp-00#section-5.2">
+     * 5.2. Machine Readable Indexes</a>
+     * in Internet-Draft OpenPGP HTTP Keyserver Protocol Document
      */
-    private static final Pattern UID_LINE = Pattern
-            .compile("uid:([^:]*):([0-9]+):([0-9]*):([rde]*)",
-                    Pattern.CASE_INSENSITIVE);
+    private static final Pattern PUB_KEY_LINE = Pattern
+            .compile( "(?<pub>pub:" +                 // group 1
+                            "(?<pubKeyID>[0-9a-fA-F]+)" +   // group 2
```
Valodim: you can use non-capturing groups here if they're not needed. also, why use named groups without referencing them later? `(?:XXX)`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/adapter/ImportKeysAdapter.java
```diff
@@ -292,6 +291,16 @@ private void mergeEntryWithKey(ImportKeysListEntry entry, CanonicalizedKeyRing k
         entry.setExpired(keyRing.isExpired());
         entry.setSecure(keyRing.isSecure());
 
+        int algorithmId = keyRing.getPublicKey().getAlgorithm();
+        int bitSize = keyRing.getPublicKey().getBitStrength();
+        String algoInfo = KeyFormattingUtils
+                .getAlgorithmInfo(algorithmId, bitSize, null);
+        if(entry.getAlgorithm() == null){
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/SettingsKeyserverFragment.java
```diff
@@ -31,16 +27,9 @@
...
 import android.support.v7.widget.RecyclerView;
 import android.support.v7.widget.helper.ItemTouchHelper;
-import android.view.LayoutInflater;
-import android.view.Menu;
-import android.view.MenuInflater;
-import android.view.MenuItem;
-import android.view.MotionEvent;
-import android.view.View;
-import android.view.ViewGroup;
+import android.view.*;
```
Valodim: we don't use wildcard imports

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysListFragment.java
```diff
@@ -277,8 +272,50 @@ public void onLoadFinished(
         mAdapter.setData(data.getResult());
         int size = mAdapter.getItemCount();
 
+        toggleKeyboard(false);
+
         mBinding.setNumber(size);
-        mBinding.setStatus(size > 0 ? STATUS_LOADED : STATUS_EMPTY);
+        if(size > 0){
```
Valodim: code style

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/adapter/ImportKeysAdapter.java
```diff
@@ -83,7 +84,10 @@ public void setData(List<ImportKeysListEntry> data) {
         mKeyStates = new KeyState[data.size()];
         for (int i = 0; i < mKeyStates.length; i++) {
             ImportKeysListEntry entry = mData.get(i);
-
+            entry.setIsPreferredKeyserverAddress(
+                    Preferences.getPreferences(mActivity).getPreferredKeyserver().getUrl()
```
Valodim: just check if the keyserver instances are equal

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/bindings/ImportKeysBindings.java
```diff
@@ -84,4 +83,12 @@ public static void setCreation(TextView textView, Date creationDate, boolean rev
         textView.setTextColor(ImportKeysBindingsUtils.getColor(context, revokedOrExpiredOrInsecure));
     }
 
+    @BindingAdapter("keyKeyserver")
+    public static void setKeyserver(TextView textView, CharSequence keyserver) {
+        String text = "";
+        if (keyserver == null) keyserver = "";
```
Valodim: code style: missing braces

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/bindings/ImportKeysBindings.java
```diff
@@ -84,4 +83,12 @@ public static void setCreation(TextView textView, Date creationDate, boolean rev
...
+        else keyserver = keyserver.toString().replaceFirst("^hkps?://", "");
```
Valodim: introduce something like `getReadableName`

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
Valodim: ah, I see. alright then :)

with non-capturing groups you can avoid capturing all parenthesis, so you don't need to skip numbers (2, 3, 4 instead of 2, 4, 6)

jatoko:
> The problem with it is that we can expect users to always click this button if there are no results. That essentially makes it a useless click - if there may be more results, why ask the user if they should be loaded, if there is no other actionable way? 

I understand your concern and agree that users will probably continue the search in most cases. In my opinion, for some "advanced" users who set their own keyservers, the search should not just contiue and present "arbitary" results without their aknowledgement. Though I agree that the button does not solve the problem acurately. 

> most keyservers are part of the pool, so it's unlikely we'll get more results if we search on others

I do not agree, especially with the pool I experience very long response times from some servers or make the app throw errors due to invalid certificates. So I really like to "hot"-switch servers without lots of menu clicking.
To not bother common users with such issues and to further hide keysearch complexity from them, I suggest to rework the search and will open a feature request to rethink the issue.

> I would definitely like to merge the hkp compatibility commit though, care to make a separate PR for it?

[Done](https://github.com/open-keychain/open-keychain/pull/2337)

dschuermann:
closed because: keys.openpgp.org is the correct answer to most problems.

PR #2329, opened 2018-05-22
Extend WKD support to include WKD URLs

wiktor-k:
## Description
This change allows handling Web Key Directory links by OpenKeychain (they always have scheme `https` and path starting with `/.well-known/openpgpkey/hu`).

This change also fixes two issues with WKD: 1) handling e-mails with leading/trailing spaces 2) following redirects on WKD links, just like gnupg does (some hosts like `kernel.org` use it). It also adds tests for WKD URL construction utilities.

## Motivation and Context
This change completes support of WKD in #2270 and allows importing keys over WKD using links similarly to how Facebook or HKP links are handled.

## How Has This Been Tested?

Deployed debug build on a phone and tested links mentioned in #2270.

After clicking on a link.
screenshot_1526978160

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
@@ -207,14 +207,14 @@ public class ImportKeysActivity extends BaseActivity implements ImportKeysListen
                 String fbUsername = FacebookKeyserverClient.getUsernameFromUri(dataUri);
 
                 Preferences.CloudSearchPrefs cloudSearchPrefs =
-                        new Preferences.CloudSearchPrefs(false, true, true, false, null);
+                        Preferences.CloudSearchPrefs.createSocialOnly();
                 // search immediately
                 startListFragment(null, null, fbUsername, cloudSearchPrefs);
                 break;
             }
             case ACTION_IMPORT_KEY_FROM_WEB_KEY_DIRECTORY: {
                 Preferences.CloudSearchPrefs cloudSearchPrefs =
-                        new Preferences.CloudSearchPrefs(false, false, false, true, null);
+                        Preferences.CloudSearchPrefs.createWebKeyDirectoryOnly();
                 // search immediately
                 startListFragment(null, null, dataUri.toString(), cloudSearchPrefs);
                 break;
@@ -223,8 +223,7 @@ public class ImportKeysActivity extends BaseActivity implements ImportKeysListen
                 // get keyserver from URL
                 HkpKeyserverAddress keyserver = HkpKeyserverAddress.createFromUri(
                         dataUri.getScheme() + "://" + dataUri.getAuthority());
-                Preferences.CloudSearchPrefs cloudSearchPrefs = new Preferences.CloudSearchPrefs(
-                        true, false, false, false, keyserver);
+                Preferences.CloudSearchPrefs cloudSearchPrefs = Preferences.CloudSearchPrefs.createKeyserverOnly(keyserver);
                 Timber.d("Using keyserver: " + keyserver);
 
                 // process URL to get operation and query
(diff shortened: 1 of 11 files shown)


OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/WebKeyDirectory.java
```diff
@@ -0,0 +1,64 @@
...
+import android.support.annotation.Nullable;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class WebKeyDirectory {
```
Valodim: *Util

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -208,6 +212,13 @@ protected void handleActions(@NonNull Intent intent) {
                 startListFragment(null, null, fbUsername, cloudSearchPrefs);
                 break;
             }
+            case ACTION_IMPORT_KEY_FROM_WEB_KEY_DIRECTORY: {
+                Preferences.CloudSearchPrefs cloudSearchPrefs =
```
Valodim: would be nicer to use a factory method here

wiktor-k:
Okay, now it should be better. I wanted to delay AutoValue until further PR but if you're insisting I added it here. Also, I've added the preference pane, the commit looks small. I hope this entire PR did not grow too big ;-)

Thanks for your review!

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/Preferences.java
```diff
@@ -362,62 +361,32 @@ public void setKeySignaturesTableInitialized() {
...
 
-    public static class CloudSearchPrefs implements Parcelable {
-        public final boolean searchKeyserver;
-        public final boolean searchKeybase;
-        public final boolean searchFacebook;
-        public final boolean searchWebKeyDirectory;
-        public final HkpKeyserverAddress keyserver;
+    @AutoValue
+    public static abstract class CloudSearchPrefs implements Parcelable {
+        public abstract boolean searchKeyserver();
```
wiktor-k: I didn't have better name for these accessor methods... `isSearchKeyserver`? `isSearchingKeyserver`? `shouldSearchKeyserver`? If you've got better idea tell me and I'll adjust the code.

Valodim: Perhaps isKeyserverEnabled?

wiktor-k:
Okay @Valodim, this should be it. Phew...

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
wiktor-k: Ah, that's what you meant, I thought this was about `autovalue` only ;) Okay, let me fix that....

wiktor-k: Okay fixed, there are two new methods used now: `createWebKeyDirectoryOnly` and `createKeyserverOnly`.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysActivity.java
```diff
@@ -203,17 +207,23 @@ protected void handleActions(@NonNull Intent intent) {
                 String fbUsername = FacebookKeyserverClient.getUsernameFromUri(dataUri);
 
                 Preferences.CloudSearchPrefs cloudSearchPrefs =
-                        new Preferences.CloudSearchPrefs(false, true, true, false, null);
+                        Preferences.CloudSearchPrefs.create(false, true, true, false, null);
```
Valodim: :)

Valodim:
Nice work, thanks!

PR #2331, opened 2018-05-22
WIP: Disable personalized learning for password input widgets

### LINKED ISSUE #2330: Passphrase stored in keyboard predictions?
From the mailinglist:

## Current Behavior
I am using Google keyboard and I realize that the passphrase I use is stored in the predictions. I I believe it is possible to switch the keyboard in "private mode" before asking the passphrase.

## Expected Behavior
Not stored as predictions

## Possible Solution

https://developer.android.com/reference/android/support/v13/view/inputmethod/EditorInfoCompat#IME_FLAG_NO_PERSONALIZED_LEARNING

hagau:
Fix #2330 

## Description
Disable personalized learning for password input widgets

## How Has This Been Tested?
Not tested.

## Types of changes


hagau:
According to:

https://developer.android.com/reference/android/support/v13/view/inputmethod/EditorInfoCompat#IME_FLAG_NO_PERSONALIZED_LEARNING

>Applications need to be aware that the flag is not a guarantee, and some IMEs may not respect it.

This is obvious, but not encouraging. Forcing a non-third-party IME for password input would be preferable.

Valodim:
This still a WIP? Even if it's not ideal, if it's the best we can do we should merge it. Using a non-third-party IME seems intrusive to me, and only marginally improves the extremely bad situation where all other passwords of the user get uploaded to the cloud.

PR #2351, opened 2018-06-19
Use SqlDelight in favor of ContentProviders

Valodim:
This is a huge PR, moving our database query interface from ContentProviders to a bunch of specialized DataAccessObjects (Daos), using SqlDelight for database model generation and LiveData for loading semantics.

This is not a finished effort, some of the more tricky queries are still handled by KeychainProvider. I was able however to move the RecyclerView in KeyListFragment over, using the fabulous FlexibleAdapter lib. That move alone dropped more than a thousand lines of code :)


Valodim:
This is coming along quite nicely. I replaced almost all loading of key data in KeychainProvider with DAOs at this point, using LiveData and ViewModels for loading. This simplifies the loading code quite a bit in many places, sql queries are nicely readable, and data is also refreshed more smoothly than before. Almost dropped 5k lines here! :)

Valodim:
as ready as it'll ever be!

PR #2358, opened 2018-06-28
Add launcher shortcuts for encrypting text and files

Valodim:
Adds encrypt text and files shortcuts from the launcher icon. I used this in development, but it might make sense to have those? Didn't really think about what shortcuts could be useful to have, just opening this as a suggestion


Valodim:
this was at some point integrated into #2351

Issue #2387, opened 2018-08-01
How to export public key to clipboard?


gulkily:
Please forgive me if this is obvious, but I've looked all over the UI and the FAQ and haven't found the answer. How do I export a public key from my keychain so that I can paste it into another app?

## Expected Behavior

Easy way to copy public key exists

## Current Behavior

Cannot figure out how to copy public key to clipboard.

## Steps to Reproduce (for bugs)

1. Open the OpenKeychain app on Android
2. Ensure that at least one key exists in keychain
3. Attempt to copy public key to clipboard somehow
4.

## Context

I'm trying to copy my public key so that I can use it in another application. Not being able to copy my key makes it impossible to register a profile in the application I'm using.

## Your Environment

* Android Version: 6.0
* Device Model: Blu Vivo 5 mini
* OpenKeychain Version: 49200
* From Google Play or F-Droid?: Google Play

hagau:
First of all: update your OpenKeychain! Version 4.9.2 is fairly old by now.

To copy a public key to the clipboard:
- select the key then select `Advanced` from the options menu on the top right hand side of the screen.
- in the `Advanced` properties of the key, select the `Share` tab
- select `Share with...`  to either share with other apps or copy to clipboard

Valodim:
If anyone wants to make a PR, we'd be happy to have a Share button in the view key activity.

Grossdm:
I'm with @Valodim w.r.t. having a more accessible means of copying/sharing your public key.

A button would be fine, but also on a context menu would be nice.

Thanks to all the developers!
Doug

gulkily:
Thank you guys so much for responding so quickly and thoroughly! You're awesome.

Grossdm:
This Pull Request #2388 looks great!

Thanks for the fast response!

PR #2388, opened 2018-08-03
Add key sharing buttons to key view

hagau:
## Description
Add buttons for sharing a key to primary key view.

## Motivation and Context
See https://github.com/open-keychain/open-keychain/issues/2387

## How Has This Been Tested?
Tested manually.

## Types of changes


Valodim:
Hmm, four icons in that position feels like one too many to me. I wonder if we can get away with only one Share button, and put "Share to clipboard" as one of the actions using `EXTRA_INITIAL_INTENTS` (like [here](https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/DecryptListFragment.java#L555))?

Issue #1881, opened 2016-05-25
Randomness in tests


nmikhailov:
Travis build failed.
Probably due non determenism of https://github.com/open-keychain/open-keychain/blob/ce5e5f36c56fe14cb4a90a912eece07e4cb92285/OpenKeychain/src/test/java/org/sufficientlysecure/keychain/util/TestingUtils.java#L30

Using random(especially without fixed seed) in unit tests is generally a bad practice.

Valodim:
A failure due to `genPassphrase` would mean that there are passphrases which break our routines. I agree though that even if tests fail when they "detect" corner cases like this due to randomness, it's very difficult to find the actual cause. We might want to get rid of the randomness in tests..

runnerway:
Today i got this on a Travis build. I don't know if it's related.

```
org.sufficientlysecure.keychain.pgp.UncachedKeyringTest > testArmorIdentity FAILED
    java.lang.RuntimeException
        Caused by: java.lang.NegativeArraySizeException at UncachedKeyringTest.java:86


org.sufficientlysecure.keychain.pgp.UncachedKeyringTest > testDecodeEncodeMulti FAILED
    java.lang.Exception
        Caused by: java.lang.NullPointerException at UncachedKeyringTest.java:130
```

PR #2389, opened 2018-08-07
WIP: Use static passphrases in tests

hagau:
## Description
Remove all usage of `genPassphrase` and use common, static passphrases instead. 

## Motivation and Context
Fix https://github.com/open-keychain/open-keychain/issues/1881

## How Has This Been Tested?
Passes several test suite runs.

## Types of changes

### DIFF
--- a/OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/BackupOperationTest.java
+++ b/OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/BackupOperationTest.java
@@ -79,11 +79,9 @@ import static org.mockito.Mockito.when;
 @RunWith(KeychainTestRunner.class)
 public class BackupOperationTest {
 
-    static Passphrase mPassphrase = TestingUtils.genPassphrase(true);
+    static Passphrase mPassphrase = TestingUtils.testPassphrase0;
 
     static UncachedKeyRing mStaticRing1, mStaticRing2;
-    static Passphrase mKeyPhrase1 = TestingUtils.genPassphrase(true);
-    static Passphrase mKeyPhrase2 = TestingUtils.genPassphrase(true);
 
     static PrintStream oldShadowStream;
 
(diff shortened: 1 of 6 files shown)


OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/BackupOperationTest.java
```diff
@@ -79,11 +79,11 @@
 @RunWith(KeychainTestRunner.class)
 public class BackupOperationTest {
 
-    static Passphrase mPassphrase = TestingUtils.genPassphrase(true);
+    static Passphrase mPassphrase = TestingUtils.testPassphrase0;
```
hagau: 1. This is unused. Extracting the constants and cleaning up the tests might be a worthwhile but boring task (for another PR?).

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/pgp/PgpEncryptDecryptTest.java
```diff
@@ -464,7 +464,7 @@ public void testAsymmetricSignDetached() {
     @Test
     public void testAsymmetricEncryptDecrypt() {
 
-        String plaintext = "dies ist ein plaintext ☭" + TestingUtils.genPassphrase(true);
+        String plaintext = "dies ist ein plaintext ☭" + TestingUtils.testPassphrase0;
```
Valodim: I don't think appending passphrases to plaintext like this is necessary, just keep the plaintext on its own

hagau: Yeah, I know this is pretty silly, but I'll rather do a seperate PR for cleaning up the tests a bit.

hagau: Ok, resolved.

OpenKeychain/src/test/java/org/sufficientlysecure/keychain/operations/BackupOperationTest.java
hagau: Ok, I've resolved this.

Valodim:
:+1:

PR #2393, opened 2018-09-06
Fix saving XMPP URIs in phone Contact Book

wiktor-k:
## Description
Fixes saving XMPP URIs in phone Contact Book

## Motivation and Context

Previously XMPP URIs were saved as non functional e-mails in Android built-in Contact Book. This change saves them as a special IM field that is used [directly by other applications](https://github.com/siacs/Conversations/issues/3158#issuecomment-418985376) (e.g. Conversations).

## How Has This Been Tested?

Tested on a real device with a key with UID `xmpp:test@example.com`. After signing that key and synchronizing contacts OpenKeychain created a contact with e-mail field from primary UID and Jabber IM field

Key in OpenKeychain (debug):
key-sm

Created contact in Phone Book:
contact-details-sm

## Types of changes


Valodim:
Yeah why not. lgtm, thanks :)

PR #2398, opened 2018-10-01
WIP: Subkey selection for authentication

hagau:
## Description
Add subkey selection for authentication.

## Motivation and Context
A keyring might have multiple valid auth subkeys, so allow for selecting one of those.

## How Has This Been Tested?
Works for me.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/AndroidManifest.xml
+++ b/OpenKeychain/src/main/AndroidManifest.xml
@@ -927,6 +927,11 @@
             android:theme="@style/Theme.Keychain.Transparent"
             android:label="@string/app_name" />
         <activity
+            android:name=".remote.ui.dialog.RemoteSelectAuthenticationSubKeyActivity"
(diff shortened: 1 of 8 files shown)


Valodim:
As elaborated offline, I believe this is the wrong approach to the issue. This kind of subkey micromanagement *should not* be done by the user, because it muddies up the role of the primary key as a cryptographic identity. I'd rather not have workflows in OpenKeychain that involve manual selection of subkeys.

There is a middle way here that should cover all users' workflows however: Simply offer all available subkeys of the selected primary key to the openssh server during connection attempt, one after another. Incidentally, that is also what ssh-agent does. Adding this is a low priority enhancement, but I'd be happy to merge if anyone else wants to give that a whirl.

On a related note, the bug we had with subkey selection that sometimes selected revoked or expired subkeys should be fixed with #2397.

PR #2404, opened 2018-10-19
Add Secalot hardware token support.

Secalot:
## Description
This patch adds support for the Secalot hardware token, www.secalot.com.
Secalot supports OpenPGP Card specification version 2.1.1 with RSA keys up to 2048 bits.
Initialization with open-keychain is possible, card reset is not.

## Motivation and Context
We want to be compatible with as many applications as possible.

## How Has This Been Tested?
The change was tested with a Secalot device connected to a Galaxy S8 phone with a USB-A to USB-C adapter. Card initialization, signing and encryption work.

## Types of changes

## P.S.
Dear maintainers, if you want a device for future regression testing, can you please drop a message at hi@secalot.com with an address where to send, and we will send you one for free.


dschuermann:
Secalot works, I tested key gen, sign, decrypt, reset.

PR #2436, opened 2019-01-13
add Oversec to listed apps

mgeier63:
## Description
Add Oversec to listed apps

## Motivation and Context
Oversec is now OSS


mgeier63:
Hmm, the failed Travis build doesn't seem to be caused by my change. 
What's the best way to re-run it?

PR #2488, opened 2019-09-08
Add checksum-dependency-plugin for verification of plugin/dependency checksums

vlsi:
`checksum-dependency-plugin` is a superset of `gradle-witness`, and it enables to increase the level of security.

See https://github.com/vlsi/vlsi-release-plugins#checksum-dependency-plugin

## Description

This PR adds PGP-based verification for the dependencies.
I don't remove `gradle-witness.jar` yet, however it becomes outdated, and I can remove that if you like.

## Motivation and Context

`gradle-witness` is cool, however it has inherent issues (like inability to verify plugins, lack of support for `java-library`, inability to verify PGP checksums).

So I've implemented `checksum-dependency-plugin`:
* Gradle plugins can be verified (grade-witness doesn't track plugins)
* All Gradle configurations are supported (e.g. `java-library` plugin is supported). `checksum-dependency-plugin` intercepts detached configurations as well (e.g. the ones that are created on demand)
* PGP can be used for verification. PGP can be used with or without checksum. PGP enables to detect and prevent issues like https://blog.autsoft.hu/a-confusing-dependency/

## How Has This Been Tested?

Technically speaking it is build-only change, so I don't think extra tests need to be added

## Types of changes
- ✅ Buildscript-only change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ android:
 
 script:
   - if [ ${TEST} == "unit" ]; then
-    ./gradlew --stacktrace testDebug jacocoTestReport;
+    ./gradlew --stacktrace testDebug jacocoTestReport -PchecksumPrint -PchecksumFailOn=build_finish;
     fi
   - if [ ${TEST} == "android" ]; then
     echo no | android create avd --force --name test --target $ANDROID_TARGET --abi $ANDROID_ABI;
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ ext {
 #### Add new library
 * You can add the library as a Maven dependency or as a git submodule (if patches are required) in the "extern" folder.
 * You can get all transitive dependencies with ``./gradlew -q dependencies OpenKeychain:dependencies``
-* If added as a Maven dependency, pin the library using [Gradle Witness](https://github.com/WhisperSystems/gradle-witness) (Do ``./gradlew -q calculateChecksums`` for Trust on First Use)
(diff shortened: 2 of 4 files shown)


vlsi:
Hi, I see the PR has not been reviewed much yet :)

You might be interested that [Gradle 6.2 introduces in-core dependency verification](https://github.com/gradle/gradle/issues/10443#issuecomment-568741472)

The documentation can be reviewed here: https://github.com/gradle/gradle/pull/11755

So, for now, the options are like "upgrade to Gradle 6.2 and use in-core verification" or "use checksum-dependency-plugin.

From what I know Gradle would cover more cases when compared with `checksum-dependency-plugin`. For instance, it will be able to verify `pom.xml` which are implicitly fetched by Gradle when resolving transitive dependencies and probably other cases.

Some bits can be previewed in the current release candidates, release nightly builds and master nightly builds (see https://gradle.org/releases/ )

It would be nice if you could preview the feature and provide your feedback.

Valodim:
Hi @vlsi, thanks for your work on verification, I see you also contributed to the gradle PR :)

If gradle has builtin support for verifications that's definitely the way to go on the long run. I just did some preliminary testing and android studio didn't seem to like verification-metadata.xml yet.

hannesa2:
This pull request is obsolete, there is no Travis anymore.
This would be the replacement https://github.com/open-keychain/open-keychain/pull/2555

@dschuermann

vlsi:
Just in case, #2555 is not a replacement. #2555 does not verify the dependency supply chain. An alternative option is Gradle Dependency Verification: https://docs.gradle.org/current/userguide/dependency_verification.html

hannesa2:
The Gradle verify action verify SHA-256 checksum the https://github.com/marketplace/actions/gradle-wrapper-validation

PR #2493, opened 2019-09-28
Adding support for WKD Advanced method

Anders-Linden:
This change extends the WKD support with Advanced mode stated in
RFC Draft: draft-koch-openpgp-webkey-service-08 section 3.1

Advanced mode shall be queried first and use direct mode as fallback
this implements the design from the draft. 
Notice: This might change in future
versions of RFC Draft draft-koch-openpgp-webkey-service

## Description
Added advanced method advanced method URI lookup and previous 
direct mode as fallback. 

## Motivation and Context
When bigger providers or WKD providers are available they would
probably go with advanced mode for custom domains. 
Key owners which cannot add .well-known folder on web service due
to services provider limitation needs an other method for publish keys
on WKD with subdomain openpgpkey a WKD provider or own http
server can be used.

## How Has This Been Tested?
Adding unit tests. Tested in emulator, WKD Direct mode tested and WKD Advanced mode
with real domains. 

## Types of changes
Added Boolean variable wkdMethodAdvanced  parameter on function WebKeyDirectoryUtil.toWebKeyDirectoryURL()


dschuermann:
Great work thanks!

PR #2529, opened 2020-02-16
Add query parameter to WKD URLs

eehakkin:
This change adds the l=LOCAL-PART query parameter to WKD URLs as
specified in RFC Draft: draft-koch-openpgp-webkey-service-09 section 3.1
since version 07.

## Motivation and Context
The current code without this change does not work with WKD servers which expect the l=LOCAL-PART query parameter specified in the RFC Draft since version 07 to be present.

## How Has This Been Tested?
I ran the test functions in the WebKeyDirectoryUtilTest.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/WebKeyDirectoryUtil.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/util/WebKeyDirectoryUtil.java
@@ -2,8 +2,10 @@ package org.sufficientlysecure.keychain.util;
 
 import android.support.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


Valodim:
Hey there, thanks for this PR!

Can you add a test to check that uri encoding is correctly applied? otherwise, lgtm :+1:

eehakkin:
> Can you add a test to check that uri encoding is correctly applied? otherwise, lgtm +1

I added a test which checks that test–wkd@openkeychain.org (with an en dash instead of a hyphen) is encoded correctly.

Valodim:
I woulda expected an encoded :hankey: or somethin', but that'll do. Thanks! ;)

PR #2586, opened 2020-09-05
Drop more stuff

Valodim:
Dropping more outdated and unused features. Reduce code size and get rid of some cruft. Most of the stuff few will miss - analytics and the shop fragment, in particular.

One thing I dropped here that could be argued about is the "contacts sync" feature, which connects keys to the user's contacts. While this is a useful feature on paper, I doubt it was used much or at all (and our analytics confirm this). More importantly though, it causes issues and crashes on some platforms, and we didn't keep it up to date in years.


mathieujobin:
I don't know why you could remove account sync.
maybe your analytics showed it as unused because it was broken?

how am I suppose to find the keys from everyone in my contact list without contact sync?

I have to search for everyone individually?

Valodim:
The contact sync feature didn't include a way to do what you are describing.

mathieujobin:
but if it could sync contacts, at least maybe I could click each contact individually to search for available keys.
now, I need to switch app and type their name/email to search... 
anyway, I think I'm just gonna delete the app and wait another decade

PR #2596, opened 2020-09-27
Implement password KDF

### LINKED ISSUE #2368: Bad PIN! when using my gnuk token
I've just got a usb token with gnuk (I replaced neug with gnuk in an FST-01G). I put a keyring on it and things seem to work fine on my computer. In OpenKeychain, I imported the keys from the token in such a way that I get a new entry in "My Keys". Nevertheless, when trying to decrypt a file using the token, I always get "Bad PIN!" error message. The PIN I enter is good. I even cleared the password cache and restarted the phone.

I tried to change the user PIN from gnupg and got the same "Bad PIN!" message after that in OpenKeychain. 

I tried to change the user PIN from OpenKeychain and got 3 input fields: 1 for the admin PIN and 2 for the user PIN. When I validated the popup, I got a "Hold Security Token against the NFC…" popup even though my token was inserted already. I unplugged the token and pluged it back and got a "Bad PIN!" error message again.

## Your Environment

* Android Version: 8.1.0 with LineageOS 15.1-20180709
* Device Model: OnePlus 1 (bacon)
* OpenKeychain Version: 5.1.4
* From Google Play or F-Droid?: F-Droid

hagau:
## Description
Implement password KDF as described in OpenPGP application specifications v3.3 chapter 4.3.2.

This fixes #2368.

## Motivation and Context
OpenPGP application specifications v3.3 introduce the ability to send a repeatedly hashed and salted password instead of plaintext to a token.

## How Has This Been Tested?
Text encryption/decryption tested with a Nitrokey Start and a Yubikey with KDF enabled.
Backwards compatibility tested with a random card with KDF disabled.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
@@ -26,12 +26,20 @@ import androidx.annotation.Nullable;
 import com.google.auto.value.AutoValue;
 
 
+/**
+ * References:
+ * [0] `Functional Specification of the OpenPGP application on ISO Smart Card Operating Systems`
+ *      version 3.4.1
+ *      https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.1.pdf
+ */
 @SuppressWarnings("unused") // just expose all included data
 @AutoValue
 public abstract class OpenPgpCapabilities {
+    // Extended Capabilites flag bit offsets are defined on page 32 of [0]
     private final static int MASK_SM = 1 << 7;
     private final static int MASK_KEY_IMPORT = 1 << 5;
     private final static int MASK_ATTRIBUTES_CHANGABLE = 1 << 2;
+    private final static int MASK_KDF_DO = 1;
 
     private static final int MAX_PW1_LENGTH_INDEX = 1;
     private static final int MAX_PW3_LENGTH_INDEX = 3;
@@ -63,6 +71,7 @@ public abstract class OpenPgpCapabilities {
     abstract boolean isHasSM();
     abstract boolean isHasAesSm();
     abstract boolean isHasScp11bSm();
+    abstract boolean isHasKdf();
 
     @Nullable
     abstract Integer getMaxCmdLen();
@@ -135,6 +144,7 @@ public abstract class OpenPgpCapabilities {
         abstract Builder hasSM(boolean hasSm);
         abstract Builder hasAesSm(boolean hasAesSm);
         abstract Builder hasScp11bSm(boolean hasScp11bSm);
+        abstract Builder hasKdf(boolean hasKdf);
 
         abstract Builder maxCmdLen(Integer maxCommandLen);
         abstract Builder maxRspLen(Integer MaxResponseLen);
@@ -147,6 +157,7 @@ public abstract class OpenPgpCapabilities {
             hasSM(false);
             hasAesSm(false);
             hasScp11bSm(false);
+            hasKdf(false);
         }
 
         Builder updateWithTLV(Iso7816TLV[] tlvs) {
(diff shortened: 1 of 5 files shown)


hagau:
Error handling should probably be extended.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
```diff
@@ -304,6 +309,140 @@ public void clearSecureMessaging() {
 
     // region pin management
 
+    private byte[] executeKdf(byte[] pin, int password_type) throws IOException {
```
Valodim, line 312: This is a long function, that does two separate things:
* retrieve and parse the KDF DO
* calculate actual kdf based on parameters

I suggest splitting those parts up, as such:
```
byte[] calculateKdfIfNecessary(byte[] pin, int type) {
  if (!capabilities.isHasKdf) return
  KdfParameters kdfParameters = retrieveKdfDo(); // can cache internally
  return KdfCalculator.calculateKdf(kdfParameters.forType(type), pin);
}
```

the actual kdf calculation is a pure function, and we should definitely test it. there is a test vector in the [opgnpgp card spec, page 20](https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.pdf).

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
```diff
@@ -32,6 +32,7 @@
     private final static int MASK_SM = 1 << 7;
     private final static int MASK_KEY_IMPORT = 1 << 5;
     private final static int MASK_ATTRIBUTES_CHANGABLE = 1 << 2;
+    private final static int MASK_KDF_DO = 1;
```
Valodim, line 35: while we're at it, please add a comment to the [spec](https://gnupg.org/ftp/specs/OpenPGP-smart-card-application-3.4.pdf) and page (32) where those constants come from

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
```diff
@@ -304,6 +309,140 @@ public void clearSecureMessaging() {
...
+        digester.update(data, 0, r);
+
+        byte[] digest = new byte[digester.getDigestSize()];
+        digester.doFinal(digest, 0);
+
+        // delete secrets from memory
+        Arrays.fill(data, (byte) 0);
+
+        // these are less sensitive, but let's be thorough
+        Arrays.fill(salt, (byte) 0);
```
Valodim, line 436: these are available without authentication from the card, so they indeed aren't sensitive. I think we might as well cache the DO once we retrieved it (see above)

Valodim (changes requested):
Hey Christian, good to see you're still around :wave: 

I suggested one refactoring, in particular I'd love to have a test against the vector from the spec. Otherwise everything looks straightforward and functional, thanks for the PR! :+1:

hagau:
Hi Vincent :wave: 
Those are all perfectly valid suggestions, I'll implement them later tonight

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
hagau: I've removed those lines.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/OpenPgpCapabilities.java
hagau: I've added the spec as a reference here & wherever it made sense to do so.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/SecurityTokenConnection.java
hagau: I've split the function up into smaller, more sensible units & added a test for the KDF calculation.
Any comment on the resulting code would be greatly appreciated :slightly_smiling_face:

szszszsz:
Hi! Just bumping this PR up. What's left here to do?

Valodim:
thanks! will re-review later :)

PR #2600, opened 2020-10-15
Update Regex for QRcode scanned VCard

### LINKED ISSUE #2599: QR Code + VCard | Syntax not working
In the current version of the OpenKeychain App, The QRcode Vcard syntax does not seem to work.
Using the proper syntax, OpenKeychain reports improper format.
## Expected Behavior

Using the vcard syntax:
```
BEGIN:VCARD
FN:Austin Milne
EMAIL:[REDACTED]
KEY:OPENPGP4FPR:6E8329BB07B245820C91E88B409B6A5A30BE7AAD
END:VCARD
```

Then, using various QRcode generating tools ([qrencode](https://linux.die.net/man/1/qrencode) being the best example), I should be able to create a QRCode.

<details>
<summary> The generated QRcode </summary>

test
</details>

I should then be able to scan this QRCode with OpenKeychain and Import the key from the PGP servers.

<details>
<summary> My Actual use case (Both Tested) </summary>

```
BEGIN:VCARD
VERSION:4.0
PROFILE:VCARD

FN:Austin William Milne
N:Milne;Austin;William;;
GENDER:M
BDAY:[REDACTED]
TEL;TYPE=HOME:[REDACTED]
ADR;TYPE=HOME:;;;Ottawa;ON;K1S3N4;CA

EMAIL;TYPE=HOME:[REDACTED]
TITLE; Mechanical Engineering Student
ORG:University of Waterloo;Mechanical Engineering

PHOTO;MEDIATYPE=image/jpeg:gravatar.com/avatar/25b5a0f791326094bf9e3fa93ab97cc4?s=300

URL:me.awbmilne.ca
URL;TYPE=LinkedIn:linkedin.com/in/austin-w-milne-254b92170/
URL;TYPE=Facebook:facebook.com/A.Will.Milne/
URL;TYPE=Instagram:instagram.com/awbmilne/
URL;TYPE=Github:github.com/awbmilne
URL;TYPE=Gitlab:gitlab.com/awbmilne

NOTE:PGP: 0x6E8329BB07B245820C91E88B409B6A5A30BE7AAD
KEY:OPENPGP4FPR:6E8329BB07B245820C91E88B409B6A5A30BE7AAD

UID:urn:uuid:e391c1cd-43fa-4544-aaa4-eaa852c20fbb
REV:20201014T173000Z
END:VCARD
```

test
</details>

## Current Behavior

When scanning the QRCode, I get the error:
```
[ERROR] QR Code format not supported (format specification: https://github.com/open-keychain/open-keychain/wiki/QR-Codes)!
```

## Possible Solution

Could be simple regex parsing error

1. Generate VCard QRCode
2. Attempt Scanning
3. Experience Failure
4. Cry Inside

## Context

The primary reason I've come accross this issue is that I am trying to have single QRCode for Sharing my Contact Info and PGP information.

When importing VCards, mobile devices tend to ignore the `KEY:` values. But the OpenKeychain does not, so it should be easy to have my cake and eat it too.

## Environment

* Android Version: 11 Build: RP1A.200720.009
* Device Model: Pixel 3 XL
* OpenKeychain Version: 5.5 (55000)

Awbmilne:
Fix issues with QRCode VCard Key importing | closes #2599 

Regex needs to be corrected for handling values in a VCard format.

It also seems possible to extend the import functionality to the URL syntax of the Vcard Standard.

## Description
Currently:
- Fix parsing of `KEY:OPENPGP4FPR:[HASH]` with better regex
- Extend regex to full-length search URLs in VCard standard format\
     For example: `KEY;MEDIATYPE=application/pgp-keys:https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x[40-CHAR-HASH]`

Prospective:
- Extend to All VCard formats using URL import
    ```
    2.1:   KEY;PGP:http://example.com/key.pgp
    3.0:   KEY;TYPE=PGP:http://example.com/key.pgp
    4.0:   4.0: KEY;MEDIATYPE=application/pgp-keys:http://example.com/key.pgp
    ```
## Motivation and Context
I have been looking to streamline my VCard Contact and PGP information, This would make it easier to deal with on android. Easier to share Keys and Contact info with same QRcode on Business card or such. Refer to #2599 

## How Has This Been Tested?
Hasnt been tested yet. Not ready for Pull yet.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java
@@ -136,7 +136,20 @@ public class ImportKeysProxyActivity extends FragmentActivity
         }
     }
 
-    private static final Pattern VCARD_KEY_PATTERN = Pattern.compile("\nKEY:(.*)\n");
+// XXX : Bookmark for issues #2592 (https://github.com/open-keychain/open-keychain/issues/2592) @awbmilne
+/**
+ * ─── Regex Explaination ───
+ * (?<=\n)(?:KEY:OPENPGP4FPR:|KEY.*http.*search=0x)((?:[A-F]|[a-f]|\d){40})(?=\n)
+ * 
+ * Matches:
+ * KEY:OPENPGP4FPR:[HASH]
+ * KEY[anything]https[anything]search=0x[HASH]
+ * 
+ * Examples:
+ * KEY:OPENPGP4FPR:1234567890123456789012345678901234567890
+ * KEY;MEDIATYPE=application/pgp-keys:https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x1234567890123456789012345678901234567890
+ */
+    private static final Pattern VCARD_KEY_PATTERN = Pattern.compile("(?<=\n)(?:KEY:OPENPGP4FPR:|KEY.*http.*search=0x)((?:[A-F]|[a-f]|\d){40})(?=\n)");
 
     private void processScannedContent(String content) {
         // if a VCard was scanned try to extract the KEY field


Awbmilne:
For someone more experienced with this project...

Is there a significant reason we cannot import PGP keys from VCards that have the VCard standard syntax?
Why we cant import from custom URL?

This seems like a really useful feature... But maybe there is a security reason or something else?

It seems to me that looking up PGP keys by Fingerprint on the PGP servers would be the same security as from a custom `https://` host. The VCard maker would have to have set that URL, same as setting the Fingerprint in the custom VCard field.

Awbmilne:
With the Latest commit:
```
 ─── Regex Explaination ───
 - Full String -
(?<=\n)(?:KEY:OPENPGP4FPR:|KEY.*?https.*?\/pks\/lookup\?op=get&search=0x|KEY.*?https:\/\/keys.openpgp.org\/vks\/v1\/by-fingerprint\/)([a-fA-F0-9]{40})(?=\n)

?:KEY:OPENPGP4FPR:                                           <- OPENPGP4FPR values
KEY.*?https.*?\/pks\/lookup\?op=get&search=0x                <- Majority of SKS Server pool URLs
KEY.*?https:\/\/keys.openpgp.org\/vks\/v1\/by-fingerprint\/  <- keys.openpgp.org URLs
([a-fA-F0-9]{40})                                            <- Fingerprint

Matches:
KEY:OPENPGP4FPR:[HASH]
(Most) PGP Server URLs with Full Hash embedded

Examples:
KEY:OPENPGP4FPR:ABAF11C65A2970B130ABE3C479BE3E4300411886
KEY;PGP:https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xABAF11C65A2970B130ABE3C479BE3E4300411886
KEY;TYPE=PGP:https://keys.openpgp.org/vks/v1/by-fingerprint/ABAF11C65A2970B130ABE3C479BE3E4300411886
KEY;MEDIATYPE=application/pgp-keys:https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xabaf11c65a2970b130abe3c479be3e4300411886
```

wiktor-k:
Hi @Awbmilne, the reason why fingerprint format is supported but not arbitrary URLs is the following: the fingerprint provides end-to-end verification of the key while the URL could point to any key.

HTTPS does not protect against malicious server/VPS admin replacing the key or against your domain expiring and someone squatting your domain name. The fingerprint format on the other hand does not relay on the format but verifies that the key is the correct one.

Do note that business cards don't change often so it's important to have a correct key info all the time there.

Awbmilne:
Thanks @wiktor-k 

>  issue with your QR encoded VCard in the original issue is that you used \r\n instead of \n

Good find! Wouldve never though of line endings. Windows is really a pain for raw text files.🥴\
I made changes to my suggested Regex in 87d4bda, so that it can handle CRLF and LF line endings.

> HTTPS does not protect against malicious server/VPS admin replacing the key or against your domain expiring and someone squatting your domain name. The fingerprint format on the other hand does not relay on the format but verifies that the key is the correct one.

This is a good point, I agree. Forcing the use of well-established and well-maintained key servers is a good idea.

Luckily, My existing changes comply with this sentiment!\
The purpose of the Regex is to extract the 40 Character Key Fingerprint from the `KEY:` field's URL. OpenKeychain then uses that fingerprint to search for the key on the established Key Servers. This means the URL is never used, it just serves as another format for the Fingerprint.\
While this doesnt work for direct download links of Public Key Files, Like: It does work for Links to Public Keys pages on SKS Servers. For example, my own key: http://keys.gnupg.net/pks/lookup?op=get&search=0x6E8329BB07B245820C91E88B409B6A5A30BE7AAD

This may not be useful for many people, but it does expand the options for QRcode based VCards.

If you are looking for unit test examples, [Lines 154-158](https://github.com/open-keychain/open-keychain/blob/87d4bdacc3f7d07a1d3543796e7e75f8a08018b5/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/ImportKeysProxyActivity.java#L154-158) has some typical examples of URLs that this type of Regex fingerprint extraction should work for. (_Fingerprint used being Linus Torvalds' Key. May want to change for unit tests, as his key is multiple MB these days_)

Awbmilne:
## IMPORTANT NOTE:

The VCard standard [RFC6350](https://tools.ietf.org/html/rfc6350) states in [Section 3.2](https://tools.ietf.org/html/rfc6350#section-3.2) that:
> Individual lines within vCard are delimited by the [RFC5322] line break, which is a CRLF sequence (U+000D followed by U+000A).

Meaning that VCard files are "supposed" to be standardized with `CRLF` (`\r\n`) line ending.

Personally, This seems like a dumb decision for the standardization. (Feels like Microsoft had some financial pull there)\
But, I believe it would be best to handle both Unix and Windows line ending.

The commits above handle both cases (though, Unit testing should be added for verification, Which is past my knowledge).

Awbmilne:
### Secondary Note:

The [`openpgp4fpr:<fingerprint>` standard ](https://metacode.biz/openpgp/openpgp4fpr) outlines that the Fingerprint is:
>... exactly 40 characters of lower and upper-case letters A, B, C, D, E, F and digits (0-9).

The use of Uppercase and Lowercase digits is accounted for in the new Regex.

PR #2624, opened 2020-12-15
partial fix for issue 2590

schch:
change behavior to what GnuPG does.
c.f. `static void
parse_algorithm_attribute (app_t app, int keyno)` line 5952 in https://github.com/gpg/gnupg/blob/master/scd/app-openpgp.c#L5952

According to the spec (ftp://ftp.gnupg.org/specs/OpenPGP-smart-card-application-3.4.1.pdf Section 4.4.3.9 Algorithm Attributes) the last byte can only be optional in the sense that you set it to 0x00. The other allowed value seems to be 0xFF. It is not clear from the spec, but that is what GnuPG does.

This does not fix the problem with a Yubikey setting random values for the last byte in the first response.

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KeyFormat.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KeyFormat.java
@@ -57,7 +57,7 @@ public abstract class KeyFormat {
                     throw new IllegalArgumentException("Bad length for EC attributes");
                 }
                 int len = bytes.length - 1;
-                if (bytes[bytes.length - 1] == (byte)0xff) {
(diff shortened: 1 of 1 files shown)


dschuermann:
According to the spec it can be 0xff or it is not present.
GnuPG seems to remove 0x00 if it is present, but does not consider it a pubkey template.
I would consider merging it, when we know why 0x00 should be removed from the oid bytes.

Do you know why GnuPG does this? git blame shows that it's there since the beginning.

schch:
Sorry, I don't know why. I guess, it is because implementations send 0 anyway as the flag is optional. It looks like the OID-Values are always considered to be a string, hence 0 termination sounds reasonable.

PR #2626, opened 2020-12-17
set the key attr to ECDSA for AUT when using EC keys

dangfan:
Set the key attr to ECDSA for AUT when using EC keys.

## Description
According to the [spec](ftp://ftp.gnupg.org/specs/OpenPGP-smart-card-application-3.4.1.pdf), in Section 4.4.3.9 Algorithm Attributes, ID with ECDSA should be used for INT-AUT, which is currently set to ECDH.

## Motivation and Context
For the specification compliance.

## Types of changes
- ✅ Bug fix


dschuermann:
thanks! Fixed here: https://github.com/open-keychain/open-keychain/commit/fb0651eda1b0ea45679cbc31ec73467b2c47dded

Issue #2570, opened 2020-07-26
Generating RSA keys to NFC token


kusyka911:
## Expected Behavior
The app can successfully generate keys to hardware security token

## Current Behavior
After gey was generated on the mobile phone, the app trying to write then on security token, but it fails every time.

### key generate
1. Open app
2. try to create a new key with security token
3. chose `RSA3072` or `RSA3096` key types
4. Pay attention to the result

## Your Environment
* Android Version: 10.0.5 (MIUI Global 11.0.5 Stable)
* Device Model: Remi Note 8 pro
* OpenKeychain Version: 5.5
* From Google Play or F-Droid?: F-Droid
* Security token: Yubikey 5 NFC

kusyka911:
Looks like issue was fixed.

Issue #2235, opened 2017-12-07
OpenPGP Card 3: key generation fails with 'key export to security token failed'


techge:
When trying to generate a key other than RSA2048 (that is RSA3072, RSA-4096 and ECC options) the key generation on OpenPGP Card v3 on Nitrokey Pro is failing.

## Expected Behavior
Generate keys on device in other form than RSA2048 as proposed in process.

## Current Behavior
Fails before PIN can typed in, but after accepting everything else.

## Steps to Reproduce (for bugs)
1. put in a factory-reset device with OpenPGP Card v3 or factory-reset it with Open-keychain
2. try to create new keys
3. choose something different than RSA2048

![screenshot](https://user-images.githubusercontent.com/25867995/33718947-76195f88-db5f-11e7-968e-88884e212f6a.png)

## Your Environment
* Android Version: 7.1.2 (LineageOS of Nov 2017)
* OpenKeychain Version:  48001
* From F-Droid

dschuermann:
Interesting, as you made two reports about the Nitrokey Pro, I assume it is possible to switch between OpenPGP spec 2 and 3? Or do I need to flash my Nitrokey Pro for this?

techge:
No it is not possible to simply switch. It is a change in hardware (changing the used card on the device). I have a testdevice with OpenPGP Card 3 here. Theoretically this can be done with existing device by opening the case and having a OpenPGP Card 3 at hand.

PR #2627, opened 2020-12-18
Computes the lengths in the template dynamically

dangfan:
Computes the lengths in the template dynamically according to the mod length when importing RSA keys.

## Description
Currently, the lengths of the RSA templates are hard-coded as 0x80 and 0x100, which are the values of RSA 2048.
However, RSA 3072 & 4096 have different values. Therefore, importing keys other than 2048 might be failed.

## Motivation and Context
Fixes the hard-coded lengths in the RSA templates when importing RSA keys to smart tokens.

## Types of changes
- ✅ Bug fix


dangfan:
May fix #2570

dangfan:
May fix #2235

Valodim:
looks good, thanks :+1:

PR #2643, opened 2021-02-04
Fixed root property check in build.gradle

fgndev:
## Description

The check for the compileSdkVersion property must not include the classpath
Since I'm building with current SDK and build tools I ran into this.
A suggestion for checking the buildToolsVersion in the same manner is added to this PR.

## How Has This Been Tested?
Tested with android-30 and buildToolsVersion 30.0.2

## Suggestion
Add the same checks to opengpg-lib as well since it is included in the same way as sshauthentication.

### DIFF
--- a/sshauthentication-api/build.gradle
+++ b/sshauthentication-api/build.gradle
@@ -13,11 +13,14 @@ buildscript {
 }
 
 android {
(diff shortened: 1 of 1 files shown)


dschuermann:
Thanks!

PR #2652, opened 2021-02-17
Add missing break statement in KDF-DO parser

### LINKED ISSUE #2650: Nfc error when approaching a yubico 5 nfc
after updating to version 5.6 the encryption via NFC no longer works.  in practice, before I approached yubico Nfc to the cell phone and confirmation of encryption came out, now instead when it asks to approach the yubico via Nfc, it crashes or opens Chrome.  the control no longer applies at the time of the nfc

Screenshot_20210215-184004

Screenshot_20210215-184028

hagau:
## Description
The KDF-DO parser was missing a break statement, resulting in parsing a token with KDF ability, but having KDF disabled, as a token with KDF enabled, thus resulting in a Autovalue error upon trying to parse the KDF-DO returned by the token.

## Motivation and Context
Fixes #2650

## How Has This Been Tested?
Not tested, since obviously correct :).

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KdfParameters.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/KdfParameters.java
@@ -85,6 +85,7 @@ public abstract class KdfParameters {
                             case (byte)0x00:
                                 // no KDF, plain password
                                 hasUsesKdf(false);
+                                break;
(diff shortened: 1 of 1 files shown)


Valodim:
ho boy. thanks for the fix :)

hagau:
Sorry for not catching that earlier. I remember auditing the code in January, but apparently overlooked this :flushed:

PR #2672, opened 2021-03-16
Don't run BouncyCastle tests (make master work)

hannesa2:
It prevents from 

>* What went wrong:
> Could not determine the dependencies of task ':extern:bouncycastle:test'.
>
> Task with path ':core:test' not found in project ':extern:bouncycastle'.

bouncycastle should test there own stuff in there repo


hannesa2:
@dschuermann

PR #2676, opened 2021-03-17
Simplify Jitpack

hannesa2:
As it looks like this Jitpack file is obsolete

## Output without this change

image

See https://jitpack.io/#open-keychain/open-keychain/v5.7

## Output with this change

image

See https://jitpack.io/#hannesa2/open-keychain/5.7.1

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,2 +1,0 @@
-install:
-  - ./gradlew clean :sshauthentication-api:assemble :sshauthentication-api:publishToMavenLocal :openpgp-api-lib:assemble :openpgp-api-lib:publishToMavenLocal


dschuermann:
If I don't include this file, Jitpack will build the full project (which takes a long time and may run out of memory).
The file ensures that only the two libraries are built.

hannesa2:
For me it's noise

dschuermann:
Check out old logs and you will see that Jitpack can easily run out of memory: https://jitpack.io/com/github/open-keychain/open-keychain/v5.6.2/build.log

Newer log is much shorter: https://jitpack.io/com/github/open-keychain/open-keychain/v5.7/build.log

PR #2679, opened 2021-03-27
Only skip hashing when creating an EDDSA authentication signature

### LINKED ISSUE #2677: Pre-hash EdDSA signatures with NFC Security Key for signature subkeys
I have a yubico 5 nfc with key ed25519 and pin kdf enabled on on, when I try to encrypt and sign even a simple image file it goes into signature error.  I attach error.  can it be kdf pin compatibility problem or key eddsa signature?  I am attaching screenshot error

Screenshot_20210321-185225_OpenKeychain

dhedberg:
## Description
A comment in bug #2677 notes that hashing should only be skipped when generating a signature for ssh authentication.

## How Has This Been Tested?
I have verified that ssh authentication still works with a yubikey 5c nfc and openssh/okc-agent as in #2662. I have also tested encrypting / signing files and text using the built-in functionality, and decrypting / verifying them both using the built-in functionality and GnuPG.

Fixes #2677


dhedberg:
I'm unable to reproduce the failing tests locally, they all appear to pass for me (using openjdk-1.8.0 from fedora 33)

dschuermann:
Thx for the contributions. There is some randomness in the tests, which we should fix at some point...

Issue #2690, opened 2021-04-16
Key search does not work


stelian42:
Key search on standard keyserver yields no results.

## Expected Behavior

It should find the key.

## Current Behavior

It doesn't find anything.

## Your Environment

* OpenKeychain Version: latest git, locally compiled

stelian42:
Fixed by PR #2691

Valodim:
Pretty odd that this broke along the way. I remember testing this not too long ago when I fixed #2638

dschuermann:
I released 5.7.4 with this fix.

PR #2691, opened 2021-04-16
Fix parsing of EOL in keyserver answer

stelian42:
Searching for keys on the default keyserver (keys.openpgp.org) yields no results.

## Description

Upon debug, it looks like the answer from the server contains Windows-style EOL, which is incorrectly parsed by the current regexps.

## Motivation and Context

Fixes bug #2690 

## Types of changes


Valodim:
Looks good. Thanks!

dschuermann:
I released 5.7.4 with this fix.

PR #2692, opened 2021-04-17
fix HkpKeyserverClient not handling unix-style line endings

reyawn:
## Description
In frustration while debugging a separate issue, I noticed that the PGP keyserver protocol importer was always failing to import my keys regardless of which keyserver I was pulling from.  I noticed that the correct data was being fetched but always failed to parse.

As an example of input, see the result of the exact same query submitted by the importer when searching for my own key [when searching for my own key](https://keyserver.ubuntu.com/pks/lookup?op=index&options=mr&search=reyawn%40pobox.com):

```
info:1:1
pub:D05646A4B5826453:1:4096:1432686821::
uid:Robert Earl Yawn <reyawn@pobox.com>:1609715025:1672787025:
uid:Robert Earl Yawn <reyawn@stokedmind.net>:1609715025:1672787025:
uid:Robert Earl Yawn <reyawn@gmail.com>:1609715025:1672787025:
uid:Robert Yawn <robert@ridedott.com>:1609715025:1672787025:

```

The parser correctly matches the UID lines because the UID regex uses `\r?\n` for matching line endings, but fails to match the public key line since the public key regex uses `\r\n` (a DOS line ending).  Updating the regex fixes the issue and will match both DOS and UNIX style line endings.

## Motivation and Context

Allow importing from PGP keyservers that use UNIX-style line endings in output.

I couldn't find a specific issue, but a number of issues referenced flakiness in general with the keyserver imports that could potentially be resolved by this.

## How Has This Been Tested?

I tested by searching for `reyawn@pobox.com`, my own primary UID.

This change is localized only to the PGP keyserver importer.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/keyimport/HkpKeyserverClient.java
@@ -128,7 +128,7 @@ public class HkpKeyserverClient implements KeyserverClient {
                             "(?::([0-9]*)" +      // group 5
                             "(?::([0-9]*)" +      // group 6
                             "(?::((?:(?=(?:r(?!(.?r))|d(?!(.?d))|e(?!(.?e))))[rde]){0,3})" + // group 7
-                            ")?)?)?)?)?\r\n)"// pub line
+                            ")?)?)?)?)?\r?\n)"// pub line
                             + "(" + UID_LINE.pattern() + // group 11
                             "+)", // one or more uid lines
                     Pattern.CASE_INSENSITIVE


reyawn:
The tests passed locally, need to see what's up here.  The change shouldn't have impacted anything.

dschuermann:
thanks! This is already fixed on master with https://github.com/open-keychain/open-keychain/pull/2691

reyawn:
> thanks! This is already fixed on master with https://github.com/open-keychain/open-keychain/pull/2691

This fixed the fix on master.  There is a difference. :). The pubkey Line makes the carriage return required there, which breaks UNIX-style line endings.

dschuermann:
you are right thanks :) . I fixed it on master.

PR #2693, opened 2021-04-21
fix the compatibility of compressed and uncompressed format in ECDH

### LINKED ISSUE #2681: "Error: Invalid KEK!" with security key and ECC NIST curve
## Expected Behavior

I'm trying to decrypt a file using my healthy key hosted in a security token.

## Current Behavior

Instead of decrypting the file if shows me an error (more on it later).

1. Open menu
2. Choose Encrypt/Decrypt
3. Click Select input file
4. Choose a *.gpg file 
A message is displayed that ask me to "Hold Security Token against the NFC marker at the back of your device."
5. Hold the security token like requested
A message is shown: "Error: Invalid KEK!"

## Your Environment

* Android Version: 10 - LineageOS 17.1
* Device Model: Xiaomi Redmi Note 8T
* OpenKeychain Version: 5.7.2 (57200)

dangfan:
Fix #2681 

## Description
From rfc6637#section-7: The input of KDF should be the x portion of the point.
However, the result of ECDH can be expressed in two formats: compressed and uncompressed. GnuPG can deal with both cases but before Open-Keychain 5.7, it only supports the compressed format and after 5.7 (since commit e82fb82) it only supports the uncompressed format.
This PR first checks if the ECDH is the compressed format and then extracts the x portion accordingly.

## Motivation and Context
See #2681 

## How Has This Been Tested?
Tested on SmartPGP, which has the same behavior as YubiKey.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/operations/PsoDecryptTokenOp.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/operations/PsoDecryptTokenOp.java
@@ -161,14 +161,23 @@ public class PsoDecryptTokenOp {
        */
         byte[] keyEncryptionKey = response.getData();
 
-        int xLen;
-        if (eckf.isX25519()) {
-            xLen = keyEncryptionKey.length;
-        } else {
+        /* From rfc6637#section-7 :
+        The input of KDF should be the x portion of the point.
+        As the result of ECDH can be expressed in two formats: compressed and uncompressed,
+        we have to deal with each case.
+        */
+        int xLen, startPos;
+        if (keyEncryptionKey[0] == 0x04 && keyEncryptionKey.length % 2 == 1) {
+            // uncompressed format
             xLen = (keyEncryptionKey.length - 1) / 2;
+            startPos = 1;
+        } else {
+            // compressed format
+            xLen = keyEncryptionKey.length;
+            startPos = 0;
         }
         final byte[] kekX = new byte[xLen];
-        System.arraycopy(keyEncryptionKey, eckf.isX25519() ? 0 : 1, kekX, 0, xLen);
+        System.arraycopy(keyEncryptionKey, startPos, kekX, 0, xLen);
 
         final byte[] keyEnc = new byte[encryptedSessionKeyMpi[mpiLength + 2]];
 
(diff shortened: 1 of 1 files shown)


dschuermann:
Hi @dangfan thanks for helping us out with this. What configurations have you tested exactly?

dangfan:
https://github.com/ANSSI-FR/SmartPGP

dschuermann:
and both compressed and uncompressed curves?

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/operations/PsoDecryptTokenOp.java
```diff
@@ -161,14 +161,23 @@ The shared secret obtained is a KEK (Key Encryption Key) that is used to wrap th
...
-        if (eckf.isX25519()) {
-            xLen = keyEncryptionKey.length;
-        } else {
+        /* From rfc6637#section-7 :
+        The input of KDF should be the x portion of the point.
+        As the result of ECDH can be expressed in two formats: compressed and uncompressed,
+        we have to deal with each case.
+        */
+        int xLen, startPos;
+        if (keyEncryptionKey[0] == 0x04 && keyEncryptionKey.length % 2 == 1) {
```
dschuermann, line 170: do we really need ``&& keyEncryptionKey.length % 2 == 1``? Shouldn't 0x04 be sufficient?

dangfan:
> and both compressed and uncompressed curves?

I've updated the information above.

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/securitytoken/operations/PsoDecryptTokenOp.java
dangfan: An uncompressed point is encoded as `04 || x || y`, with `x` and `y` are of the same size. However, a valid `x` may be led with `04`, so 0x04 is not sufficient.

dschuermann: can you add this as a comment?

dangfan: Sure, I've added the description.

dschuermann:
@dangfan Thanks! I improved the variable naming and logic in two subsequent commits on master. Can you verify that the master still behaves correctly with your test configuration?

dangfan:
I have verified the lastest commit a9b28dd and it worked correctly with these two test configurations. Thanks!

dschuermann:
I released 5.7.4 with this fix.

Issue #2688, opened 2021-04-09
Signature not verified when signing using an OpenPGP card


dingwen07:
When using FairEmail or K-9 Mail to create a signed (signed and encrypted or signed only) e-mail, the signature of the e-mail cannot be verified on the mail clients using OpenPGP.js (Thunderbird 78, ProtonMail, Mailvelop tested). This is not a problem with OpenPGP.js, because signed emails sent via GpgOL can be verified on OpenPGP.js. Moreover, if the OpenPGP card is not used when signing, the problem will not arise.

## Expected Behavior
The signature is verified on all clients

## Current Behavior
Signature created using OpenPGP card is not verified on OpenPGP.js based clients

## Steps to Reproduce (for bugs)

1. Generate a PGP key pair
2. Import it into an OpenPGP card
3. Send a signed e-mail from FairEmail or K-9 Mail using the key on the OpenPGP card
4. Open Thunderbird to receive the message, the signature could not be verified, with the error message "Invalid Digital Signature"
5. Import the private key of the PGP key into OpenKeychain
6. Send a signed e-mail from FairEmail or K-9 Mail again, this time with private key stored in OpenKeychain
7. Open Thunderbird to receive the message, the signature could be verified

## Your Environment

* Android Version: 11
* Device Model: SM-G9750, Xiaomi M2007J3SC
* OpenKeychain Version: 5.7.3
* From Google Play or F-Droid?: Google Play

stelian42:
I see this here as well (step 1 to 4), running latest git version on Android 9.

Didn't try steps 5 to 7 though.

dschuermann:
Is this issue also happening without external hardware?

stelian42:
If I understand correctly the OP steps 5 to 7, I'd say no.

dingwen07:
> Is this issue also happening without external hardware?

No, only with hardware token for my tests.

stelian42:
I've just discovered that the signature is verified correctly in mutt , even when generated by a key on a hardware token. This means that the contents of the signature is ok, it must be some formatting issue somewhere which disturbs Thunderbird.

dingwen07:
> I've just discovered that the signature is verified correctly in mutt , even when generated by a key on a hardware token. This means that the contents of the signature is ok, it must be some formatting issue somewhere which disturbs Thunderbird.


[...]

PR #2695, opened 2021-04-22
Fix digest handling when signing using a hardware token

stelian42:
The previous code used java.security.MessageDigest.digest()
more than once, which is incorrect because the digest is reset
after each call.

This made OpenKeychain generate invalid signatures when
using a hardware token: the two bytes called "fingerprint",
which are supposed to be the first two bytes of the unsigned
digest, were incorrectly filled inside the signature.

Some tools do not check this fingerprint, only the signed digest,
so the signature was reported as valid. GPG is one of these tools.

But some other tools (like OpenPGP.js which uses rnp) did check it,
and reported an invalid signature, assorted with an error message
saying "wrong lbits". Thunderbird is one of the users of OpenPGP.js

This fixes bug #2688 

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/NfcSyncPGPContentSignerBuilder.java
+++ b/OpenKeychain/src/main/java/org/bouncycastle/openpgp/operator/jcajce/NfcSyncPGPContentSignerBuilder.java
@@ -111,6 +111,8 @@ public class NfcSyncPGPContentSignerBuilder
 
         return new PGPContentSigner()
         {
+            private byte[] mDigest;
+
             public int getType()
             {
                 return signatureType;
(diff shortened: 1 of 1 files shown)


dschuermann:
@stelian42 Based on your findings I re-worked the NfcSyncPGPContentSignerBuilder file and pushed some commits to master. Can you try if the current master fixes the issue the same way your PR does?

dschuermann:
@dangfan 
@dhedberg 
You both helped a lot in previous issues with hardware tokens :heart: . Can you also try the current master and check  authentication over NFC with tokens using
1) Ed25519
2) other algorithms, such as NIST P-256

and report if it still works correctly?

stelian42:
@dschuermann dschuermann I just tested your changes and it works properly for my usecase: send a signed (or signed + encrypted) mail from K9-Mail (using OpenKeychain with a hardware dongle) and succesfully decrypt and verify the signature on both mutt and thunderbird. !

You can close this PR and bug #2688 .

dhedberg:
@dschuermann I have tested ssh authentication using ed25519 with my yubikey 5c nfc, and it still works fine for me on the current master branch

dschuermann:
Thanks for testing.
We still need someone to test **without** Ed25519. @dangfan maybe?

stelian42:
@dschuermann My test was done using RSA keys.

dschuermann:
Thanks! Released as 5.7.5

dangfan:
Sorry for the delayed response. I've tested signing and decrypting using 5.7.5 (which is the master) on
- SmartPGP applet with ECC P-256
- CanoKey with Ed25519 & Curve25519
Both work fine.

dschuermann:
@dangfan Awesome thanks! No worries, you were still really fast in compared to my delays working on Openkeychain ;)

PR #2735, opened 2021-09-20
Add automatic theme

mimi89999:


Valodim:
thanks. the dark theme isn't really complete, but it's certainly better than before with this :)

PR #2776, opened 2022-07-19
 Support Create curve25519 type ECC Key

Radium-bit:
## Description
Support Create curve25519 type ECC Key.


Radium-bit:
Hope to support this feature soon.
@open-keychain

Issue #2694, opened 2021-04-21
openkeychain stopped on processing


tigernero79:
on s10 plus when i try to encrypt a text and sign it, the app gets stuck on the screen as attached photo.  "processing" 

![Screenshot_20210421-223506_OpenKeychain](https://user-images.githubusercontent.com/35575744/115618115-87db1780-a2f2-11eb-86a8-99c7e31b6348.jpg)

although the encrypted text exists in the clipboard.  I have to force close the application to be able to reuse it.  this happens on samsung s10 plus.  on my second pixel 2Xl phone the bug doesn't exist.  can it be solved?  I am attaching screenshots of the problem

Valodim:
should be fixed with 1e584a368dc8aae1a7882aecfc9e55fe5f572133 (if anyone still cares :shrug:)

PR #2816, opened 2023-01-30
Fix: #2694 openkeychain stopped on processing

Longhao-Chen:
Prevent Dialog from calling close before it is created.

## Motivation and Context
For encryption and signature of text and small files, it is possible to call close before the processing box is created, which will lead to the phenomenon in #2694 .

## Types of changes


Valodim:
Hi, thanks for your contribution!

While the issue is correctly identified, and the fix works in practice, I don't believe waiting for a short while is the correct way to fix operations that finish too fast.

I attempted another fix that worked from my tests in 1e584a368dc8aae1a7882aecfc9e55fe5f572133, please let me know if this doesn't also fix the issue in question here.

PR #2825, opened 2023-03-18
Enable Curve25519 key generation for encryption and make it the default key type

segfault-bilibili:
## Description

**I'm not very sure whether the key generation code is correct, although it seems to work during my test.**

## Motivation and Context

`mKeyTypeSpinner` has an inconsistency here: https://github.com/open-keychain/open-keychain/blob/8d69181b65c2d5364f40845466a9e5305ca8c91b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/dialog/AddSubkeyDialogFragment.java#L202-L204

If `mUsageEncrypt` is disabled because EDDSA was once selected, it won't be re-enabled on ECC_P521 or ECC_P256 being re-selected.

Actually I see Curve25519 encrypt/decrypt is supported (#2269), so that I think this should be just enabled.

## How Has This Been Tested?

Built & tested on my device running Android 12L, together with Gpg4win.

Sent encrypted file and then decrypted on the target device successfully. (both PC <=> Android)

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/Constants.java
@@ -180,13 +180,13 @@ public final class Constants {
     }
 
     /**
-     * Default key configuration: 3072 bit RSA (certify + sign, encrypt)
+     * Default key configuration: EDDSA (certify + sign), Curve25519 ECDH (encrypt)
      */
     public static void addDefaultSubkeys(SaveKeyringParcel.Builder builder) {
-        builder.addSubkeyAdd(SubkeyAdd.createSubkeyAdd(SaveKeyringParcel.Algorithm.RSA,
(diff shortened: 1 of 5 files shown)


rahilarious:
We need more traction on this PR. Are maintainers even active?

LucasMZReal:
This is definitely annoying. It's really weird how I can import a key and have it work but can't generate one in OpenKeychain. Are there alternatives to this project?

rahilarious:
> This is definitely needed, it's annoying. It's really weird how I can import a key and have it work but can't generate one in OpenKeychain. Are there alternatives to this project?

I've switched to termux+gpg. I've to operate cli on phone which makes things slow but works pretty well.

Valodim:
thanks for those fixes, sorry we didn't look at this in a long time. I'll include them in this year's targetSdk update though.

Valodim:
cleaned up a bit and pushed commits to master

PR #2860, opened 2023-11-12
Update featured app icons

schorschii:
## Description
I updated the featured app icons which are shown when the corresponding app is not installed. In addition to that, the updated icons are added as vector graphics instead of pixel graphics.

## Motivation and Context
The featured app icons of K9, Conversations and Password Store are outdated and do not represent the current design anymore. This is misleading for new and inexperienced users.

## How Has This Been Tested?
By opening the AppListFragment.

## Types of changes


Valodim:
thanks

Issue #2881, opened 2024-02-12
OpenKeyChain fails as encryption provider for K9-Mail with a `NullPointerException` in `selectUserIdStatusByEmailLike`


pokulo:
OpenKeyChain fails as encryption provider for K9-Mail with a `NullPointerException` in `selectUserIdStatusByEmailLike`.

## Expected Behavior
If the entered contact name or mail address (or text) does not match any pgp key owner, then simply hide the autocomplete
pop-up. To avoid the crash, OpenKeyChain should not raise a `NullPointerException`, but return `null`.

## Current Behavior
k9-mail crashes completely upon entering contact name to new email editor. `adb logcat` reveales the following traceback:

```java
02-12 10:45:19.578 13611 13611 D TokenAutoComplete: Current completion text: XXXXX,
02-12 10:45:19.598  8074  8103 E DatabaseUtils: Writing exception to parcel
02-12 10:45:19.598  8074  8103 E DatabaseUtils: java.lang.NullPointerException: ResultSet returned null for UserPackets.sq:selectUserIdStatusByEmailLike
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at com.squareup.sqldelight.Query.executeAsOne(SourceFile:16)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at org.sufficientlysecure.keychain.daos.UserIdDao.getUidStatusByEmailLike(SourceFile:13)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at org.sufficientlysecure.keychain.remote.KeychainExternalProvider.loadUidStatusMap(SourceFile:14)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at org.sufficientlysecure.keychain.remote.KeychainExternalProvider.query(SourceFile:246)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at android.content.ContentProvider.query(ContentProvider.java:1455)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at android.content.ContentProvider.query(ContentProvider.java:1551)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at android.content.ContentProvider$Transport.query(ContentProvider.java:285)
02-12 10:45:19.598  8074  8103 E DatabaseUtils: 	at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:107)
[...]

PR #2882, opened 2024-02-12
🐛fix Issue #2881: avoid NullPointerException in selectUserIdStatusByEmailLike

pokulo:
## Description
after upgrade of sqlitedb library sqldelight in commit 5d84bd838744f63d46d5a08110bad13b5192ae6a the method getUidStatusByEmailLike may suddenly raise a NullPointerException. This fix attempts to restore the prior behavior or simply returning null.

## Motivation and Context
It attempts to fix Issue #2881.

## How Has This Been Tested?
I did not yet test this change, this is merely a proposal, as described in Issue #2881.

## Types of changes

### DIFF
--- a/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/daos/UserIdDao.java
+++ b/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/daos/UserIdDao.java
@@ -36,7 +36,7 @@ public class UserIdDao extends AbstractDao {
 
     public UidStatus getUidStatusByEmailLike(String emailLike) {
         return getDatabase().getUserPacketsQueries().selectUserIdStatusByEmailLike(emailLike)
-                .executeAsOne();
+                .executeAsOneOrNull();
     }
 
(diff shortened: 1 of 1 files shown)


pokulo:
@Valodim Please review.

Valodim:
Perfect, thanks! Seems this one slipped through :) Overall I'm glad the breakage wasn't too bad from that one huge commit

pokulo:
Thanks for merging. Looking forward to the first bugfix-release: 6.0.1

> Perfect, thanks! Seems this one slipped through :) Overall I'm glad the breakage wasn't too bad from that one huge commit

I got the new 6.0.0 automatically via F-Droid. Which ended up breaking my K9-Mail encryption. So maybe after such a huge change, it would have been nice, to first release the 6.0.0 as an unstable/alpha version to F-Droid. So it only hits people that explicitly install unstable versions.

Valodim:
right. sorry about that

pokulo:
Any plans for version 6.0.1 release yet?

PR #2883, opened 2024-02-13
Fix passphrase dialog and dark/light mode reversal

basilgello:
## Description

First change trivially corrects the order of array elements in
theme array.

Second change fixes the wrong assumption there is only passphrase
type that requires text input from the user. The error was masked
by code dropped in PR#2864.

## Motivation and Context

Regression in 6.0.0

## How Has This Been Tested?

Build, install, restore backed-up key, try to authenticate with remote
server.

## Types of changes


basilgello:
@Valodim please review and check against physical token (I dont have any at hand now)

OpenKeychain/src/main/java/org/sufficientlysecure/keychain/ui/PassphraseDialogActivity.java
```diff
@@ -107,15 +107,22 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // this activity itself has no content view (see manifest)
         RequiredInputParcel requiredInput = getIntent().getParcelableExtra(EXTRA_REQUIRED_INPUT);
-        if (requiredInput.mType != RequiredInputType.PASSPHRASE) {
+        if (requiredInput.mType != RequiredInputType.BACKUP_CODE &&
+            requiredInput.mType != RequiredInputType.NUMERIC_9X4 &&
+            requiredInput.mType != RequiredInputType.NUMERIC_9X4_AUTOCRYPT &&
+            requiredInput.mType != RequiredInputType.PASSPHRASE &&
+            requiredInput.mType != RequiredInputType.PASSPHRASE_AUTH &&
+            requiredInput.mType != RequiredInputType.PASSPHRASE_SYMMETRIC) {
```
tcely, line 115: The thing that would help me when reading this would be to use (!(A || B)) instead.

OpenKeychain/src/main/res/values/arrays.xml
```diff
@@ -38,8 +38,8 @@
     </string-array>
     <string-array name="theme_values" translatable="false">
         <item>"auto"</item>
-        <item>"dark"</item>
         <item>"light"</item>
+        <item>"dark"</item>
```
tcely, line 42: I understand this was the least changes needed to correct the reversed themes, but the ordering bothers me.

Is there a style guide that prefers dark after light instead of using alphabetical ordering?

basilgello:
thanks @tcely ! Fixed and building now

basilgello:
Works on my side :)

Valodim:
looks good, thanks!

tcely:
@Valodim Please create a tag as soon as possible, so F-Droid includes a fixed version in the next build cycle. Thanks!

tcely:
Thanks everyone! I received my update from F-Droid and my agent integration now works as it did before `v6.0.0` was installed.

PR #2902, opened 2024-03-15
Fix: add package queries

schorschii:
## Description
Detecting installed apps is currently not working properly in API level 30+, see [here](https://stackoverflow.com/questions/34709873/packagemanager-namenotfoundexception).

## Motivation and Context
In order to fix installed app detection.

## How Has This Been Tested?
By installing the advertised apps.

This screenshot shows the current behavior when Conversations and K9 is installed, and K9 is already set up with OpenKeychain. Both are not detected as installed (tapping them will open the download pages) and the already set up K9 is only displayed as package ID instead of the correct app name.
Bildschirmfoto vom 2024-03-15 00-28-59

## Types of changes


Sami32:
Could adding FairEmail App be considered?
`eu.faircode.email`

schorschii:
I have done that for you in a separate pull request #2932, @Sami32 .

PR #2910, opened 2024-04-07
Fastlane i18n

yurtpage:
## Motivation and Context
No matter how the app is great if users can't find it and understand if it worth installing then it won't be used widely.
Translation of the app description is very important because F-Droid doesn't even makes an auto translation.

Please add the App Metadata to the Transifex or Weblate.
Here is a sample of Conversations app metadata translation via Weblate  https://translate.codeberg.org/projects/conversations/app-store-metadata-conversations/

## Types of changes


yurtpage:
@Valodim could you please merge this PR?

